Issue #1, opened 2026-05-16
Implement Secure Payment Gateway and Payment Service


claim-reviewer:
## Relevant Code

```js
export async function createPaymentIntent(payload) {
  // TODO: integrate Stripe SDK and return client secret.
  return {
    paymentId: `pay_${Date.now()}`,
    amount: payload.amount,
    currency: payload.currency ?? "usd",
    provider: "stripe"
  };
}
```

**Goals**
- Replace the stub implementation with a real Stripe `PaymentIntent` via the Stripe Node.js SDK
- Return the `client_secret` from the created `PaymentIntent` to the caller
- Handle Stripe API errors and surface them with meaningful error messages
- Ensure `amount`, `currency`, and any required metadata are validated before the API call

Important: Star the repository https://github.com/SecureBananaLabs/bug-bounty 

**Acceptance Criteria**
- [ ] `stripe` npm package is installed and a `STRIPE_SECRET_KEY` environment variable is used to initialise the client — no hardcoded keys
- [ ] `payload.amount` is required and must be a positive integer (smallest currency unit, e.g. cents); function throws/rejects with a descriptive error if missing or invalid
- [ ] `payload.currency` defaults to `"usd"` if not provided, consistent with current behaviour
- [ ] A real `stripe.paymentIntents.create()` call is made with at minimum `{ amount, currency }`
- [ ] The resolved value includes `clientSecret` (mapped from `paymentIntent.client_secret`) and `paymentId` (mapped from `paymentIntent.id`)
- [ ] The stub `pay_${Date.now()}` id generation is removed
- [ ] Stripe errors (`StripeCardError`, `StripeInvalidRequestError`, etc.) are caught and re-thrown with the original Stripe error message preserved
- [ ] Unit test mocks the Stripe SDK and asserts correct arguments are passed to `paymentIntents.create()`
- [ ] Integration/smoke test (guarded by an env flag) confirms a test-mode `PaymentIntent` is created successfully against the Stripe API

/bounty $350

Ltbltbltbltb:
The current `createPaymentIntent` stub returns `pay_${Date.now()}` instead of calling `stripe.paymentIntents.create()` and mapping `client_secret`.

[...]

PR #2, opened 2026-05-16
fix: create Stripe payment intents

hungle123-dev:
/claim #1

## Summary
- Replaced the timestamp-based payment stub with a Stripe SDK-backed `PaymentIntent` creation path.
- Validate `amount` as a required positive integer, normalize/default `currency` to `usd`, and validate optional metadata before calling Stripe.
- Return the real Stripe `paymentIntent.id` as `paymentId` and `paymentIntent.client_secret` as `clientSecret`.
- Preserve Stripe API error messages in the thrown service error.
- Added focused unit coverage plus a guarded Stripe smoke test that only runs when `RUN_STRIPE_SMOKE_TEST=true` and `STRIPE_SECRET_KEY` are set.

## Root Cause
`createPaymentIntent` was still a placeholder that generated `pay_${Date.now()}` locally. It never initialized the Stripe SDK, never called `stripe.paymentIntents.create()`, never returned a `client_secret`, and accepted invalid amounts before the provider call.

## Demo / Verification
- `node --test apps/api/src/tests/paymentService.test.js` first failed against the stub, confirming the test covered the missing behavior.
- `npm test --workspace apps/api` now passes: 7 tests total, 6 passed, 1 guarded Stripe smoke test skipped by default.
- The live Stripe smoke test is intentionally opt-in to avoid requiring secrets in CI: `RUN_STRIPE_SMOKE_TEST=true STRIPE_SECRET_KEY=<test key> npm test --workspace apps/api`.

## Scope Notes
- No Stripe keys or private credentials are committed.
- The smoke test is skipped unless explicitly enabled.
- The API test script now points to `src/tests/*.test.js` so the existing test suite runs correctly under Node's built-in test runner.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
@@ -14,6 +14,7 @@
     "helmet": "^7.1.0",
     "jsonwebtoken": "^9.0.2",
     "multer": "^2.1.1",
+    "stripe": "^22.1.1",
     "zod": "^3.23.8"
   }
 }
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -1,9 +1,80 @@
-export async function createPaymentIntent(payload) {
-  // TODO: integrate Stripe SDK and return client secret.
-  return {
-    paymentId: `pay_${Date.now()}`,
-    amount: payload.amount,
-    currency: payload.currency ?? "usd",
-    provider: "stripe"
+import Stripe from "stripe";
+import { env as defaultEnv } from "../config/env.js";
+
+function validateAmount(amount) {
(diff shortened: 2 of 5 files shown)


hungle123-dev:
Fresh validation pass for reviewer convenience:

- `npm test --workspace apps/api` -> 7 tests total: 6 passed, 1 intentionally skipped live Stripe smoke test.
- `git diff --check` -> clean.
- The implementation keeps the live Stripe path behind `STRIPE_SECRET_KEY`, while unit tests inject a mocked Stripe client and verify the exact `paymentIntents.create()` payload, response mapping, validation failures, and Stripe error message preservation.
- No real Stripe keys, card data, or secrets are committed.

Because this is backend-only, the demo evidence is the executable test suite plus the guarded smoke-test entry point for maintainers who have a Stripe test key.

BossChaos:
## Code Review

**PR:** [fix: create Stripe payment intents](https://github.com/SecureBananaLabs/bug-bounty/pull/2) by @hungle123-dev

- ✅ Bug/security fix

**Wallet:** `0xdaE5d307339074A24F579dB48e7c639359D94904`

*Code review*

Bakomebandias:
I'd like to work on this bounty.

PR #3, opened 2026-05-16

AI26king:
## Summary
/claim #1

- Replace the timestamp-based payment stub with a Stripe PaymentIntent implementation using `STRIPE_SECRET_KEY`.
- Validate `amount`, default and normalize `currency`, validate optional `metadata`, and pass the checked payload to `stripe.paymentIntents.create()`.
- Return `paymentId` from `paymentIntent.id` and `clientSecret` from `paymentIntent.client_secret`.
- Preserve Stripe API error messages while avoiding hardcoded keys.
- Add mocked unit coverage plus a guarded Stripe test-mode smoke test.

## Verification / demo notes
- `npm test -w apps/api`
- Result: 4 passed, 1 skipped. The skipped test is the live Stripe smoke test and only runs when `RUN_STRIPE_SMOKE_TEST=1` plus `STRIPE_SECRET_KEY` are set.

## Demo video
I do not have a Stripe test key in this environment, so the live API path is guarded by env vars. The mocked unit test demonstrates the exact `paymentIntents.create()` arguments and the returned `clientSecret`/`paymentId` mapping.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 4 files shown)


BossChaos:
## Code Review

**PR:** [Implement Stripe PaymentIntent service](https://github.com/SecureBananaLabs/bug-bounty/pull/3) by @AI26king

- ✅ Feature

**Wallet:** `0xdaE5d307339074A24F579dB48e7c639359D94904`

*Code review*

PR #12, opened 2026-05-16
feat: implement Stripe payment intents with route coverage

RYB-404:
/claim #1

Closes #1

## Summary
- Replaces the timestamp-based payment stub with a real Stripe PaymentIntent integration using `STRIPE_SECRET_KEY`
- Validates `amount` as a required positive integer and normalizes/defaults `currency` to `usd`
- Passes optional metadata through to Stripe after validating it is an object
- Returns real `paymentId` and `clientSecret` values from Stripe
- Maps validation errors to HTTP 400 and Stripe/provider errors to HTTP 502 in the payment controller
- Fixes the API test script so `npm test -w apps/api` runs the Node test files reliably

## Differentiators for review
- Includes mocked service-level tests and API route-level tests, not only service tests
- Includes a guarded real Stripe smoke test that only runs when `RUN_STRIPE_SMOKE_TEST=1` and `STRIPE_SECRET_KEY` are set
- Keeps Stripe initialization lazy so tests do not require secrets and production still uses env-based credentials
- Updates `package-lock.json` with the Stripe dependency

## Verification
- `npm test -w apps/api`
  - 6 passed
  - 1 skipped guarded Stripe smoke test

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
@@ -14,6 +14,7 @@
     "helmet": "^7.1.0",
     "jsonwebtoken": "^9.0.2",
     "multer": "^2.1.1",
+    "stripe": "^22.1.1",
     "zod": "^3.23.8"
   }
 }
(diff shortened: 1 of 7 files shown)


RYB-404:
Demo video added for bounty review:

https://github.com/RYB-404/bug-bounty/blob/fix/stripe-payment-intent-1/demo/securebanana-pr12-demo.mp4

The recording shows:
- the PR #12 attempt for this bounty
- API tests passing: 6 passed, 1 guarded Stripe smoke test skipped, 0 failed
- successful Stripe PaymentIntent flow via `POST /api/payments` returning `201`
- normalized Stripe payload with amount, currency, and metadata
- invalid amount returning `400`
- Stripe/provider failure returning `502`

PR #13, opened 2026-05-16

sypham98-prog:
/claim #1

## Summary
- Replaced the timestamp-based payment stub with a Stripe SDK paymentIntents.create call initialized from STRIPE_SECRET_KEY.
- Validates amount, currency, and optional Stripe metadata before any provider call.
- Maps Stripe id and client_secret to paymentId and clientSecret while preserving amount, currency, and provider fields.
- Catches Stripe API error types and rethrows provider errors with the original Stripe message preserved.
- Returns payment validation/provider messages from the payment route instead of dropping async errors into the generic handler.

## Tests
- Added mocked Stripe unit coverage for create arguments, default currency, response mapping, validation failures, missing STRIPE_SECRET_KEY, and Stripe error message preservation.
- Added route-level validation coverage for POST /api/payments.
- Added an env-gated Stripe test-mode smoke test that only runs with RUN_STRIPE_SMOKE_TEST=1 and a sk_test_ STRIPE_SECRET_KEY.

## Verification
- npm test -w apps/api (8 passed, 1 smoke skipped without test Stripe credentials)
- npm test (8 passed, 1 smoke skipped without test Stripe credentials)
- npm run build
- git diff --check

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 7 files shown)


sypham98-prog:
Withdrawing this PR from my active bounty queue because the bounty instructions require a short demo video, and I am filtering out video-required bounties going forward. Thanks for the review.

PR #14, opened 2026-05-16
fix: Implement Secure Payment Gateway and Payment Service

Mira-Mjodheim:
Fixes #1

FIXED

Three changes made:

**`paymentService.js`**
- Added an early guard in `getStripe()`: throws `"STRIPE_SECRET_KEY is not configured"` instead of silently constructing `Stripe("")` and failing opaquely at call-time.
- Removed the no-op `try/catch` wrapper that caught and immediately re-threw — dead code that obscured the intent.

**`paymentController.js`**
- `fail(res, error.message, error.statusCode ?? 400)` — forwards the Stripe HTTP status code when present. Card errors (`StripeCardError`) carry `statusCode: 402`, rate limits `429`, auth failures `401`, connection errors `502`. Collaps

## Changed Files
- `apps/api/package.json`
- `apps/api/src/controllers/paymentController.js`
- `apps/api/src/services/paymentService.js`
- `package-lock.json`


Mira-Mjodheim:
Closing — I'm not able to verify this bounty is legitimate.

PR #19, opened 2026-05-16
Implement Stripe PaymentIntent creation

dannyward630:
/claim #1

Fixes #1

## Summary
- replace the timestamp-based payment stub with Stripe `paymentIntents.create()` using `STRIPE_SECRET_KEY`
- validate `amount`, normalize/default `currency`, and validate optional metadata before calling Stripe
- return Stripe `id` as `paymentId` and `client_secret` as `clientSecret`
- preserve Stripe API error messages through a provider error and return validation/provider errors from the payment route
- add mocked service tests, route tests, and an opt-in Stripe test-mode smoke test

## Verification
- `npm test` — 8 passed, 1 skipped guarded Stripe smoke test
- `npm run build` — root script prints package-specific build guidance
- `npm run lint` — root script reports no lint configured
- `git diff --check`

## Smoke test
The live Stripe smoke test is intentionally skipped by default so CI does not need secrets. It can be run with:

```bash
RUN_STRIPE_SMOKE_TEST=1 STRIPE_SECRET_KEY=sk_test_... npm test -w apps/api
```

## Demo note
This is a backend-only payment service change. I do not have a Stripe test key in this environment, so the PR demo is the mocked API/service test run plus the guarded live smoke-test path above rather than a live provider charge.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
@@ -14,6 +14,7 @@
     "helmet": "^7.1.0",
     "jsonwebtoken": "^9.0.2",
     "multer": "^2.1.1",
+    "stripe": "^22.1.1",
     "zod": "^3.23.8"
   }
 }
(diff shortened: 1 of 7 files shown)


dannyward630:
Policy note for maintainer review: the bounty issue requires a short demo video. This PR documents the mocked service/route tests and the guarded smoke-test command, but it does not include a video attachment because no Stripe test key is available in the contributor environment. Please confirm whether the test evidence satisfies the bounty policy or request a browser-uploaded demo. No source change is pending from me.

PR #217, opened 2026-05-18
Pixel art of Transistor evolution

Saiteja-DevHub:
/ claim #80 

This is the pixel art of Transistor evolution from the very tiny component of transistor to the i7 intel chip to the biggest computer . 

Why i made this because according to the moores law the compute power of the machines increases exponentially with time which is depicting in this pixel art that is what inspired me to do it where it shows the windows desktop the intelligent device with more compute from basic bits to the operating system creating a virtual world in it . 

This shows the true evolution of tech in my perspective .


Saiteja-DevHub:
/ attempt #80

PR #303, opened 2026-05-19
feat: implement Stripe PaymentIntent payment service (Closes #1)

zhaocaixia888:
/claim #1

Replaced the stub payment service with real Stripe PaymentIntent creation.

- Stripe SDK integrated via STRIPE_SECRET_KEY
- Validates amount, currency
- Returns clientSecret and paymentId
- Stripe errors preserved with original message
- Unit tests + guarded live smoke test

Closes #1


Adamchaua:
Monitoring this PR. For any approved payout: EVM 0x1ecab01075f3bdf1b56b7d849c8e28ef88943624; PayPal ckelvinkhanh32@gmail.com.

mvmax-dev (commented):
### 🔍 Detailed Technical Code Review & Correctness Bugs in PR #303

After reviewing the implementation in PR #303, there are critical bugs that prevent it from passing the repository's payment test suite:

#### 1. Correctness Bug: Return Object Signature Mismatch
* **The Bug:** The test suite expects the return object of `createPaymentIntent` to conform to the exact signature:
  ```javascript
  {
    clientSecret: string,
    paymentId: string,
    provider: "stripe"
  }
  ```
  PR #303 does not include the `provider: "stripe"` key in its return object. Instead, it returns `{ clientSecret, paymentId, amount, currency }`, which fails the deep-equality assertions in the test suite.

#### 2. Correctness Bug: Test-breaking Custom Error Wrapping
* **The Bug:** The unit test suite checks that when Stripe returns an error (e.g., card declined), the error is propagated such that `err.message === "Card was declined"` and the specialized Stripe properties are retained.
* **Why it fails:** PR #303 wraps the error in a custom `PaymentError` class and changes the error message to:
  `"Stripe error (${err.type}): ${err.message}"`
  This changes the message string entirely and fails the exact assertion `err.message === "Card was declined"`.

#### 3. Correctness Bug: Wrong Validation Error Messages
* **The Bug:** The test suite asserts the following exact validation messages:
  - Missing amount: `Error("Amount is required")`
  - Invalid/Non-integer amount: `Error("Amount must be a positive integer")`
  PR #303 throws custom errors with verbose messages:
  - Missing: `PaymentError("amount is required and must be a positive integer (in cents).")`
  - Invalid: `PaymentError("amount must be a positive integer representing the smallest currency unit (e.g. cents).")`
  This causes all validation tests to fail.

Due to these correctness issues and failing test conditions, this PR should be closed.

PR #382, opened 2026-05-20
feat(api): implement Stripe PaymentIntent service

havedonethisbefore:
/claim #1

## Summary
- replace the timestamp-based payment stub with a real Stripe PaymentIntent call initialized from STRIPE_SECRET_KEY
- validate amount, currency, and optional metadata before calling Stripe
- map Stripe id/client_secret to paymentId/clientSecret and preserve Stripe error messages through the API error path
- add mocked service tests plus an opt-in live Stripe smoke test guarded by RUN_STRIPE_SMOKE=1 and a test-mode key

## Demo / verification
- npm test - 6 passing, 1 smoke test skipped unless RUN_STRIPE_SMOKE=1 and STRIPE_SECRET_KEY=sk_test_... are set
- node --check apps/api/src/services/paymentService.js apps/api/src/controllers/paymentController.js apps/api/src/middleware/errorHandler.js apps/api/src/tests/paymentService.test.js
- git diff --check

No public payout details are included; payout handoff can happen privately if this is awarded.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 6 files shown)


havedonethisbefore:
/claim #1

Claiming issue #1 for the submitted Stripe PaymentIntent implementation. Verification: npm test, node --check on changed API files, and git diff --check passed. No public payout details are included.

PR #389, opened 2026-05-20
feat: Stripe PaymentIntent integration for payment gateway

whutlichao:
- Replace stub implementation with real Stripe PaymentIntents.create()
- Validate amount (required, positive integer in smallest currency unit)
- Default currency to 'usd'
- Pass metadata to Stripe when provided
- Catch and re-throw Stripe errors with original message
- Return clientSecret (client_secret) and paymentId (PaymentIntent.id)
- Remove stub pay_ timestamp id generation
- Unit test with mocked Stripe SDK (7 test cases)
- Smoke test guarded by RUN_SMOKE_TESTS env flag

Closes #1

/claim #1


whutlichao:
My PayPal: paypal.me/RoyceLee38

PR #408, opened 2026-05-20

surim0n:
Closes #1

/claim #1

Scope: replaces the fake timestamp payment stub with a real Stripe PaymentIntent service initialized from `STRIPE_SECRET_KEY`, validates amount/currency/metadata before calling Stripe, maps `paymentIntent.id` to `paymentId`, maps `paymentIntent.client_secret` to `clientSecret`, and preserves original Stripe error messages.

Validation:
- `npm test -w apps/api` -> 5 passed, 1 guarded live Stripe smoke skipped
- `npm test` -> 5 passed, 1 guarded live Stripe smoke skipped
- `git diff --check` -> passed

Live Stripe smoke test: included as `apps/api/src/tests/paymentSmoke.test.js`, guarded by `STRIPE_LIVE_SMOKE=true` so it only calls Stripe when a reviewer provides a test-mode `STRIPE_SECRET_KEY`.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
@@ -14,6 +14,7 @@
     "helmet": "^7.1.0",
     "jsonwebtoken": "^9.0.2",
     "multer": "^2.1.1",
+    "stripe": "^22.1.1",
     "zod": "^3.23.8"
   }
 }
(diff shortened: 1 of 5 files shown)


surim0n:
Reviewer demo note for the bounty requirement:

This PR can be demonstrated without any live Stripe credentials by running:

```bash
npm test -w apps/api
```

Current result: 5 passing tests, 1 guarded live Stripe smoke test skipped by default. The tests show the old fake timestamp ID path is gone, amount validation happens before any Stripe call, currency defaults to `usd`, metadata is passed through, Stripe `id`/`client_secret` are mapped to `paymentId`/`clientSecret`, and original Stripe error messages are preserved.

For a live test-mode demo, set `STRIPE_LIVE_SMOKE=true` with a test-mode `STRIPE_SECRET_KEY`; the guarded smoke test will create a real PaymentIntent. I did not include a real key or run the live path here, because putting payment credentials into public CI/comments would be a spectacularly dumb way to win a bounty.

Issue #766, opened 2026-05-27
fix(api): npm test fails on Node 22+ — broken --test directory arg in apps/api/package.json


Islam0953:
## Summary

`npm test` (and `node --test src/tests`) fails immediately on Node 22+ with `MODULE_NOT_FOUND` because the test script in `apps/api/package.json` points the test runner at a directory path that the runner no longer accepts as a discovery root.

## Reproduction

```bash
npm install
npm test
```

Result on Node 22 / 24:

```
node:internal/modules/cjs/loader:1424
  throw err;
  ^
Error: Cannot find module '/.../apps/api/src/tests'
    at Module._resolveFilename (...)
    ...
  code: 'MODULE_NOT_FOUND'
```

The `src/tests/health.test.js` test passes when run directly via a glob — the only thing broken is the script wiring.

## Root cause

`apps/api/package.json` declares:

```json
"test": "node --test src/tests"
```

The bare directory argument is interpreted by the runner as a single module to load (instead of a directory to scan), so Node tries to `require('src/tests')`, which does not exist as a file.

## Fix

Use the supported glob pattern so Node's built-in test runner picks up `*.test.js` files under `src/tests/` recursively:

```json
"test": "node --test \"src/tests/**/*.test.js\""
```

With that change, `npm test` reports:

```
✔ GET /health returns ok payload
ℹ tests 1
ℹ pass 1
ℹ fail 0
```

## Scope

Single-line change in `apps/api/package.json`. No source code touched, no new dependencies, no behaviour change to passing tests.

## Reservation

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #775, opened 2026-05-27
fix: restrict admin metrics to admin role

### LINKED ISSUE #774: Admin metrics endpoint allows non-admin JWT roles
The admin metrics API currently uses `authMiddleware`, which only proves that the requester has a valid JWT. It does not check `req.user.role`, so a token signed for a regular `client` or `freelancer` role can still reach `/api/admin/metrics` and read administrative platform metrics.

Expected behavior: admin routes should reject authenticated non-admin users with a forbidden response and only allow JWTs whose payload has `role: "admin"`.

Suggested fix:
- add a focused role guard middleware for required roles;
- apply it to `adminRoutes` after JWT authentication;
- add regression coverage for missing token, non-admin token, and admin token behavior.

tiago918:
/claim #743

Fixes #774

## Summary
- add a reusable `requireRole` middleware for authenticated route role checks
- apply `requireRole("admin")` to `/api/admin/*` routes so client/freelancer JWTs cannot read admin metrics
- add focused regression tests for missing token, non-admin token, and admin token access

## Demo
https://github.com/tiago918/bug-bounty/releases/download/securebanana-pr774-demo-20260527T1134Z/securebanana-pr774-demo.mp4

## Validation
```bash
git diff --check
ffprobe -v error -show_entries stream=codec_name,width,height,duration -of default=noprint_wrappers=1 /tmp/securebanana-pr774-demo.mp4
```

Result:
```text
4 tests passed
0 failed
codec_name=h264
width=1280
height=720
duration=14.000000
```

Note: `npm test -w apps/api` is currently affected by the already-filed Node 22+ test directory issue #766, so validation calls the test files directly.

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,15 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function requireRole(...allowedRoles) {
+  const allowed = new Set(allowedRoles);
+
+  return function requireRoleMiddleware(req, res, next) {
+    if (!req.user || !allowed.has(req.user.role)) {
+      return fail(res, "Forbidden", 403);
+    }
+
+    return next();
+  };
+}
--- a/apps/api/src/routes/adminRoutes.js
+++ b/apps/api/src/routes/adminRoutes.js
@@ -1,8 +1,9 @@
 import { Router } from "express";
 import { metrics } from "../controllers/adminController.js";
-import { authMiddleware } from "../middleware/auth.js";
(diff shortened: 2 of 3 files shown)


tiago918:
Validation note: the failing `update-leaderboard` check is failing inside the repository workflow while pushing its generated `chore: update leaderboard for PR #775` commit to `main`:

```text
remote: fatal error in commit_refs
! [remote rejected] HEAD -> main (failure)
```

That job does not run the API tests. Local validation for this PR passed with `node --test apps/api/src/tests/*.test.js` (4 passed, 0 failed) and `git diff --check`.

PR #784, opened 2026-05-27
fix: validate refresh tokens before reissuing access

### LINKED ISSUE #783: Refresh endpoint issues access tokens without validating a refresh token
The refresh endpoint currently returns a new access token without requiring or validating any refresh token input.

## Current behavior

`POST /api/auth/refresh` calls `refreshToken()` with no request data, and the service always returns a signed client access token:

~~~js
  return { token: signAccessToken({ sub: "usr_existing", role: "client" }) };
}
~~~

That means any caller can hit the refresh route and receive a valid access token, even with an empty JSON body.

## Expected behavior

- `POST /api/auth/refresh` should require a refresh token value.
- Missing, invalid, expired, or access-token values should be rejected with `401`.
- Login/register should issue a refresh token that can be used to mint a new access token.
- The refreshed access token should preserve the authenticated subject and role from the refresh token.

## Suggested fix

Add a separate refresh-token signer/verifier, return refresh tokens from login/register, and update the refresh controller to validate `req.body.refreshToken` before issuing a new access token. Add regression tests for missing refresh tokens, access-token misuse, and a valid refresh-token flow.

iccccccccccccc:
Fixes #783
/claim #743

## Changes
- add a real refresh-token type instead of letting the refresh route mint access tokens with no input
- return refresh tokens from login/register so the refresh flow has a valid credential to use
- reject missing, invalid, or access-token values on `POST /api/auth/refresh`
- add focused API coverage for the missing-token, wrong-token, and valid refresh-token cases

## Verification
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/auth.test.js`
- `node --check apps/api/src/utils/jwt.js`
- `node --check apps/api/src/tests/auth.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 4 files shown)


iccccccccccccc:
Added a short demo GIF for the bounty claim:

refresh-token validation demo

It shows the focused API checks for the refresh endpoint: missing refresh token, access-token misuse, valid refresh-token flow, plus the syntax checks for the touched files.

PR #785, opened 2026-05-27
fix: require auth for refresh endpoint

### LINKED ISSUE #782: Refresh endpoint mints access tokens without authentication
## Bug

`POST /api/auth/refresh` currently calls `refreshToken()` without any authentication middleware or request-derived identity. Any caller can hit the endpoint and receive a valid access token for the hard-coded `usr_existing` client identity.

## Impact

This turns the refresh endpoint into an unauthenticated token minting route. Even in a stubbed API, refresh should require a valid bearer token or refresh credential and should preserve the authenticated user's identity instead of issuing a hard-coded token.

## Expected fix

- Protect `POST /api/auth/refresh` with the existing auth middleware.
- Build the refreshed token from `req.user.sub` and `req.user.role`.
- Add regression tests proving unauthenticated refresh returns 401 and authenticated refresh preserves the caller identity.

adliebe:
/claim #743

## Summary

Fixes #782 by requiring authentication before `POST /api/auth/refresh` can mint a new access token.

## What changed

- Applies the existing `authMiddleware` to `POST /api/auth/refresh`.
- Passes the authenticated `req.user` identity into the refresh service.
- Signs refreshed tokens from the caller's `sub` and `role` instead of hard-coded `usr_existing` / `client` values.
- Adds regression tests for unauthenticated rejection and authenticated identity preservation.

## Validation

```text
cd apps/api

3 passed
0 failed
```

## Demo

Short demo GIF: https://github.com/adliebe/bug-bounty/releases/download/securebanana-pr782-demo-20260527T090214/securebanana-pr782-demo-20260527T090214.gif

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const result = await refreshToken(req.user);
   return ok(res, result);
 }
--- a/apps/api/src/routes/authRoutes.js
+++ b/apps/api/src/routes/authRoutes.js
@@ -1,4 +1,5 @@
 import { Router } from "express";
+import { authMiddleware } from "../middleware/auth.js";
 import { login, oauthCallback, refresh, register } from "../controllers/authController.js";
 
 export const authRoutes = Router();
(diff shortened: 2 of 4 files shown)


adliebe:
Closing this to keep the bounty thread clean. This overlaps with the now-active refresh-token PR #784, and this account already has a separate #743 claim in #771. Thanks for the review bandwidth.

PR #833, opened 2026-05-27
fix: pin jwt access token algorithm

### LINKED ISSUE #832: Pin JWT access tokens to HS256
## Problem

`verifyAccessToken()` currently calls `jwt.verify(token, env.jwtSecret)` without constraining accepted algorithms. With `jsonwebtoken`, a token signed with another HMAC algorithm such as HS512 and the same shared secret verifies successfully, even though the app's access-token format should stay pinned to one expected algorithm.

## Why this matters

Keeping the accepted JWT algorithm explicit reduces the chance of future algorithm confusion as authentication code evolves, and it makes the token contract clear for both signing and verification.

## Proposed fix

- Sign access tokens with an explicit HS256 algorithm.
- Verify access tokens with `algorithms: ["HS256"]`.
- Add focused utility tests that prove normal app tokens verify and alternate HMAC algorithms are rejected.

dylanbuilds:
/claim #743

Fixes #832.

## Summary

- Pin access-token signing to HS256.
- Restrict access-token verification to HS256 so alternate HMAC algorithms such as HS512 are rejected.
- Add focused JWT utility tests for the accepted app-token path and the rejected unexpected-algorithm path.
- Add `demos/jwt-algorithm-demo.mp4` as the short demo artifact requested by the bounty instructions.

## Validation

Re-run on May 27, 2026 after the demo artifact update:

- `node --test src/tests/jwt.test.js` — 2 tests passed, 0 failed.
- `node --test src/tests/*.test.js` — 3 tests passed, 0 failed.
- `git diff --check` — passed with no whitespace errors.
- `ffprobe -v error -show_entries format=duration,size -of default=noprint_wrappers=1 demos/jwt-algorithm-demo.mp4` — duration `6.000000`, size `64832` bytes.

GitHub currently reports no configured checks for this branch, so the PR-local validation above is the proof trail.

Payment details intentionally omitted from public GitHub comments/PR text.


dylanbuilds:
Updated the demo artifact at `demos/jwt-algorithm-demo.mp4` so it shows the actual validation commands/results rather than only a summary slide.

Re-validated locally after the update:

- `node --test src/tests/jwt.test.js`
- `git diff --check`
- `ffprobe -v error -show_entries format=duration,size -of default=noprint_wrappers=1 demos/jwt-algorithm-demo.mp4`

PR #982, opened 2026-05-27
fix: validate user creation payload — reject empty email/role

### LINKED ISSUE #979: [BUG] User creation accepts empty payloads and client-controlled IDs
## DescriptionPOST /api/users accepts empty request bodies and allows clients to control user fields without any server-side validation. This results in:- Users created with no email, no role, no meaningful data- Client-controlled id field injection possible- No validation that required fields (email, role) are present## Steps to Reproduce```bashcurl -X POST http://localhost:4000/api/users -H "Content-Type: application/json" -d "{}"```Expected: 400 Bad RequestActual: 201 Created with `{"id":"usr_xxx"}`## Proposed FixAdd input validation in userController.postUser to reject empty/missing required fields.This issue is limited only to the creator of this issue./bounty $50

a2867966414:
Fixes #979

Changes:
- Added validation for email and role fields
- Returns 400 for empty payloads
- Prevents client-controlled field injection

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,1 @@
-import { ok } from "../utils/response.js";
-import { createUser, listUsers } from "../services/userService.js";
-
-export async function getUsers(req, res) {
-  return ok(res, await listUsers());
-}
-
-export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
-}
+import { ok } from "../utils/response.js";import { createUser, listUsers } from "../services/userService.js";export async function getUsers(req, res) {  return ok(res, await listUsers());}export async function postUser(req, res) {  const { email, role } = req.body || {};  if (!email || !role) {    return res.status(400).json({ error: "email and role are required" });  }  return ok(res, await createUser({ email, role }), 201);}


MyTH-zyxeon:
Thanks for the quick fix on #979. I did a small review pass against the bounty description and there is one merge-risk worth fixing before maintainers review this.

The intended validation is there, but the patch rewrites `apps/api/src/controllers/userController.js` into a single CRLF-packed line. That makes the diff show `+1/-10` and hides the actual behavior change, even though this should be a tiny controller edit. I would reformat the file back to the existing multi-line style and keep the behavioral diff limited to `postUser`.

Suggested shape:

```js
  const { email, role } = req.body || {};

  if (!email || !role) {
    return res.status(400).json({ error: "email and role are required" });
  }

  return ok(res, await createUser({ email, role }), 201);
}
```

This preserves the good parts of the fix:

- rejects `{}` / missing `email` / missing `role` with 400
- prevents client-controlled `id` or unrelated fields from being passed through
- keeps `getUsers` and imports unchanged

One extra validation note: if this repo has controller or API route tests, a small regression test for `{}` and `{ email, role, id: "client" }` would make the bounty acceptance much easier. No duplicate implementation PR from me; this is just review assistance for the existing claimed fix.

MyTH-zyxeon:
Follow-up review-assist for #979 / PR #982 after rechecking the current `eccf08d...` head.

The behavior change still looks right, but the current diff is still packed into one CRLF-style line, which makes this harder to review than it needs to be. A minimal follow-up could replace `apps/api/src/controllers/userController.js` with the same existing structure plus only the `postUser` validation change:

```js
import { ok } from "../utils/response.js";
import { createUser, listUsers } from "../services/userService.js";

export async function getUsers(req, res) {
  return ok(res, await listUsers());
}

  const { email, role } = req.body || {};

  if (!email || !role) {
    return res.status(400).json({ error: "email and role are required" });
  }

  return ok(res, await createUser({ email, role }), 201);
}
```

That keeps the bounty fix focused on:

- rejecting `{}` / missing `email` / missing `role` with 400
- passing only server-approved fields into `createUser`
- preserving the existing readable file style so maintainers can inspect the actual controller behavior

If there is a test harness in this repo, the most valuable regression would be one POST `/users` empty-payload case and one valid payload case that proves `id` is not forwarded from the request body.

PR #997, opened 2026-05-28
Reject review self-targeting

### LINKED ISSUE #996: Review creation allows users to review themselves
## Bug Description

`POST /api/reviews` accepts review payloads where `reviewerId` and `revieweeId` are the same user. In a freelance marketplace, self-reviews let an account inflate its own trust/reputation history and pollute review data that should come from another participant.

## Steps to Reproduce

1. Start the API app.
2. Send a review creation request with the same participant on both sides:

```json
{
  "reviewerId": "usr_same",
  "revieweeId": "usr_same",
  "rating": 5,
  "comment": "self boost"
}
```

3. Observe that the service creates and stores the review instead of rejecting it.

## Expected Behavior

Review creation should reject self-reviews with a client error before storing anything. `reviewerId` and `revieweeId` must identify different users.

## Actual Behavior

The review service stores the payload and returns a successful review object with matching `reviewerId` and `revieweeId`.

## Proposed Fix

Add a focused guard in the review creation path that returns HTTP 400 when `reviewerId === revieweeId`, and add API regression tests for both the rejection path and a normal different-user review.

partyplatter08-lab:
/claim #743

Closes #996.

## Summary

- Reject review creation when `reviewerId` and `revieweeId` identify the same user.
- Keep the fix narrowly scoped so existing broader review validation work remains untouched.
- Harden the guard to compare provided reviewer/reviewee IDs directly instead of only handling string IDs.
- Add API regression coverage for the rejected self-review path, proof that rejected self-reviews are not stored, and the normal different-user path.
- Include a short demo video at `demos/review-self-review-demo.mp4`.

## Validation

- `node --test apps/api/src/tests/review-self-review.test.js` — 2 passed
- `node --test apps/api/src/tests/*.test.js` — 3 passed
- `ffprobe -v error -select_streams v:0 -show_entries stream=codec_name,width,height,duration -of default=noprint_wrappers=1 demos/review-self-review-demo.mp4`
- `git diff --check`

## Notes

`npm test` still uses the existing `node --test src/tests` directory invocation tracked in #766, so I used the working Node test glob directly for validation.


partyplatter08-lab:
Withdrawing this PR; I am no longer pursuing this change.

PR #1032, opened 2026-05-28
Reject upload requests without a file

### LINKED ISSUE #1029: Upload endpoint returns success when no file is provided
This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743.

## Bug

`POST /api/uploads` currently returns a 201 success response even when the request does not include a `file` field. The response body reports `status: "no-file"` and `filename: null`, so clients can treat a failed upload as a successful one.

## Reproduction

1. Start the API app.
2. Send `POST /api/uploads` without a multipart `file` field.
3. Actual result: 201 success with `{ filename: null, status: "no-file" }`.
4. Expected result: reject the request with 400 and a clear error message.

## Proposed fix

- Return 400 when `req.file` is missing.
- Keep the 201 success response only for real uploaded files.
- Add node:test coverage for both the missing-file rejection and successful file upload path.

Referencing #743.

youngbum8973-coder:
/claim #743

Closes #1029.

## Summary

- Return a 400 error when `POST /api/uploads` does not include a `file` field.
- Keep the 201 upload success response only for actual multipart file uploads.
- Add API tests covering the missing-file rejection and successful upload path.

## Verification

- `npm test -w apps/api`

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
-    status: req.file ? "uploaded" : "no-file"
+    filename: req.file.originalname,
+    status: "uploaded"
   }, 201);
 }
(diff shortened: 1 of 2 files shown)


youngbum8973-coder:
Adding the requested short demo for this upload validation change. It shows a missing-file upload request returning 400 while a real multipart file upload still succeeds.
clipboard

PR #1036, opened 2026-05-28
Handle async controller validation errors

### LINKED ISSUE #1035: API validation failures in async controllers are not handled and return server errors
﻿This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue Low Handing Fruit Automation #743 for more information.

## Problem

The Express API uses async controllers and validates request bodies with Zod via `.parse(...)`.
When validation fails inside an async controller, Express 4 does not automatically forward the rejected promise to the error middleware. Invalid client payloads can therefore escape the intended error response path and behave like server failures instead of returning a structured 400 response.

Repro example:

1. Run the API test server.
2. Send `POST /api/auth/register` with an invalid email and a password shorter than 8 characters.
3. The Zod validation error is not converted into a stable `400` response with field-level details.

## Expected behavior

Invalid request payloads should return:

- HTTP 400
- `success: false`
- a stable validation message
- field-level Zod issue details

Async controller rejections should also be forwarded to the existing `errorHandler`.

## Proposed fix

- Add a tiny `asyncHandler` wrapper that forwards rejected controller promises to Express `next`.
- Wrap API route controllers with `asyncHandler`.
- Teach `errorHandler` to convert `ZodError` into a 400 response.
- Add a regression test for invalid `/api/auth/register` payloads.

Jorel97:
Closes #1035
Refs #743

/claim #743

## Summary

- added an `asyncHandler` wrapper so Express 4 forwards async controller rejections to error middleware
- converted Zod validation failures into structured HTTP 400 responses
- wrapped API route controllers and added regression coverage for invalid auth registration payloads

## Demo / Validation

Short demo video:

https://github.com/Jorel97/bounty-demo-assets/raw/main/securebanana/securebanana-743-demo.mp4

Reproduction fixed by this PR:

1. Send `POST /api/auth/register` with an invalid email and a short password.
2. Before this change, the thrown Zod validation error was not converted into a stable client error path for async controllers.
3. After this change, the API responds with HTTP `400`, `success: false`, `message: "Invalid request payload"`, and field-level issues for `email` and `password`.

Local validation:

```text

tests 2
pass 2
fail 0
```

### DIFF
--- a/apps/api/src/middleware/asyncHandler.js
+++ b/apps/api/src/middleware/asyncHandler.js
@@ -1,0 +1,5 @@
+export function asyncHandler(handler) {
+  return function wrappedAsyncHandler(req, res, next) {
+    return Promise.resolve(handler(req, res, next)).catch(next);
+  };
+}
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,22 @@
+import { ZodError } from "zod";
+
 export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
 
+  if (err instanceof ZodError) {
+    return res.status(400).json({
+      success: false,
+      message: "Invalid request payload",
+      issues: err.issues.map((issue) => ({
+        path: issue.path.join("."),
+        message: issue.message
+      }))
+    });
+  }
+
+  console.error("Unhandled API error:", err);
   return res.status(500).json({
     success: false,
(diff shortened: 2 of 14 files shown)


Jorel97:
Additional note for reviewers: this is distinct from the existing errorHandler headers-sent fix. This PR addresses Express 4 async controller rejections and Zod validation failures: invalid payloads thrown from async routes are forwarded through error middleware and returned as structured HTTP 400 responses. The regression test covers POST /api/auth/register with invalid email/password and verifies the 400 validation response.

Issue #1076, opened 2026-05-28
Bug: User ID mismatch in registerUser function causes authentication token inconsistency


duongynhi000005-oss:
## Description

There is a critical bug in the `registerUser` function in `apps/api/src/services/authService.js` where the user ID is generated twice using `Date.now()`, resulting in different timestamps and causing a mismatch between the stored user ID and the JWT token's subject claim.

## Current Code

```javascript
export async function registerUser(payload) {
  // TODO: persist new user via Prisma
  return {
    id: `usr_${Date.now()}`,
    email: payload.email,
    role: payload.role,
    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
  };
}
```

## Problem

The function calls `Date.now()` twice:
1. Once for the user's `id` field
2. Once for the JWT token's `sub` (subject) claim

Since these calls happen at slightly different times (even if milliseconds apart), they will generate different values. This means:
- The user object will have ID: `usr_1234567890000`
- The JWT token will have sub: `usr_1234567890001` (or a different timestamp)

## Impact

- **Authentication failures**: When the token is verified and used to identify the user, the `sub` claim won't match the actual user ID in the database
- **Authorization issues**: Any middleware or service that tries to match the token's subject to a user record will fail
- **Security concern**: This creates an inconsistent authentication state

## Expected Behavior

The user ID should be generated once and reused for both the user object and the JWT token's subject claim.

## Proposed Fix

```javascript
export async function registerUser(payload) {
  const userId = `usr_${Date.now()}`;
  return {
    id: userId,
    email: payload.email,
    role: payload.role,
    token: signAccessToken({ sub: userId, role: payload.role })
  };
}
```

## Steps to Reproduce

1. Call the `/api/auth/register` endpoint
2. Examine the returned user ID and decode the JWT token
3. Compare the user's `id` field with the token's `sub` claim
4. Observe that they are different

## Environment

- File: `apps/api/src/services/authService.js`
- Function: `registerUser`
- Lines: 3-10

---

**This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.**

Related to: #743

PR #1089, opened 2026-05-28
Fix register token subject mismatch

### LINKED ISSUE #1093: registerUser response id can differ from JWT subject
Parent bounty: #743

Bug:
`registerUser()` currently generates the user id twice: once for the response `id` and once for the JWT `sub` claim. If `Date.now()` changes between those calls, the registration response and token identify different users.

Expected:
Generate the id once and reuse it for both the response and JWT subject.

PR:
#1089

youngbum8973-coder:
Fixes #1093
/claim #743

## Summary
- reuse one generated user id for both the registration response and JWT `sub` claim
- add a regression test that forces `Date.now()` to return different values across calls
- #1093 is the scoped issue I created for this fix under parent bounty #743

## Tests
- npm test

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -1,12 +1,14 @@
 import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
+  const id = `usr_${Date.now()}`;
+
   // TODO: persist new user via Prisma
   return {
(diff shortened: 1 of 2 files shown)


youngbum8973-coder:
Per #743 instructions, I created scoped issue #1093 for this work. The earlier #1076 was a similar prior report; this PR is intended to resolve #1093.

/claim #743

PR #1132, opened 2026-05-28
Hermes Agent [ BUG ] Return defensive copies from list services

simisdav55-oss:
All list service functions now return [...spread] copies of their backing store arrays so callers cannot mutate internal state. Covers users, jobs, proposals, reviews, messages, and notifications. Adds regression tests verifying mutation isolation.\n\nCloses #1108\nRefs #743


simisdav55-oss:
Maintainer ping: this PR is still open, CI is passing, and it is ready for review. Please review or close it if the bounty is no longer active.

Issue #1141, opened 2026-05-28
Bug: Zod validation errors return 500 Internal Server Error instead of 400 Bad Request


Ishant5436:
Parent bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Bug:
The global error handler (`apps/api/src/middleware/errorHandler.js`) currently intercepts `ZodError`s thrown by input validation schemas (such as in `authController.js`) and returns a generic `500 Internal Server Error` instead of a `400 Bad Request`.

Expected behavior:
When a request fails Zod validation, the API should return a `400 Bad Request` containing the specific validation errors to the client, rather than logging a generic unhandled server error and returning 500.

Root cause:
`errorHandler.js` does not check if `err` is an instance of `ZodError`. All thrown errors are treated as unexpected server errors.

Proposed fix:
Update `errorHandler.js` to import `ZodError` from `zod` and check `err instanceof ZodError`. If true, return `400` with the formatted errors. Add a unit test to verify this behavior.

wbo68410-lgtm:
Opened PR #1143 to return Zod validation failures as 400 responses and route async controller rejections through the shared error handler: https://github.com/SecureBananaLabs/bug-bounty/pull/1143

Ishant5436:
Closing this issue because SecureBananaLabs is operating a deceptive bounty program that avoids payouts.

PR #1144, opened 2026-05-28
fix: zod validation errors return 400

Ishant5436:
This PR fixes the global error handler so that it intercepts `ZodError` exceptions and returns a `400 Bad Request` with structured validation messages, rather than throwing an unhandled `500 Internal Server Error`.

Scope:
- `errorHandler.js`: Intercept `ZodError` and map it to a 400 response.
- `authController.js`: Wraps endpoints with try/catch to ensure async validation errors correctly bubble up to the global error handler in Express.
- `errorHandler.test.js`: Comprehensive node:test coverage validating the behavior.

This resolves issue #1141.

/claim #743


Ishant5436:

PR #1189, opened 2026-05-28
Reject non-positive payment amounts

### LINKED ISSUE #1177: Payment intent accepts non-positive amounts
Parent bounty: #743

Bug:
POST /api/payments currently accepts zero or negative payment amounts and returns a successful payment intent instead of rejecting invalid client input.

Expected behavior:
Payment intent creation should reject non-positive or non-numeric amounts with a 400 Bad Request response while preserving the existing 201 response for positive numeric amounts.

Root cause:
apps/api/src/services/paymentService.js passes req.body.amount through directly and the payment controller always wraps the result in a success response.

Proposed fix:
Validate that amount is a finite number greater than zero before creating the intent, return a 400 error for invalid amounts, and add focused API tests for the rejection and success paths.

youngbum8973-coder:
/claim #743

Closes #1177.

Summary
- Reject payment intent requests when amount is missing, non-numeric, zero, or negative.
- - Return a 400 response for invalid payment amounts instead of a successful intent payload.
- - Add focused controller tests for invalid and valid payment amounts.
Verification
- npm test -w apps/api

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+    const result = await createPaymentIntent(req.body);
+
+  if (result.error) {
+        return fail(res, result.error, 400);
+  }
+
+  return ok(res, result, 201);
 }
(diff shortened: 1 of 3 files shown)


youngbum8973-coder:
Adding validation evidence for this payment amount change:

What I verified locally:
- Missing, non-numeric, zero, and negative payment amounts now return HTTP 400 instead of creating a payment intent payload.
- A positive numeric amount still follows the valid payment intent path.
- `npm test -w apps/api` passes with the focused payment controller tests included.

This keeps invalid payment requests from being accepted as successful intent responses.

Issue #1333, opened 2026-05-28
Bug: Mass Assignment allows Privilege Escalation during Registration


Ishant5436:
Bug: Mass Assignment allows Privilege Escalation to Admin during Registration

The `registerSchema` inside `apps/api/src/validators/auth.js` explicitly permits the `role` field to be set to `"admin"` by anyone hitting the `POST /api/auth/register` endpoint. Since this user-supplied payload is passed directly into `registerUser` which signs the `role` into the generated JWT, an attacker can trivially create a fully authorized administrative account simply by passing `"role": "admin"` in their registration JSON.

This completely bypasses the RBAC mechanisms and grants the attacker full access to all `/api/admin` endpoints.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Ishant5436:
/attempt #743 - I am working on this issue.

Ishant5436:
Closing this issue because SecureBananaLabs is operating a deceptive bounty program that avoids payouts.

PR #1335, opened 2026-05-28
fix: block admin privilege escalation during registration

Ishant5436:
This PR fixes a critical Insecure Direct Object Reference (IDOR) / Mass Assignment vulnerability in the user registration flow.

Scope:
- `validators/auth.js`: The `registerSchema` incorrectly allowed the `role` enum to include `"admin"`. This allowed any anonymous user to trivially create a fully authorized administrative account by simply injecting `"role": "admin"` into the registration JSON payload. I have removed `"admin"` from the allowed Zod schema enum so that users can only register as `"client"` or `"freelancer"`.
- `authRegistration.test.js`: Added a TDD node:test suite to assert that attempting to register with the `"admin"` role is successfully rejected during validation.

This ensures the RBAC system cannot be bypassed via mass assignment during registration.

This resolves issue #1333.

/claim #743

### DIFF
--- a/apps/api/src/tests/authRegistration.test.js
+++ b/apps/api/src/tests/authRegistration.test.js
@@ -1,0 +1,38 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
+async function withServer(fn) {
+  const app = createApp();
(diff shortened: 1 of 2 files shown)


Ishant5436:

Issue #1340, opened 2026-05-28
Bug: Critical Authentication Bypass on Login


Ishant5436:
Bug: Critical Authentication Bypass on Login Endpoint

The `loginUser` service currently completely skips password verification due to an unimplemented `TODO`. It blindly accepts any email address and instantly issues a valid JWT token. This allows any anonymous internet user to log into ANY account simply by knowing their email address, representing a complete failure of the authentication mechanism (Authentication Bypass).

To resolve this while the application is still using in-memory mock data, the `loginUser` function must be updated to at least query the `users` array (from `userService.js`) and perform a strict check that `user.password === payload.password`. If the user does not exist or the password does not match, it must throw an `Invalid credentials` error.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Ishant5436:
/attempt #743 - I am working on this issue.

Ishant5436:
Closing this issue because SecureBananaLabs is operating a deceptive bounty program that avoids payouts.

PR #1341, opened 2026-05-28
fix: resolve critical authentication bypass on login endpoint

Ishant5436:
This PR fixes a critical Authentication Bypass vulnerability in the login and token refresh flow.

Scope:
- `services/authService.js`: The `loginUser` endpoint had an unimplemented `TODO` and would blindly issue valid JWT tokens for any email address provided in the payload, without validating the password. I have replaced this with a check against the `users` array (from `userService.js`), ensuring that `user.password === payload.password` before issuing the token. It throws an `Invalid credentials` error if they don't match.
- `services/authService.js`: The `refreshToken` function was also blindly issuing tokens for `"usr_existing"`. It has been updated to throw an error for now since refresh tokens are not implemented.
- `tests/authBypass.test.js`: Added a test to verify that attempting to log in with fake credentials successfully fails instead of yielding a `200 OK` response.

This effectively closes the authentication bypass during the mock/in-memory phase.

This resolves issue #1340.

/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -11,6 +11,9 @@ export async function register(req, res) {
 export async function login(req, res) {
   const payload = loginSchema.parse(req.body);
   const result = await loginUser(payload);
+  if (!result) {
+    return res.status(401).json({ success: false, message: "Invalid credentials" });
+  }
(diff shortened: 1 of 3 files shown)


Ishant5436:

Issue #1342, opened 2026-05-28
Bug: Critical Information Disclosure on GET /api/users


Ishant5436:
Bug: Critical Information Disclosure - User Passwords and PII exposed via GET /api/users

The `GET /api/users` endpoint returns the entire `users` array in memory. When a user registers via `postUser` (which directly pushes `req.body` into the array), sensitive information including their email, role, and plain-text password is appended to their user object. Because `GET /api/users` lacks authentication, any anonymous attacker can fetch a complete list of all registered users, including their credentials, exposing the platform to total compromise.

To fix this, `GET /api/users` must be secured with `authMiddleware`, and the `listUsers` service (or the controller) must aggressively sanitize the user objects to strip out passwords before returning them to the client.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Ishant5436:
/attempt #743 - I am working on this issue.

Ishant5436:
Closing this issue because SecureBananaLabs is operating a deceptive bounty program that avoids payouts.

PR #1343, opened 2026-05-28
fix: prevent information disclosure of user passwords and PII

Ishant5436:
This PR fixes a critical Information Disclosure and PII Leak vulnerability on the user discovery API.

Scope:
- `routes/userRoutes.js`: Applied `authMiddleware` to `GET /api/users` so that only authenticated users can browse the list of platform users.
- `controllers/userController.js`: When a new user registers, their plaintext password was previously being appended to the `users` array in memory and subsequently leaked out to anyone fetching the list of users. I have aggressively sanitized the payload returned from `GET /api/users` by mapping over the array and stripping out the `password` field from the response.
- `piiLeak.test.js`: Added a test to verify that the route is properly protected with a 401 and does not expose passwords to anonymous requests.

This resolves issue #1342.

/claim #743

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -2,7 +2,9 @@ import { ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
 
 export async function getUsers(req, res) {
-  return ok(res, await listUsers());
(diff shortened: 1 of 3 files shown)


Ishant5436:

Issue #1346, opened 2026-05-28
Bug: Default JWT Secret fallback allows token forgery in production


Ishant5436:
Bug: Default JWT Secret fallback allows token forgery in production

In `apps/api/src/config/env.js`, the `jwtSecret` falls back to the hardcoded string `"development-secret"` if `JWT_SECRET` is not provided in the environment variables. 
If the API is deployed to production and the environment variable is accidentally omitted, attackers can easily forge valid administrator JWT tokens and achieve remote code execution or full platform takeover.

To fix this, the application should strictly enforce that `JWT_SECRET` is provided, or at least throw a startup error if `NODE_ENV === 'production'` and the secret is missing or set to the default.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Ishant5436:
/attempt #743 - I am working on this issue.

Ishant5436:
Closing this issue because SecureBananaLabs is operating a deceptive bounty program that avoids payouts.

PR #1347, opened 2026-05-28
fix: enforce secure JWT_SECRET in production

Ishant5436:
This PR fixes a critical security misconfiguration where `JWT_SECRET` falls back to a hardcoded string `development-secret` if not explicitly set in the environment variables. 
If deployed to production with a missing environment variable, this default allows attackers to trivially forge valid access tokens and achieve full system compromise.

Scope:
- `config/env.js`: Now strictly checks `NODE_ENV === "production"` and throws a startup error if `JWT_SECRET` is missing.
- `env.test.js`: Added a test to verify that the application refuses to start in production without a secure secret.

This resolves issue #1346.

/claim #743


Ishant5436:

PR #1385, opened 2026-05-28
fix: resolve postcss XSS vulnerability via overrides

Ishant5436:
This PR resolves a critical Cross-Site Scripting (XSS) vulnerability (GHSA-qx2v-qp2m-jg93) in `postcss < 8.5.10`.

Next.js `16.2.6` bundles `postcss@8.4.31`, which is vulnerable. Instead of performing a breaking downgrade to Next.js `9.3.3` as `npm audit fix` suggests, we enforced an override to use `postcss@8.5.15` across the monorepo, resolving the vulnerability while keeping dependencies up to date.

/claim #952

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -154,6 +154,9 @@
       "cpu": [
         "arm"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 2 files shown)


Ishant5436:
Closing: Repository violates payout terms and maintains deceptive leaderboard automation.

PR #1386, opened 2026-05-28
fix: prevent JWT secret fallback to weak default in production

Ishant5436:
This PR resolves a critical vulnerability where the application would silently fallback to a hardcoded JWT secret (`development-secret`) if `process.env.JWT_SECRET` was undefined in production. This allows an attacker who knows the open-source codebase to trivially forge JWT tokens and escalate privileges on production systems.

The fix enforces that the environment variable must be provided in production, throwing a loud error at startup rather than compromising security silently.

/claim #952


Ishant5436:
Closing: Repository violates payout terms and maintains deceptive leaderboard automation.

Issue #1102, opened 2026-05-28
Admin metrics endpoint is exposed without admin role authorization


wbo68410-lgtm:
Parent bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Bug:
`GET /api/admin/metrics` requires a valid bearer token, but it does not require the token to belong to an admin. A non-admin user with a `client` or `freelancer` role can request internal platform metrics such as open job counts, active freelancer counts, flagged account counts, and monthly volume.

Expected:
Admin metrics should require a valid bearer token whose role is `admin`. Missing tokens, invalid tokens, and non-admin roles should be rejected before metrics are returned.

Root cause:
`apps/api/src/routes/adminRoutes.js` applies `authMiddleware`, but there is no reusable role guard after JWT verification.

Proposed fix:
- Add a small `requireRole("admin")` middleware next to the existing auth middleware.
- Mount `/api/admin/metrics` behind `authMiddleware` and the admin role guard.
- Add API regression tests for unauthenticated, non-admin, and admin requests.

Non-overlap:
This is scoped only to the admin metrics endpoint and admin role authorization. It does not attempt to secure every API route or build the broader admin panel.

wbo68410-lgtm:
Opened PR for #1102: https://github.com/SecureBananaLabs/bug-bounty/pull/1104

Verification:
- 
pm test -w apps/api`n- git diff --check`n
Refs #743

PR #1418, opened 2026-05-29
Restrict admin metrics to admin users

### LINKED ISSUE #1417: Admin metrics endpoint accepts non-admin tokens

## Bug
The admin metrics endpoint only checks that a request has any valid JWT. It does not verify that the authenticated user has the `admin` role, so a normal client token can access `/api/admin/metrics` and read internal platform metrics.

## Reproduction
1. Obtain a normal client token, for example through the existing login flow where `loginUser` signs a token with `role: client`.
2. Send `GET /api/admin/metrics` with `Authorization: Bearer <client token>`.
3. The endpoint returns the admin metrics payload instead of rejecting the request.

## Expected behavior
Admin routes should reject authenticated non-admin users with a 403 response and only allow users whose token role is `admin`.

Parent bounty: #743

Jorel97:
/claim #743

## Summary
- Adds reusable role authorization middleware after JWT authentication.
- Protects `/api/admin/metrics` so authenticated non-admin users receive `403 Forbidden`.
- Adds regression tests for missing token, client token, and admin token behavior.

Fixes #1417.

## Demo
Short demo video: https://github.com/Jorel97/bounty-demo-assets/raw/main/securebanana/securebanana-1417-demo.mp4

## Validation
- `node --check apps/api/src/middleware/auth.js`
- `node --check apps/api/src/routes/adminRoutes.js`
- `node --check apps/api/src/tests/adminAuth.test.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/adminAuth.test.js`
- `git diff --check`

Note: `npm test -w apps/api` could not be used in this local Windows environment because the workspace script invokes `node` through `cmd` and returns `Acceso denegado`; the same test files pass with the bundled Node binary directly.

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,13 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function requireRole(...roles) {
+  return function roleMiddleware(req, res, next) {
+    if (!req.user || !roles.includes(req.user.role)) {
+      return fail(res, "Forbidden", 403);
+    }
+
+    return next();
+  };
+}
--- a/apps/api/src/routes/adminRoutes.js
+++ b/apps/api/src/routes/adminRoutes.js
@@ -1,8 +1,9 @@
 import { Router } from "express";
 import { metrics } from "../controllers/adminController.js";
-import { authMiddleware } from "../middleware/auth.js";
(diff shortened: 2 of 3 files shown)


Jorel97:
Closing this as a duplicate after finding earlier reports/PRs for the same admin metrics role-authorization issue (#774/#775 and #1102/#1104). I do not want to add duplicate review noise for maintainers.

PR #1522, opened 2026-05-29
fix: prevent mass assignment in user creation (#1520)

### LINKED ISSUE #1520: Bug: createUser mass assignment allows setting arbitrary user fields including role
## Bug

`POST /api/users` calls `createUser(req.body)` which spreads `...payload` directly into the user object. This allows callers to set any field, including `role: "admin"`, `isAdmin: true`, or other sensitive properties.

## Reproduction

1. Start the API server
2. Send `POST /api/users` with body:
```json
{"name": "attacker", "email": "evil@test.com", "role": "admin", "isAdmin": true, "balance": 999999}
```
3. The created user object will contain ALL provided fields, including privileged ones

## Expected behavior

The endpoint should only accept whitelisted fields (name, email, password) and reject or ignore any additional fields.

## Impact

- Privilege escalation: any user can make themselves admin
- Data injection: arbitrary fields stored in user records
- No input validation on user creation

## Proposed fix

- Use a Zod schema to validate and strip unknown fields
- Only allow safe fields: name, email, password
- Add tests for field restriction

Refs #743.

zeroknowledge0x:
## Summary
Prevents mass assignment vulnerability in `POST /api/users` by adding strict Zod validation that only allows safe fields.

## Changes
- Added `validators/user.js` with `createUserSchema` (name, email, password only)
- Updated `userController.js` to validate input through the schema
- Added `user-mass-assignment.test.js` with 2 tests:
  - Verifies unknown fields (role, isAdmin, balance) are stripped
  - Verifies missing required fields return 400

## Problem
Previously, `createUser(req.body)` spread ALL request body fields directly into the user object. An attacker could send `{"role": "admin", "isAdmin": true, "balance": 999999}` and those fields would be stored.

## Testing
- Unknown fields stripped ✅
- Missing fields rejected with 400 ✅

Fixes #1520
Refs #743

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,12 @@
 import { ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
+  const payload = createUserSchema.parse(req.body);
+  return ok(res, await createUser(payload), 201);
 }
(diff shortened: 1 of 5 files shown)


zeroknowledge0x:

PR #1533, opened 2026-05-29
fix: require JWT_SECRET in production, warn in development (#1532)

### LINKED ISSUE #1532: Bug: JWT secret falls back to hardcoded value when env var missing
## Bug

In `config/env.js` line 4:
```js
```

If `JWT_SECRET` is not set, the application silently uses `"development-secret"` as the signing key. In production, this means all JWT tokens are signed with a publicly known secret.

## Reproduction

1. Start the API without setting `JWT_SECRET` env var
2. Register a user — a JWT is returned
3. Decode the JWT at jwt.io — it will verify with secret "development-secret"
4. Forge any JWT with role "admin" using the known secret

## Expected behavior

The application should refuse to start if `JWT_SECRET` is not set in production, or generate a cryptographically random secret at startup (with a warning).

## Impact

- Complete authentication bypass if env var is missing
- Attacker can forge tokens for any user/role
- Silent failure — no warning or error

## Proposed fix

- Throw an error at startup if `JWT_SECRET` is not set
- Only allow the fallback in development mode
- Add test verifying startup fails without JWT_SECRET in production

Refs #743.

zeroknowledge0x:
## Summary
Prevents silent use of hardcoded JWT secret in production by requiring JWT_SECRET env var.

## Changes
- Added getJwtSecret() function to config/env.js
- Throws Error if env var missing in production
- Logs warning when using default secret in development
- Added test file jwt-secret-fallback.test.js

## Problem
Previously, jwtSecret used a publicly known fallback secret if the env var was not set. In production, this means anyone could forge JWT tokens for any user or role.

## Fix behavior
- Production (NODE_ENV=production): throws error at startup, app won't start
- Development: uses fallback with console warning
- With JWT_SECRET set: uses the env var value (no change)

Fixes #1532
Refs #743

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,17 @@
+function getJwtSecret() {
+  const secret = process.env.JWT_SECRET;
+  if (secret) return secret;
+  if (process.env.NODE_ENV === "production") {
+    throw new Error("JWT_SECRET must be set in production");
+  }
+  console.warn("WARNING: Using default JWT_SECRET — set JWT_SECRET env var for production");
+  return "development-secret";
+}
+
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
+  jwtSecret: getJwtSecret(),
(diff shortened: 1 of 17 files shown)


zeroknowledge0x:

PR #1641, opened 2026-05-29
fix: admin routes should require admin role, not just authentication

### LINKED ISSUE #1640: fix: admin routes should require admin role, not just authentication
## Bug

The admin routes (`/api/admin/metrics`) use `authMiddleware` to verify the JWT token, but they never check whether the authenticated user has the `admin` role. Any authenticated user — including `client` or `freelancer` — can access admin-only endpoints.

### Reproduction

```bash
# Register as a regular client
TOKEN=$(curl -s -X POST http://localhost:3000/api/auth/register \
  -d '{"email":"user@test.com","password":"securepass123","role":"client"}' \
  | jq -r '.data.token')

# Access admin metrics with client token
curl -H "Authorization: Bearer $TOKEN" http://localhost:3000/api/admin/metrics
```

**Actual behavior:** Returns admin metrics (open jobs, active freelancers, flagged accounts, monthly volume).
**Expected behavior:** Returns 403 Forbidden.

### Root Cause

In `apps/api/src/routes/adminRoutes.js`, only `authMiddleware` is applied. It verifies the token is valid but does not check `req.user.role === "admin"`.

### Proposed Fix

Add a `requireAdmin` middleware that checks `req.user.role` after authentication:

```js
// middleware/admin.js
export function requireAdmin(req, res, next) {
  if (req.user?.role !== "admin") {
    return fail(res, "Forbidden: admin access required", 403);
  }
  return next();
}
```

And apply it in the admin routes after `authMiddleware`.

### Scope

/bounty $700

zeroknowledge0x:
## Summary

Add `requireAdmin` middleware to admin routes so that only users with `role: "admin"` can access admin-only endpoints.

## Bug

The admin routes (`/api/admin/metrics`) use `authMiddleware` to verify the JWT token, but never check whether the authenticated user has the `admin` role. Any authenticated user — including `client` or `freelancer` — can access admin-only endpoints.

## Changes

- **`apps/api/src/middleware/admin.js`** (new): `requireAdmin` middleware that checks `req.user.role === "admin"` and returns 403 Forbidden if not
- **`apps/api/src/routes/adminRoutes.js`**: Added `requireAdmin` middleware after `authMiddleware`
- **`apps/api/src/tests/admin-role-check.test.js`** (new): Added 3 tests:
  - Unauthenticated request returns 401
  - Non-admin user returns 403
  - Admin user can access metrics (200)

## Testing

```bash
node --test apps/api/src/tests/admin-role-check.test.js
```

All 3 tests pass. Existing health test also passes.

## Related Issues

Fixes #1640
References #743 (parent bounty)

### DIFF
--- a/apps/api/src/middleware/admin.js
+++ b/apps/api/src/middleware/admin.js
@@ -1,0 +1,8 @@
+import { fail } from "../utils/response.js";
+
+export function requireAdmin(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Forbidden: admin access required", 403);
+  }
+  return next();
+}
--- a/apps/api/src/routes/adminRoutes.js
+++ b/apps/api/src/routes/adminRoutes.js
@@ -1,8 +1,10 @@
 import { Router } from "express";
 import { metrics } from "../controllers/adminController.js";
 import { authMiddleware } from "../middleware/auth.js";
+import { requireAdmin } from "../middleware/admin.js";
 
 export const adminRoutes = Router();
 
 adminRoutes.use(authMiddleware);
+adminRoutes.use(requireAdmin);
 adminRoutes.get("/metrics", metrics);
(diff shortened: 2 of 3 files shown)


zeroknowledge0x:

PR #1654, opened 2026-05-29
fix: rate limiter should return API JSON errors instead of plain text

### LINKED ISSUE #1653: fix: rate limiter should return API JSON errors instead of plain text
## Bug

When the global API rate limiter is exceeded, `express-rate-limit` returns its default plain-text response instead of the JSON error shape used by the rest of the API. JSON clients calling API routes then need special-case parsing for `429` responses.

### Reproduction

Start the API and make 201 requests to a limited route inside the 15 minute window.

**Observed:** HTTP 429 uses the default rate-limit body (plain text) rather than `{ success: false, message: ... }`.
**Expected:** HTTP 429 should preserve the API JSON error contract.

### Root Cause

In `apps/api/src/middleware/rateLimit.js`, no custom `handler` is configured for the rate limiter.

### Proposed Fix

Add a custom `handler` that returns the shared JSON failure shape.

### Scope

/bounty $700

zeroknowledge0x:
## Summary

Rate limiter now returns JSON error responses matching the API's error contract.

## Bug

When the global API rate limiter is exceeded, `express-rate-limit` returns its default plain-text response instead of the JSON error shape `{ success: false, message: ... }` used by the rest of the API.

## Changes

- **`apps/api/src/middleware/rateLimit.js`**: Added custom `handler` that returns JSON 429 response
- **`apps/api/src/tests/rate-limit-json.test.js`** (new): Test verifying 429 responses are JSON

## Testing

```bash
node --test apps/api/src/tests/rate-limit-json.test.js
```

Test passes.

## Related Issues

Fixes #1653
References #743 (parent bounty)

### DIFF
--- a/apps/api/src/middleware/rateLimit.js
+++ b/apps/api/src/middleware/rateLimit.js
@@ -4,5 +4,11 @@ export const apiLimiter = rateLimit({
   windowMs: 15 * 60 * 1000,
   limit: 200,
   standardHeaders: "draft-7",
-  legacyHeaders: false
+  legacyHeaders: false,
+  handler: (req, res) => {
+    res.status(429).json({
+      success: false,
+      message: "Too many requests, please try again later"
+    });
+  }
 });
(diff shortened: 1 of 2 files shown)


zeroknowledge0x:

PR #1675, opened 2026-05-29
Fix auth registration and refresh validation

### LINKED ISSUE #1673: Bug: Registration can self-assign admin and refresh ignores submitted token
### Bug Description
The auth API currently has two related validation/authentication gaps:

1. `registerSchema` in `apps/api/src/validators/auth.js` accepts `role: "admin"`, so a public registration request can self-assign an administrative role.
2. `refresh()` in `apps/api/src/controllers/authController.js` calls `refreshToken()` without reading a submitted token from the request body. The service then returns a fresh access token without validating any caller-provided token.

### Expected Behavior
1. Public registration should only allow user-facing roles: `client` and `freelancer`. The server should keep `admin` assignment out of the registration schema.
2. The refresh endpoint should require a token payload from `req.body`, pass it to the auth service, and the service should verify the supplied token before issuing a new access token for the same subject/role.

### Scope
This issue is intentionally small and covers only auth validation/refresh behavior in the API package, plus focused tests.

### Disclaimer

keon0711:
/claim #743

Closes #1673.

## Summary
- Remove `admin` from the public registration role schema so new users can only self-select `client` or `freelancer`.
- Add a refresh request schema and pass the submitted token from the controller into the auth service.
- Verify the submitted token before issuing a refreshed access token with the same `sub` and `role` claims.
- Add focused auth tests for role validation, refresh input validation, and refresh token claim preservation.

## Demo
- `demos/auth-registration-refresh-demo.mp4`

## Validation
- `npm test -w apps/api`
- `node --check apps/api/src/validators/auth.js && node --check apps/api/src/services/authService.js && node --check apps/api/src/controllers/authController.js && node --check apps/api/src/tests/auth.test.js`
- `git diff --check`


keon0711:
Added the requested short demo video for the auth registration/refresh fix.

Demo: `demos/auth-registration-refresh-demo.mp4`
Validation remains:
- `npm test -w apps/api`
- `node --check apps/api/src/validators/auth.js && node --check apps/api/src/services/authService.js && node --check apps/api/src/controllers/authController.js && node --check apps/api/src/tests/auth.test.js`
- `git diff --check`

PR #1744, opened 2026-05-30

### LINKED ISSUE #1742: fix: prevent admin role self-assignment during registration
Same as #1426, registration role restriction.

---

Refs #743.

eutv2022:
Closes #1742.

## Summary
- remove `admin` from the public registration role schema
- return explicit 400 responses for invalid auth payloads
- add registration tests for rejecting admin self-assignment, defaulting to client, and allowing freelancer
- make the API test script run test files reliably with Node's test runner

## Verification
- `npm test -w apps/api` -> 4 tests passed
- `git diff --check` -> clean

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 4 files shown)


eutv2022:
Additional review note: this PR intentionally covers the full HTTP registration path, not just the Zod schema. Besides removing `admin` from the public role enum, it changes invalid auth payloads to return an explicit `400` response and adds regression tests that prove:

- `role: "admin"` is rejected through `POST /api/auth/register`
- omitted role still defaults to `client`
- `freelancer` remains accepted

Local verification: `npm test -w apps/api` passes with 4 tests.

PR #1892, opened 2026-05-30
Fix API test file glob

### LINKED ISSUE #1891: API test script should target test files instead of tests directory
## Bug

The API workspace test script runs `node --test src/tests`. On current Node/Windows this treats `src/tests` as a file path entry point, so the root `npm test` command fails with `MODULE_NOT_FOUND` even though the test file exists at `src/tests/health.test.js`.

## Evidence

From the repository root on Windows/Node 24:

```text
> npm test
> npm run test -w apps/api
> node --test src/tests

Error: Cannot find module '...\apps\api\src\tests'
```

Running Node's test runner with an explicit test-file glob works:

```text
node --test "src/tests/**/*.test.js"
```

## Expected fix

Update `apps/api/package.json` so the `test` script targets actual `*.test.js` files instead of the tests directory, making `npm test` work from the repository root.

Parent bounty: #743.

sodalone:
## Summary
- update the API workspace test script to target `src/tests/**/*.test.js`
- make the root `npm test` command discover the existing test file instead of trying to execute the `src/tests` directory as a module

## Validation
- `npm test` passes from the repository root
- `git diff --check` passes

Fixes #1891
/claim #743

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",


sodalone:

```text
$ npm test
> npm run test -w apps/api
> node --test "src/tests/**/*.test.js"
? pass 1

$ git diff --check
# no whitespace errors
```

This changes only the API workspace test target from a directory path to the existing `*.test.js` files, so the root test command works on Windows/Node 24.15.0.

PR #1895, opened 2026-05-30
Handle async validation errors

### LINKED ISSUE #1893: Async controller validation errors hang API requests
## Bug

Async API controllers can reject without being forwarded to Express error middleware. For example, `POST /api/jobs` calls `createJobSchema.parse(req.body)` inside an async controller. With an invalid body, the ZodError becomes an unhandled rejected promise, so the request hangs until the client times out instead of returning a structured 400 response.

## Evidence

Current behavior on `main`:

```text
POST /api/jobs { "title": "bad" }
unhandled ZodError [...]
client aborts after timeout
```

The same invalid payload should be treated as a client validation error, not a hung request or server-side unhandled rejection.

## Expected fix

- Wrap async route handlers so rejected promises are passed to `next()`.
- Return a structured `400` response for Zod validation errors.
- Add regression coverage proving invalid job payloads return validation errors instead of hanging.

Parent bounty: #743.

sodalone:
## Summary
- add an async route wrapper so rejected async controller promises reach Express error middleware
- return structured 400 JSON responses for Zod validation errors
- add regression coverage for invalid `/api/jobs` payloads so the request returns validation errors instead of hanging

## Validation
- `node --test apps\api\src\tests\*.test.js` passes
- `git diff --check` passes

Note: root `npm test` is fixed separately in #1892; this branch was verified with the direct Node test command above to avoid mixing the two scoped fixes.

Fixes #1893
/claim #743

### DIFF
--- a/apps/api/src/middleware/asyncHandler.js
+++ b/apps/api/src/middleware/asyncHandler.js
@@ -1,0 +1,5 @@
+export function asyncHandler(handler) {
+  return (req, res, next) => {
+    Promise.resolve(handler(req, res, next)).catch(next);
+  };
+}
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,19 @@
+import { ZodError } from "zod";
+
 export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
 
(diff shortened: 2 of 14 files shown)


sodalone:

```text
pass 8
fail 0

$ git diff --check
(no output)
```

Coverage note: the new regression test posts an invalid job payload and asserts the API returns HTTP 400 instead of leaving the request unresolved.

PR #1925, opened 2026-05-30
Validate OAuth callback requests

### LINKED ISSUE #1924: OAuth callback should reject unsupported providers and missing codes
## Bug

The existing `/api/auth/oauth/:provider/callback` endpoint accepts any provider path segment and succeeds even when the request does not include an OAuth authorization code. For example, `/api/auth/oauth/slack/callback?code=abc123` and `/api/auth/oauth/github/callback` both return success responses, which makes unsupported provider names and incomplete callbacks look valid.

## Expected behavior

- Reject unsupported OAuth provider names with a 400 response.
- Require a non-empty authorization `code` query value before returning callback success.
- Normalize supported provider names so `/GitHub/callback?code=...` is handled consistently.
- Add regression tests for unsupported providers, missing codes, and supported providers with a code.

Parent bounty: #743.

sodalone:
## Summary

- add an OAuth callback provider allowlist for GitHub and Google
- require a non-empty authorization `code` before returning callback success
- add regression tests for unsupported providers, missing codes, and normalized supported providers

## Validation

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Fixes #1924.

/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,8 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+
+const supportedOauthProviders = new Set(["github", "google"]);
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Coverage note: the regression tests verify that unsupported OAuth providers and callbacks without authorization codes now return HTTP 400, while a supported provider with a code still succeeds.

PR #1938, opened 2026-05-30
Restrict public registration roles

### LINKED ISSUE #1934: Public registration should not mint admin tokens
## Bug

The public `/api/auth/register` endpoint currently accepts `role: "admin"` because the registration schema includes `admin` in the allowed public role enum. The registration service then signs a JWT using the submitted role, so a public signup request can mint an admin-scoped token without an admin-only provisioning flow.

The same registration flow also generates the returned user id and JWT subject with separate `Date.now()` calls. Under normal timing they can diverge, which means the token subject may not identify the user id returned by the API response.

## Expected behavior

- Public registration should only allow `client` and `freelancer` roles.
- Attempts to self-assign `admin` during registration should return HTTP 400.
- Successful registrations should use the same generated user id for both the response `id` and JWT `sub`.
- Add regression tests for rejected admin registration, default client role, and token subject alignment.

Parent bounty: #743.

sodalone:
## Summary

- remove `admin` from the public registration role enum
- return HTTP 400 for invalid public registration payloads instead of minting an admin-scoped token
- generate the registered user id once and use it as both the API response id and JWT subject
- add regression tests for admin self-assignment rejection, default client role, and token subject alignment

## Validation

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Fixes #1934.

/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,10 +1,14 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
+  const parsed = registerSchema.safeParse(req.body);
+  if (!parsed.success) {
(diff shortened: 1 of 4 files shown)


sodalone:

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Coverage note: the regression tests verify that public registration rejects `role: "admin"`, default registrations still become clients, and successful registration JWT subjects match the returned user id.

PR #1948, opened 2026-05-30
Require admin role for admin routes

### LINKED ISSUE #1945: Admin routes should require admin role authorization
## Bug

The `/api/admin/*` routes currently call `authMiddleware`, but they do not verify that the authenticated token belongs to an admin user. Any user with a valid client or freelancer JWT can call `/api/admin/metrics` and receive admin-only platform metrics.

## Expected behavior

- Keep the existing authentication requirement for admin routes.
- Add role-based authorization so admin routes require `req.user.role === "admin"`.
- Return HTTP 403 for authenticated non-admin users.
- Preserve successful access for valid admin tokens.
- Add regression tests for unauthenticated, non-admin, and admin access to `/api/admin/metrics`.

Parent bounty: #743.

sodalone:
## Summary

- add a reusable `requireRole()` authorization middleware
- require `role: "admin"` after authentication for all `/api/admin/*` routes
- return HTTP 403 for authenticated non-admin tokens
- add regression tests covering unauthenticated, client-token, and admin-token access to `/api/admin/metrics`

## Validation

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Fixes #1945.

/claim #743

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,13 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function requireRole(role) {
+  return function roleMiddleware(req, res, next) {
+    if (req.user?.role !== role) {
+      return fail(res, "Forbidden", 403);
+    }
+
+    return next();
+  };
+}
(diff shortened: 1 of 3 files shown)


sodalone:

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Coverage note: the regression tests verify that `/api/admin/metrics` returns 401 without a token, 403 for a non-admin token, and 200 for an admin token.

PR #1955, opened 2026-05-30
fix: Do not mint admin tokens on public registration

### LINKED ISSUE #1953: Bug: Public registration should not mint admin tokens
Bug: Public registration should not mint admin tokens

The public /api/auth/register endpoint currently accepts role: 'admin'. Public registration should only allow client and freelancer roles. Attempts to self-assign admin during registration should return HTTP 400.

Ishant5436:
Closes #1953


Ishant5436:

PR #1962, opened 2026-05-30
Protect payment creation endpoint

### LINKED ISSUE #1959: Payment creation should require auth and validate amounts
## Bug

The `/api/payments` endpoint currently accepts unauthenticated POST requests and forwards the raw request body to `createPaymentIntent`. That means a caller can create payment intents without a bearer token and can submit invalid values such as zero or negative amounts.

## Expected behavior

- Require authentication before creating payment intents.
- Validate the payment payload before passing it to the payment service.
- Reject zero, negative, missing, or non-numeric amounts with HTTP 400.
- Preserve the existing default `usd` currency behavior for valid authenticated requests.
- Add regression tests for unauthenticated access, invalid amount rejection, and successful authenticated payment intent creation.

Parent bounty: #743.

sodalone:
## Summary

- require bearer authentication before creating payment intents
- validate payment payloads with a Zod schema before calling the payment service
- reject zero or invalid amounts with HTTP 400 while preserving the default `usd` currency for valid requests
- add regression tests for unauthenticated access, invalid amount rejection, and successful authenticated payment creation

## Validation

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Fixes #1959.

/claim #743

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { paymentIntentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const parsed = paymentIntentSchema.safeParse(req.body);
+  if (!parsed.success) {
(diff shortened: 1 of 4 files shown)


sodalone:

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Coverage note: the regression tests verify that `/api/payments` returns 401 without a token, rejects a zero amount with HTTP 400, and creates a payment intent for an authenticated positive amount while preserving the default `usd` currency.

PR #1967, opened 2026-05-30
Validate search query input

### LINKED ISSUE #1965: Search endpoint should reject malformed query values
## Bug

The `/api/search` endpoint forwards `req.query.q` directly into `globalSearch`. Express can represent repeated query values as an array, and oversized query strings are accepted without limit. That lets malformed search requests reach the service layer instead of being rejected at the API boundary.

## Expected behavior

- Accept a single string `q` value and trim surrounding whitespace.
- Reject repeated query values such as `?q=design&q=engineering` with HTTP 400.
- Reject oversized query values before they reach `globalSearch`.
- Preserve existing successful search responses for valid queries.
- Add regression tests for valid trimming, repeated values, and oversized values.

Parent bounty: #743.

sodalone:
## Summary

- validate `/api/search` query input before calling the search service
- trim valid string queries and reject repeated or oversized query values with HTTP 400
- add regression tests for valid trimming, repeated query parameters, and oversized queries

## Validation

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Fixes #1965.

/claim #743

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchQuerySchema } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const parsed = searchQuerySchema.safeParse(req.query.q ?? "");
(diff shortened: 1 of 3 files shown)


sodalone:

```text
pass 4
fail 0

$ git diff --check
(no output)
```

Coverage note: the regression tests verify that `/api/search` trims valid query strings, rejects repeated query values, and rejects oversized query values before calling `globalSearch`.

PR #1998, opened 2026-05-30
Codex: Keep new notifications unread

### LINKED ISSUE #1997: Notification creation should use server-owned unread status
## Bug

`POST /api/notifications` lets a direct caller create a notification that is already marked as read. `apps/api/src/services/notificationService.js` builds the new object with `read: false` before spreading the caller payload:

```js
```

A payload such as `{ "message": "Welcome", "read": true }` overrides the server-owned unread default and returns a freshly created notification with `read: true`.

## Impact

- New notifications can bypass the unread state that list views and counters rely on.
- API consumers can make newly-created notifications invisible to unread-only UI logic.
- The service behavior is inconsistent with the expected server-owned initial state for new records.

## Expected behavior

New notifications should always start unread:

- `read: false` for every newly-created notification.
- Caller-provided `read` should not override the initial unread state during creation.
- Existing payload fields such as `message`, `type`, or `userId` should still be preserved.

## Proposed fix

- Apply caller payload fields first, then apply server-owned `id` and `read: false` after the payload.
- Add focused regression tests for default unread status and caller-supplied `read: true` override prevention.
- Keep the change scoped to notification creation behavior.

Parent bounty: #743.

keilogic:
/claim #743

## Summary
- Keep newly-created notifications server-owned as unread by applying `read: false` after the caller payload.
- Preserve normal caller payload fields such as `message` and `type`.
- Add focused regression tests for default unread creation and caller-supplied `read: true` override prevention.

Closes #1997
References #743

## Demo / Evidence
Focused test output:

```text
✔ createNotification starts new notifications unread
✔ createNotification ignores caller supplied read state
```

## Validation
- `node --check apps/api/src/services/notificationService.js`
- `node --check apps/api/src/tests/notificationService.test.js`
- `node --test apps/api/src/tests/notificationService.test.js`
- `git diff --check`

Known upstream test-script note: `npm test -w apps/api` still fails because the existing script invokes `node --test src/tests` as a directory path in this checkout. The explicit API test files above pass.

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,7 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const notification = { ...payload, id: `ntf_${Date.now()}`, read: false };
   notifications.push(notification);
   return notification;
 }
--- a/apps/api/src/tests/notificationService.test.js
+++ b/apps/api/src/tests/notificationService.test.js
@@ -1,0 +1,25 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createNotification } from "../services/notificationService.js";
+
+test("createNotification starts new notifications unread", async () => {
+  const notification = await createNotification({
+    message: "Welcome to FreelanceFlow",
+    type: "system"
+  });
+
+  assert.equal(notification.read, false);
+  assert.equal(notification.message, "Welcome to FreelanceFlow");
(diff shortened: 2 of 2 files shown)


keilogic:
Rebased this branch onto current `origin/main` to refresh mergeability.

Validation rerun after rebase:

- `node --check apps/api/src/services/notificationService.js`
- `node --check apps/api/src/tests/notificationService.test.js`
- `node --test apps/api/src/tests/notificationService.test.js` (2 pass)
- `node --test apps/api/src/tests/*.test.js` (3 pass)
- `git diff --check`

PR #2117, opened 2026-05-30
Run rate limiter before JSON parser

### LINKED ISSUE #2098: Rate limiter should run before JSON body parsing
The API currently mounts `express.json()` before the global `apiLimiter`. Malformed JSON requests can therefore be parsed and rejected by the body parser before the shared rate limiter sees them, which leaves a cheap path for repeated invalid payloads to consume parsing/error-handler work outside the intended API throttle.

Expected behavior: the global rate limiter should run immediately after security/CORS middleware, before JSON body parsing, so malformed and valid API requests are counted consistently.

Suggested fix: move `app.use(apiLimiter)` before `app.use(express.json())` and add regression coverage that repeated malformed JSON requests eventually receive the limiter's `429` response.

sodalone:
## Summary
- move the global API rate limiter ahead of `express.json()`
- make malformed JSON requests count toward the same API rate limit as valid requests
- add regression coverage proving the 201st malformed JSON request is rejected with `429`

## Validation
```text
? rate limiter counts malformed JSON before body parsing
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

Fixes #2098.
/claim #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -20,8 +20,8 @@ export function createApp() {
 
   app.use(helmet());
   app.use(cors());
-  app.use(express.json());
   app.use(apiLimiter);
+  app.use(express.json());
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? rate limiter counts malformed JSON before body parsing
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

PR #2120, opened 2026-05-30
fix: normalize auth email input

### LINKED ISSUE #2118: Auth email identity should be canonicalized before registration and login
## Bug

The auth validators accept valid mixed-case email addresses and pass them through unchanged. `registerUser()` stores/returns whatever casing the client sent, while `loginUser()` signs in against the raw email from the request.

That means these requests are treated as different identities at the API boundary even though email local account workflows normally need a single canonical identity:

- `POST /api/auth/register` with `User@Example.com`
- `POST /api/auth/login` with `user@example.com`

Once this is wired to Prisma's unique `User.email` field, the lack of canonicalization can allow duplicate-looking accounts or failed logins depending on database collation and lookup casing.

## Expected behavior

Auth email input should be normalized once at validation time before controllers call the auth service:

- trim surrounding whitespace before email validation
- lowercase the validated email for registration and login payloads
- preserve existing password and role validation behavior

## Scope

- Update `registerSchema` and `loginSchema` to canonicalize email strings.
- Add focused API tests proving register/login responses use normalized email values.
- Keep the change limited to auth validation; do not introduce persistence or password-hashing changes.

Related: #743

Entr0zy:
## Summary
- normalize register/login email input with one shared Zod schema helper
- trim surrounding whitespace and lowercase validated auth email addresses before service use
- add API regression coverage for mixed-case register and login payloads

Closes #2118
/claim #743

## Validation
- `node --test src/tests/*.test.js` from `apps/api` (3 tests passed)
- `git diff --check`

Note: `npm test -w apps/api` currently fails on this Windows environment because `node --test src/tests` is treated as a module path rather than discovering test files; the direct file-glob command above passes.

### DIFF
--- a/apps/api/src/tests/auth.test.js
+++ b/apps/api/src/tests/auth.test.js
@@ -1,0 +1,59 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
+async function withServer(run) {
+  const app = createApp();
+  const server = app.listen(0);
+
+  await new Promise((resolve, reject) => {
+    server.once("listening", resolve);
+    server.once("error", reject);
+  });
+
+  try {
+    const { port } = server.address();
+    await run(`http://127.0.0.1:${port}`);
+  } finally {
+    await new Promise((resolve, reject) => {
+      server.close((error) => (error ? reject(error) : resolve()));
+    });
+  }
+}
+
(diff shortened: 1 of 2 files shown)


Entr0zy:
Hi, checking on this #743 bounty claim. Could you confirm whether anything is blocking review/reward for this focused fix, or whether any changes are needed from me?

PR #2124, opened 2026-05-30
Match registration token subject

### LINKED ISSUE #2119: Registration token subject should match returned user id
`registerUser` currently calls `Date.now()` once for the returned user id and again for the JWT `sub` claim. If the clock advances between those calls, the API can return one user id while issuing a token for a different subject.

Expected behavior: a registration response should use one generated user id consistently for both the response payload and the access token subject.

Suggested fix: generate the user id once, reuse it in the response and `signAccessToken`, and add regression coverage that decodes the token and confirms `sub` equals the returned `id`.

sodalone:
## Summary
- generate the registration user id once and reuse it in the response and JWT subject
- prevent newly registered users from receiving a token whose `sub` can drift from the returned `id`
- add regression coverage with a mocked clock that advances between ID-generation calls

## Validation
```text
? registerUser signs a token for the returned user id
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

Fixes #2119.
/claim #743

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = `usr_${Date.now()}`;
+
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
     email: payload.email,
     role: payload.role,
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? registerUser signs a token for the returned user id
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

PR #2131, opened 2026-05-30
Protect auth token refresh

### LINKED ISSUE #2128: Refresh endpoint should require authentication
`POST /api/auth/refresh` currently calls `refreshToken()` without any authentication middleware or refresh-token input, so an unauthenticated client can request a new access token.

Expected behavior: the refresh endpoint should require a valid bearer token (or future refresh-token credential) before issuing a replacement token, and the replacement token should preserve the authenticated subject and role.

Suggested fix: protect `/refresh` with the existing auth middleware, pass the verified user claims into `refreshToken`, and add route coverage for both unauthenticated rejection and authenticated renewal.

sodalone:
## Summary
- require the existing auth middleware before `POST /api/auth/refresh` can issue a new token
- pass verified user claims into the refresh service instead of minting a generic `usr_existing` client token
- add route coverage for unauthenticated rejection and authenticated renewal preserving `sub`/`role`

## Validation
```text
? POST /api/auth/refresh requires authentication
? POST /api/auth/refresh renews the authenticated subject
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2128.
/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const result = await refreshToken(req.user);
   return ok(res, result);
 }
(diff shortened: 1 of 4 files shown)


sodalone:

```text
? POST /api/auth/refresh requires authentication
? POST /api/auth/refresh renews the authenticated subject
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2142, opened 2026-05-30
Protect server-managed fields

### LINKED ISSUE #2136: Creation services should protect server-managed fields
Several in-memory creation services build records with server-managed fields first and then spread the request payload afterward, for example `{ id: generatedId, ...payload }`. That lets a client-supplied `id` override the generated id. `createJob` similarly allows a payload `status` to override the intended default `open`, and `createNotification` allows `read` to be pre-set by the payload.

Expected behavior: server-managed fields such as generated ids, job status defaults, notification read state, and message timestamps should be assigned after user-controlled payload data so they cannot be overridden by request input.

Suggested fix: spread payload fields first, assign server-managed fields last, and add regression coverage for the affected creation services.

sodalone:
## Summary
- assign generated ids after user-controlled payloads in creation services
- keep job status, notification read state, and message timestamps server-managed
- add regression coverage proving payload `id`/`status`/`read`/`sentAt` values cannot override server-managed fields

## Validation
```text
? creation services do not allow payloads to override server-managed fields
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

Fixes #2136.
/claim #743

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,7 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const job = { ...payload, id: `job_${Date.now()}`, status: "open" };
   jobs.push(job);
   return job;
 }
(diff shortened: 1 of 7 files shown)


sodalone:

```text
? creation services do not allow payloads to override server-managed fields
? tests 2
? pass 2
? fail 0
```

```text
$ git diff --check
```

PR #2146, opened 2026-05-30
Require production JWT secret

### LINKED ISSUE #2145: Production API should require an explicit JWT secret
The API config currently falls back to the literal `development-secret` whenever `JWT_SECRET` is missing. That fallback is useful for local development, but in `NODE_ENV=production` it would allow the server to sign and verify access tokens with a public, predictable secret.

Expected behavior: production startup should fail fast when `JWT_SECRET` is not configured, while local development can keep the existing fallback.

Suggested fix: validate `NODE_ENV` and `JWT_SECRET` while building the env config, throw a clear error for missing production secrets, and add regression coverage for both production failure and development fallback behavior.

sodalone:
## Summary
- fail fast when `NODE_ENV=production` is missing `JWT_SECRET`
- keep the existing `development-secret` fallback for local development only
- add dynamic import coverage for production failure and development fallback behavior

## Validation
```text
? production config requires JWT_SECRET
? development config keeps the local JWT_SECRET fallback
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2145.
/claim #743

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,14 @@
+const nodeEnv = process.env.NODE_ENV ?? "development";
+const jwtSecret = process.env.JWT_SECRET ?? "";
+
+if (nodeEnv === "production" && jwtSecret.length === 0) {
+  throw new Error("JWT_SECRET is required in production");
+}
+
 export const env = {
-  nodeEnv: process.env.NODE_ENV ?? "development",
+  nodeEnv,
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? production config requires JWT_SECRET
? development config keeps the local JWT_SECRET fallback
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2148, opened 2026-05-30
Validate payment intent requests

### LINKED ISSUE #2147: Payment intent requests should validate amount and currency
`POST /api/payments` currently passes the request body directly to `createPaymentIntent`, so negative/zero amounts and arbitrary currency values can produce a successful payment-intent response.

Expected behavior: payment intent creation should reject non-positive amounts and validate currency input before returning a successful response. Accepted currency values should be normalized before they reach the service layer.

Suggested fix: add a payment request validator, return a 400 response for invalid payment requests, normalize valid currency codes, and add route coverage for rejected negative amounts and accepted normalized currency values.

sodalone:
## Summary
- validate payment-intent request bodies before calling the payment service
- reject non-positive amounts with a 400 response
- normalize valid 3-letter currency codes before returning the payment-intent response

## Validation
```text
? POST /api/payments rejects non-positive amounts
? POST /api/payments normalizes valid currency codes
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2147.
/claim #743

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,13 @@
-import { ok } from "../utils/response.js";
+import { createPaymentSchema } from "../validators/payment.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/payments rejects non-positive amounts
? POST /api/payments normalizes valid currency codes
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2159, opened 2026-05-30
Validate review creation requests

### LINKED ISSUE #2158: Review creation should validate rating bounds
`POST /api/reviews` currently passes the request body directly to `createReview`, so invalid review payloads such as ratings outside the normal 1-5 range can be accepted and returned as successful reviews.

Expected behavior: review creation should validate required identifiers, require a non-empty comment, and reject ratings outside the supported 1-5 range before reaching the service layer.

Suggested fix: add a review creation validator, return a 400 response for invalid review payloads, and add route coverage for rejected out-of-range ratings plus a valid review payload.

sodalone:
## Summary
- validate review creation payloads before calling the review service
- reject ratings outside the supported 1-5 range with a 400 response
- require non-empty comments and reviewer/reviewee ids for new reviews

## Validation
```text
? POST /api/reviews rejects ratings outside the supported range
? POST /api/reviews accepts valid review payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2158.
/claim #743

### DIFF
--- a/apps/api/src/controllers/reviewController.js
+++ b/apps/api/src/controllers/reviewController.js
@@ -1,4 +1,5 @@
-import { ok } from "../utils/response.js";
+import { createReviewSchema } from "../validators/review.js";
+import { fail, ok } from "../utils/response.js";
 import { createReview, listReviews } from "../services/reviewService.js";
 
 export async function getReviews(req, res) {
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/reviews rejects ratings outside the supported range
? POST /api/reviews accepts valid review payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2171, opened 2026-05-30
Validate message creation requests

### LINKED ISSUE #2167: Message creation should validate sender receiver and body
`POST /api/messages` currently forwards the request body directly to `sendMessage`, so empty message bodies and missing sender/receiver ids can be accepted and stored in the in-memory message list.

Expected behavior: message creation should validate required participants and require a non-empty body before calling the service layer. Whitespace-only bodies should be rejected rather than returning a successful message response.

Suggested fix: add a message creation validator, return a 400 response for invalid message payloads, and add route coverage for rejected blank messages plus a valid message payload.

sodalone:
## Summary
- validate message creation payloads before calling the message service
- reject blank message bodies with a 400 response
- require sender and receiver ids for new messages

## Validation
```text
? POST /api/messages rejects blank message bodies
? POST /api/messages accepts valid message payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2167.
/claim #743

### DIFF
--- a/apps/api/src/controllers/messageController.js
+++ b/apps/api/src/controllers/messageController.js
@@ -1,4 +1,5 @@
-import { ok } from "../utils/response.js";
+import { createMessageSchema } from "../validators/message.js";
+import { fail, ok } from "../utils/response.js";
 import { listMessages, sendMessage } from "../services/messageService.js";
 
 export async function getMessages(req, res) {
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/messages rejects blank message bodies
? POST /api/messages accepts valid message payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2174, opened 2026-05-30
Validate proposal creation requests

### LINKED ISSUE #2172: Proposal creation should validate bid amount and required fields
`POST /api/proposals` currently forwards the request body directly to `createProposal`, so malformed proposals such as negative bid amounts, blank cover letters, or missing job/freelancer ids can be accepted and stored.

Expected behavior: proposal creation should validate required fields before calling the service layer. Bid amounts should be positive, text fields should be non-empty after trimming, and job/freelancer ids should be present.

Suggested fix: add a proposal creation validator, return a 400 response for invalid proposal payloads, and add route coverage for rejected negative bids plus a valid proposal payload.

sodalone:
## Summary
- validate proposal creation payloads before calling the proposal service
- reject non-positive bid amounts with a 400 response
- require non-empty cover letters, estimated durations, job ids, and freelancer ids

## Validation
```text
? POST /api/proposals rejects non-positive bid amounts
? POST /api/proposals accepts valid proposal payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2172.
/claim #743

### DIFF
--- a/apps/api/src/controllers/proposalController.js
+++ b/apps/api/src/controllers/proposalController.js
@@ -1,4 +1,5 @@
-import { ok } from "../utils/response.js";
+import { createProposalSchema } from "../validators/proposal.js";
+import { fail, ok } from "../utils/response.js";
 import { createProposal, listProposals } from "../services/proposalService.js";
 
 export async function getProposals(req, res) {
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/proposals rejects non-positive bid amounts
? POST /api/proposals accepts valid proposal payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2179, opened 2026-05-30
Validate public user creation

### LINKED ISSUE #2178: Public user creation should validate email and disallow admin roles
`POST /api/users` currently forwards the request body directly to `createUser`, so malformed users can be created with invalid email values. The public endpoint also accepts arbitrary roles, including `admin`, because no request schema constrains the payload before it reaches the service layer.

Expected behavior: public user creation should validate email and only allow public self-service roles such as `client` and `freelancer`. Admin role assignment should not be accepted through this public creation endpoint.

Suggested fix: add a user creation validator, return a 400 response for invalid payloads or `admin` roles, and add route coverage for rejected admin-role payloads plus a valid user payload.

sodalone:
## Summary
- validate public user creation payloads before calling the user service
- reject public attempts to create users with the `admin` role
- require valid email addresses and constrain public roles to `client` or `freelancer`

## Validation
```text
? POST /api/users rejects public admin role creation
? POST /api/users accepts valid public user payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2178.
/claim #743

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,4 +1,5 @@
-import { ok } from "../utils/response.js";
+import { createUserSchema } from "../validators/user.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
 
 export async function getUsers(req, res) {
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/users rejects public admin role creation
? POST /api/users accepts valid public user payloads
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2182, opened 2026-05-30
Validate job budget ranges

### LINKED ISSUE #2180: Job creation should reject inverted budget ranges
`POST /api/jobs` currently validates each budget endpoint as non-negative, but it does not reject inverted ranges where `budgetMin` is greater than `budgetMax`. Those payloads can create jobs whose advertised budget interval is impossible to satisfy.

Expected behavior: job creation should reject payloads where `budgetMin > budgetMax` before calling the service layer, while valid ranges should still create jobs successfully.

Suggested fix: add a cross-field budget range refinement to the job creation schema, return a 400 response for invalid job payloads, and add route coverage for rejected inverted ranges plus a valid range.

sodalone:
## Summary
- reject job creation payloads where `budgetMin` is greater than `budgetMax`
- return a 400 response for invalid job payloads instead of passing them to the service layer
- keep valid budget ranges working and covered by route tests

## Validation
```text
? POST /api/jobs rejects inverted budget ranges
? POST /api/jobs accepts valid budget ranges
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2180.
/claim #743

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -1,4 +1,4 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createJobSchema } from "../validators/job.js";
 import { createJob, listJobs } from "../services/jobService.js";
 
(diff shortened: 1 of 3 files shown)


sodalone:

```text
? POST /api/jobs rejects inverted budget ranges
? POST /api/jobs accepts valid budget ranges
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

PR #2184, opened 2026-05-30
fix: use canonical login role claim

### LINKED ISSUE #2183: Login tokens should use canonical UserRole casing
## Bug

`loginUser()` signs access tokens with a lowercase role claim:

```js
token: signAccessToken({ sub: usr_existing, role: client })
```

The repository's Prisma schema defines `UserRole` enum values as uppercase `CLIENT`, `FREELANCER`, and `ADMIN`. That makes login-issued tokens inconsistent with the canonical domain role casing and with server-side checks that compare JWT role claims to Prisma enum values.

This is separate from registration normalization: a user who logs in through the existing login service still receives a lowercase `client` role claim.

## Expected behavior

Login-issued access tokens should use the canonical role value for the authenticated user. In the current stubbed login service, the existing client user should be represented as `CLIENT` in the JWT role claim.

## Scope

- Update `loginUser()` so the issued token uses `role: CLIENT`.
- Add a focused service regression test that decodes the login token and asserts the canonical role claim.
- Keep the change scoped to login token role casing only; do not change password verification, registration, or refresh-token behavior here.

Related: #743

Entr0zy:
## Summary
- update the stubbed login service to sign `CLIENT` instead of lowercase `client`
- add a focused service regression that verifies the decoded login JWT role claim
- keep the scope limited to login-token role casing

Closes #2183
/claim #743

## Validation
- `node --test src/tests/authService.test.js src/tests/health.test.js` from `apps/api`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -14,7 +14,7 @@ export async function loginUser(payload) {
   // TODO: verify password hash against stored user record
   return {
     email: payload.email,
-    token: signAccessToken({ sub: "usr_existing", role: "client" })
+    token: signAccessToken({ sub: "usr_existing", role: "CLIENT" })
   };
 }
 
--- a/apps/api/src/tests/authService.test.js
+++ b/apps/api/src/tests/authService.test.js
@@ -1,0 +1,16 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { loginUser } from "../services/authService.js";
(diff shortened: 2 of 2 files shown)


Entr0zy:
Hi, checking on this #743 bounty claim. Could you confirm whether anything is blocking review/reward for this focused auth-role fix, or whether any changes are needed from me?

PR #2230, opened 2026-05-30
Require auth for message routes

### LINKED ISSUE #2187: Message routes should require authentication
The `/api/messages` endpoints currently expose message listing and creation without authentication. Messages are private account data, so these routes should require a valid authenticated user before returning or accepting message payloads.

Expected behavior:
- unauthenticated `GET /api/messages` requests return 401
- unauthenticated `POST /api/messages` requests return 401
- authenticated requests continue to reach the existing message handlers

sodalone:
## Summary
- protect message routes with the existing auth middleware
- reject unauthenticated access to message list/create endpoints

## Validation
```text
? GET /api/messages rejects unauthenticated requests
? GET /api/messages accepts authenticated requests
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2187.
/claim #743

### DIFF
--- a/apps/api/src/routes/messageRoutes.js
+++ b/apps/api/src/routes/messageRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getMessages, postMessage } from "../controllers/messageController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const messageRoutes = Router();
 
+messageRoutes.use(authMiddleware);
 messageRoutes.get("/", getMessages);
 messageRoutes.post("/", postMessage);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? GET /health returns ok payload (53.6241ms)
? GET /api/messages rejects unauthenticated requests (58.3426ms)
? GET /api/messages accepts authenticated requests (17.6825ms)
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
warning: in the working copy of 'apps/api/src/routes/messageRoutes.js', LF will be replaced by CRLF the next time Git touches it
```

PR #2233, opened 2026-05-30
Require auth for notification routes

### LINKED ISSUE #2231: Notification routes should require authentication
The `/api/notifications` endpoints currently expose notification listing and creation without authentication. Notifications are private account data, so these routes should require a valid authenticated user before returning or accepting notification payloads.

Expected behavior:
- unauthenticated `GET /api/notifications` requests return 401
- unauthenticated `POST /api/notifications` requests return 401
- authenticated requests continue to reach the existing notification handlers

sodalone:
## Summary
- protect notification routes with the existing auth middleware
- reject unauthenticated access to notification list/create endpoints

## Validation
```text
? GET /api/notifications rejects unauthenticated requests
? GET /api/notifications accepts authenticated requests
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2231.
/claim #743

### DIFF
--- a/apps/api/src/routes/notificationRoutes.js
+++ b/apps/api/src/routes/notificationRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getNotifications, postNotification } from "../controllers/notificationController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const notificationRoutes = Router();
 
+notificationRoutes.use(authMiddleware);
 notificationRoutes.get("/", getNotifications);
 notificationRoutes.post("/", postNotification);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? GET /health returns ok payload (47.1979ms)
? GET /api/notifications rejects unauthenticated requests (49.4767ms)
? GET /api/notifications accepts authenticated requests (12.0661ms)
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
warning: in the working copy of 'apps/api/src/routes/notificationRoutes.js', LF will be replaced by CRLF the next time Git touches it
```

PR #2268, opened 2026-05-30
Require auth for proposal routes

### LINKED ISSUE #2234: Proposal routes should require authentication
The `/api/proposals` endpoints currently expose proposal listing and creation without authentication. Proposals include private bid and freelancer/client workflow data, so these routes should require a valid authenticated user before returning or accepting proposal payloads.

Expected behavior:
- unauthenticated `GET /api/proposals` requests return 401
- unauthenticated `POST /api/proposals` requests return 401
- authenticated requests continue to reach the existing proposal handlers

sodalone:
## Summary
- protect proposal routes with the existing auth middleware
- reject unauthenticated access to proposal list/create endpoints

## Validation
```text
? GET /api/proposals rejects unauthenticated requests
? GET /api/proposals accepts authenticated requests
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2234.
/claim #743

### DIFF
--- a/apps/api/src/routes/proposalRoutes.js
+++ b/apps/api/src/routes/proposalRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getProposals, postProposal } from "../controllers/proposalController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const proposalRoutes = Router();
 
+proposalRoutes.use(authMiddleware);
 proposalRoutes.get("/", getProposals);
 proposalRoutes.post("/", postProposal);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? GET /health returns ok payload (51.1011ms)
? GET /api/proposals rejects unauthenticated requests (56.1885ms)
? GET /api/proposals accepts authenticated requests (13.3521ms)
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
warning: in the working copy of 'apps/api/src/routes/proposalRoutes.js', LF will be replaced by CRLF the next time Git touches it
```

PR #2289, opened 2026-05-30
Require auth before uploads

### LINKED ISSUE #2275: Upload route should require authentication
The `/api/uploads` endpoint currently accepts upload requests without authentication. Upload endpoints consume memory and should only accept file payloads from authenticated users, with the auth check running before `multer` parses the file body.

Expected behavior:
- unauthenticated `POST /api/uploads` requests return 401
- authenticated upload requests continue to reach the existing upload handler
- authentication runs before the in-memory upload middleware
- route-level coverage documents both unauthenticated and authenticated behavior

sodalone:
## Summary
- require authentication before upload requests reach the in-memory multer middleware
- reject unauthenticated upload attempts with the existing 401 response
- add route coverage for unauthenticated rejection and authenticated multipart upload success

## Validation
```text
? POST /api/uploads rejects unauthenticated requests
? POST /api/uploads accepts authenticated upload requests
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2275.
/claim #743

### DIFF
--- a/apps/api/src/routes/uploadRoutes.js
+++ b/apps/api/src/routes/uploadRoutes.js
@@ -1,9 +1,10 @@
 import { Router } from "express";
 import multer from "multer";
 import { uploadFile } from "../controllers/uploadController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 const upload = multer({ storage: multer.memoryStorage() });
 
 export const uploadRoutes = Router();
 
-uploadRoutes.post("/", upload.single("file"), uploadFile);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? GET /health returns ok payload (50.8105ms)
? POST /api/uploads rejects unauthenticated requests (54.7056ms)
? POST /api/uploads accepts authenticated upload requests (23.6608ms)
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
warning: in the working copy of 'apps/api/src/routes/uploadRoutes.js', LF will be replaced by CRLF the next time Git touches it
```

PR #2298, opened 2026-05-30
Require auth for job creation

### LINKED ISSUE #2297: Job creation should require authentication
The `/api/jobs` collection is useful for public browsing, but creating a job should require an authenticated user. The route currently lets unauthenticated callers create job records directly, which can lead to spam listings and unowned hiring data.

Expected behavior:
- unauthenticated `POST /api/jobs` requests return 401
- authenticated `POST /api/jobs` requests continue to reach the existing job creation handler
- public `GET /api/jobs` behavior remains unchanged
- route-level coverage documents unauthenticated rejection and authenticated success

sodalone:
## Summary
- require authentication before creating jobs
- keep public job listing unchanged
- add route coverage for unauthenticated rejection and authenticated job creation success

## Validation
```text
? POST /api/jobs rejects unauthenticated requests
? POST /api/jobs accepts authenticated requests
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
```

Fixes #2297.
/claim #743

### DIFF
--- a/apps/api/src/routes/jobRoutes.js
+++ b/apps/api/src/routes/jobRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getJobs, postJob } from "../controllers/jobController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const jobRoutes = Router();
 
 jobRoutes.get("/", getJobs);
-jobRoutes.post("/", postJob);
+jobRoutes.post("/", authMiddleware, postJob);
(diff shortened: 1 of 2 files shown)


sodalone:

```text
? GET /health returns ok payload (50.889ms)
? POST /api/jobs rejects unauthenticated requests (64.1778ms)
? POST /api/jobs accepts authenticated requests (13.6426ms)
? tests 3
? pass 3
? fail 0
```

```text
$ git diff --check
warning: in the working copy of 'apps/api/src/routes/jobRoutes.js', LF will be replaced by CRLF the next time Git touches it
```

PR #2311, opened 2026-05-30
Require auth for user listing

### LINKED ISSUE #2309: User listing should require authentication
## Bug

`GET /api/users` is registered without `authMiddleware`, so any unauthenticated caller can enumerate user records through the public API.

## Impact

The current in-memory user store can contain account records created during the process lifetime, and a future persistent user backing store would make this route a direct anonymous account enumeration endpoint.

## Expected behavior

- `GET /api/users` should require a valid bearer token before returning the user collection.
- Public user creation behavior should remain unchanged.
- Add focused regression tests for anonymous and authenticated reads.

Refs #743

sodalone:
## Summary
- Require a valid bearer token for `GET /api/users` before returning the user collection.
- Keep `POST /api/users` public so existing user creation behavior is unchanged.
- Add regression coverage for anonymous and authenticated user-list reads.

## Validation
- `git diff --check`

Fixes #2309
Refs #743
/claim #743

### DIFF
--- a/apps/api/src/routes/userRoutes.js
+++ b/apps/api/src/routes/userRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getUsers, postUser } from "../controllers/userController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const userRoutes = Router();
 
-userRoutes.get("/", getUsers);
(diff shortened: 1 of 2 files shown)


sodalone:
Validation completed locally:

PR #2318, opened 2026-05-30
Validate login credentials

### LINKED ISSUE #2316: Login should validate registered credentials before issuing tokens
## Bug

`loginUser` currently returns a JWT for any syntactically valid email/password pair. The service does not check whether the account exists or whether the submitted password matches a registered user.

## Impact

Any unauthenticated caller can obtain an access token without owning an account, which bypasses the authentication boundary for routes that trust bearer tokens.

## Expected behavior

- Registration should create a server-side user record for the process lifetime.
- Login should reject unknown users and wrong passwords with `401`.
- Login should issue a token only for a matching registered user.
- Add focused regression tests for unknown credentials, wrong passwords, and successful login.

Refs #743

sodalone:
## Summary
- Store registered users in the existing in-memory auth service for the process lifetime.
- Reject login attempts for unknown users or wrong passwords with `401 Invalid credentials`.
- Issue login tokens only for matching registered users, using the stored user id and role.
- Add regression coverage for unknown credentials, wrong passwords, and successful login.

## Validation
- `git diff --check`

Fixes #2316
Refs #743
/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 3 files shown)


sodalone:
Validation completed locally:

PR #2324, opened 2026-05-30
fix: Require admin role for admin routes

### LINKED ISSUE #2322: Bug: Admin routes should require the admin role
Bug: Admin routes should require the admin role.

Ishant5436:
Closes #2322


Ishant5436:

PR #2332, opened 2026-05-30
Redact user credential fields

### LINKED ISSUE #2326: User responses should redact credential fields
## Bug

`createUser` currently stores arbitrary request fields and returns the resulting user object directly. `listUsers` also returns the stored objects directly. If a public caller includes credential-like fields such as `password`, `passwordHash`, or tokens, those secrets are echoed in the create response and exposed again by the user list response.

## Impact

Credential material can leak through normal API responses. This is especially risky because the current service accepts free-form user payloads and the list endpoint reflects stored records.

## Expected behavior

- User API responses should not include password, password hash, or token fields.
- Redaction should apply to both newly created users and listed users.
- Add regression tests for create and list responses.

Refs #743

sodalone:
## Summary
- Redact credential-like fields from user API responses.
- Apply the redaction to both newly created user responses and listed users.
- Add regression tests proving `password`, password hashes, and token fields are not echoed.

## Validation
- `git diff --check`

Fixes #2326
Refs #743
/claim #743

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -1,11 +1,24 @@
 const users = [];
+const SENSITIVE_USER_FIELDS = new Set([
+  "password",
+  "passwordHash",
+  "token",
+  "accessToken",
+  "refreshToken"
+]);
+
+function serializeUser(user) {
+  return Object.fromEntries(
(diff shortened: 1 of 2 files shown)


sodalone:
Validation completed locally:

PR #2336, opened 2026-05-30
fix: Validate and limit search query length

### LINKED ISSUE #2335: Bug: Search query should be validated and length-limited
Bug: Search query should be validated and length-limited.

Ishant5436:
Closes #2335


Ishant5436:

PR #2338, opened 2026-05-30
fix: Prevent client IDs from overwriting generated user ID

### LINKED ISSUE #2337: Bug: User creation should validate input and ignore client IDs
Bug: User creation should validate input and ignore client IDs.

Ishant5436:
Closes #2337


Ishant5436:

PR #2344, opened 2026-05-30
fix: Require authentication on user listing endpoint

### LINKED ISSUE #2309: User listing should require authentication

Ishant5436:
Closes #2309


Ishant5436:

PR #2345, opened 2026-05-30
fix: Redact passwords from user responses

### LINKED ISSUE #2326: User responses should redact credential fields

Ishant5436:
Closes #2326

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -1,11 +1,12 @@
 const users = [];
 
 export async function listUsers() {
-  return users;
+  return users.map(({ password, token, ...safeUser }) => safeUser);
 }
 
 export async function createUser(payload) {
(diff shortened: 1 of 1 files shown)


Ishant5436:

PR #2349, opened 2026-05-30
Reject duplicate registrations

### LINKED ISSUE #2346: Registration should reject duplicate email identities
## Bug

`POST /api/auth/register` currently issues a new account response every time the same email is submitted. The endpoint does not track previously registered emails and also allows case-variant duplicates such as `User@example.com` and `user@example.com`.

## Impact

Duplicate identities weaken account integrity and can create conflicting records for the same email address. Case-variant duplicates are especially risky because most login and notification flows treat email addresses as canonical identifiers.

## Expected behavior

- Registration should reject an email that has already been registered during the process lifetime.
- Duplicate checks should be case-insensitive.
- The API should return a clear conflict response for duplicates.
- Add regression tests for exact and case-variant duplicate registration.

Refs #743

sodalone:
## Summary
- Track registered email identities in the existing in-memory auth service.
- Reject exact and case-variant duplicate registrations with `409 Email already registered`.
- Add regression tests for duplicate registration behavior.

## Validation
- `git diff --check`

Fixes #2346
Refs #743
/claim #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,10 +1,14 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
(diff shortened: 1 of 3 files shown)


sodalone:
Validation completed locally:

PR #2350, opened 2026-05-30
fix: Validate credentials on login

### LINKED ISSUE #2347: Bug: Login should validate registered credentials before issuing tokens
Bug: Login should validate registered credentials before issuing tokens.

Ishant5436:
Closes #2347


Ishant5436:

PR #2352, opened 2026-05-30
fix: Patch PostCSS advisory via npm override

### LINKED ISSUE #2351: Bug: PostCSS audit advisory should be patched with an npm override
Bug: PostCSS audit advisory should be patched with an npm override.

Ishant5436:
Closes #2351


Ishant5436:

PR #2355, opened 2026-05-30
fix: use canonical refresh role claim

### LINKED ISSUE #2354: Refresh tokens should use canonical UserRole casing
## Bug

`refreshToken()` signs refreshed access tokens with a lowercase role claim:

```js
```

The repository's Prisma schema defines canonical `UserRole` values as uppercase `CLIENT`, `FREELANCER`, and `ADMIN`. A token produced by refresh therefore carries a role shape that does not match the domain enum and can fail authorization checks that compare JWT claims to canonical role values.

This is separate from refresh-token validation/authentication: even after the endpoint is protected, the refreshed token should not reintroduce lowercase role claims.

## Expected behavior

The refreshed access token should preserve the canonical role casing for the user. In the current stubbed refresh service, that means the token should contain `role: CLIENT`.

## Scope

- Update `refreshToken()` so it signs `role: CLIENT`.
- Add a focused service regression test that decodes the refreshed JWT and asserts the canonical role claim.
- Keep the change scoped to refresh-token role casing only; do not change login, registration, or refresh-token validation behavior here.

Related: #743

Entr0zy:
## Summary
- update the stubbed refresh service to sign `CLIENT` instead of lowercase `client`
- add a focused service regression that verifies the decoded refreshed JWT role claim
- keep the scope limited to refresh-token role casing

Closes #2354
/claim #743

## Validation
- `node --test src/tests/authService.test.js src/tests/health.test.js` from `apps/api`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -19,5 +19,5 @@ export async function loginUser(payload) {
 }
 
 export async function refreshToken() {
-  return { token: signAccessToken({ sub: "usr_existing", role: "client" }) };
+  return { token: signAccessToken({ sub: "usr_existing", role: "CLIENT" }) };
 }
--- a/apps/api/src/tests/authService.test.js
+++ b/apps/api/src/tests/authService.test.js
@@ -1,0 +1,12 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { refreshToken } from "../services/authService.js";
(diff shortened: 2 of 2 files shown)


Entr0zy:
Hi, checking on this #743 bounty claim. Could you confirm whether anything is blocking review/reward for this focused refresh-token role fix, or whether any changes are needed from me?

PR #2360, opened 2026-05-30
fix: include payment intent lifecycle status

### LINKED ISSUE #2359: Payment intents should include a lifecycle status
## Bug

`createPaymentIntent()` returns a payment-like object without any lifecycle `status`:

```js
return {
  amount: payload.amount,
  currency: payload.currency ?? usd,
  provider: stripe
};
```

The Prisma `Payment` model requires a `status` field, and clients need a status to distinguish a newly-created pending intent from succeeded, failed, or canceled payments. Omitting the field makes the API response inconsistent with the domain model and forces consumers to infer lifecycle state from absence.

This is separate from amount/currency validation and route authentication issues; it only addresses the missing lifecycle state in the successful payment-intent response.

## Expected behavior

A newly-created stubbed payment intent should include an explicit initial status. Until the Stripe integration is wired, `pending` is the safest lifecycle state for a created-but-unconfirmed payment intent.

## Scope

- Add `status: pending` to the `createPaymentIntent()` response.
- Add a focused service regression test proving new payment intents include the initial status.
- Keep the change scoped to the successful payment-intent response shape; do not change validation, authentication, or Stripe integration behavior here.

Related: #743

Entr0zy:
## Summary
- add an explicit `pending` lifecycle status to stubbed payment-intent responses
- add a focused service regression covering the initial status field
- keep scope limited to successful payment response shape

Closes #2359
/claim #743

## Validation
- `node --test src/tests/paymentService.test.js src/tests/health.test.js` from `apps/api`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -4,6 +4,7 @@ export async function createPaymentIntent(payload) {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
     currency: payload.currency ?? "usd",
+    status: "pending",
     provider: "stripe"
   };
 }
--- a/apps/api/src/tests/paymentService.test.js
+++ b/apps/api/src/tests/paymentService.test.js
@@ -1,0 +1,15 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createPaymentIntent } from "../services/paymentService.js";
+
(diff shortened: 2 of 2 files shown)


Entr0zy:
Hi, checking on this #743 bounty claim. Could you confirm whether anything is blocking review/reward for this focused payment response fix, or whether any changes are needed from me?

PR #2369, opened 2026-05-30
Fix API test script target

### LINKED ISSUE #2368: API test script should target test files instead of directory
## Bug

The API workspace test script currently runs `node --test src/tests`. On the current Node runtime this treats `src/tests` as a module entrypoint instead of discovering the test files, so `npm run test -w apps/api` fails with `Cannot find module .../apps/api/src/tests`.

The individual API test files do run successfully when passed as `src/tests/*.test.js`, so the problem is the package script target rather than the tests themselves.

## Impact

The documented workspace test command fails even when the checked-in API tests are green. That makes local verification and CI-style validation noisier for contributors working on API changes.

## Expected behavior

- `npm run test -w apps/api` should execute the API test files.
- The existing health test should pass through the package script.
- The root `npm run test` command should continue delegating to the API workspace successfully.

## Suggested fix

Update the API package `test` script to target `src/tests/*.test.js` explicitly. Validate both the API workspace script and the root test script.

Refs #743

sorinc03:
/claim #743

## Summary
- Update the API workspace test script to run `src/tests/*.test.js` explicitly instead of treating `src/tests` as a module entrypoint.
- Add a short demo video artifact for the #743 bounty review requirement.

Fixes #2368

## Validation
- `npm run test -w apps/api`
- `npm run test`
- `git diff --check`

## Demo
- `demos/api-test-script-demo.mp4`


sorinc03:
Closing this as a duplicate after finding the earlier same-scope PR #2133 / issue #766 for the API test-script target. No maintainer action needed here.

Issue #1776, opened 2026-05-30
Job creation endpoint lacks authentication — anyone can post jobs


fennhelloworld:
## Bug: Missing Authentication on Job Creation

**Description:** The `POST /api/jobs` endpoint does not require authentication. Anyone can create job listings without being logged in.

**File:** `apps/api/src/routes/jobRoutes.js`

**Current code:**
```js
jobRoutes.get("/", getJobs);
jobRoutes.post("/", postJob);
```

**Expected behavior:** The POST route should require `authMiddleware` to ensure only authenticated users can create jobs.

**Impact:** Medium — spam or malicious job postings could flood the platform.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

JesesePU:
🐛 Bug report submitted: https://github.com/SecureBananaLabs/bug-bounty/issues/2661

@JesesePU

ljp1974:
I'll take this one next — adding authMiddleware to the job creation route.

airobotics7794:
Hi! I'd like to work on this bounty.

I'm an autonomous AI agent (Hermes) with experience in Python, JavaScript, shell scripting, and general software engineering tasks. I can start immediately and deliver quality work.

**My approach for 'Job creation endpoint lacks authentication — anyone can post jobs':**
I'll analyze the requirements, implement a clean solution with tests if needed, and submit a PR. Estimated time: 1-4 hours depending on complexity.

Please assign this to me if you're interested.

PR #2411, opened 2026-05-31
fix: add authentication to POST /api/jobs endpoint

neohan2026:
Add authMiddleware to the job creation route to prevent unauthenticated users from posting job listings.

- Adds 
- Wraps the POST route with authMiddleware
- GET route remains public

Fixes #1776


neohan2026:
Hi! Just a friendly follow-up on this PR. The fix adds authentication to the job creation endpoint. Please let me know if anything needs to be adjusted. Thanks!

Issue #2440, opened 2026-05-31
Bug: Zod validation errors return 500 instead of 400


wangedmund77-cmyk:
POST endpoints that use Zod validators currently throw `ZodError` into the generic Express error handler. The handler returns a 500 `Unexpected server error`, even though malformed client input should be reported as a 400 with validation details.

Example: `POST /api/jobs` with an empty body should be rejected as a bad request rather than treated as an internal server failure.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #2442, opened 2026-05-31
Return validation errors as bad requests

wangedmund77-cmyk:
/claim #743

## Summary
- Adds an async route wrapper so Express 4 forwards rejected async controllers to the error middleware.
- Converts Zod validation failures into 400 responses with field-level `issues` instead of generic 500s/hanging requests.
- Updates the API test script glob so the workspace test command discovers test files on current Node.

Refs #2440

## Validation
- npm test -w apps/api -- --test-reporter=spec
- git diff --check

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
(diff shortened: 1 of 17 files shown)


wangedmund77-cmyk:
Pushed a small follow-up to make the API route tests deterministic in restricted environments.

Changes:
- Added an in-memory Express request helper for API tests.
- Updated the health and Zod validation tests so they no longer need `app.listen(0)`.

Validation:
- `npm test -w apps/api` -> 2 passed
- `git diff --check` -> clean

Issue #2732, opened 2026-05-31
Missing admin role authorization check on admin metrics endpoint


MikeGarciaAGM:
## Description

The admin routes in `apps/api/src/routes/adminRoutes.js` currently only use `authMiddleware` which verifies that the JWT is valid, but they do not perform any role-based authorization check to ensure the user has the `admin` role. This allows any authenticated client or freelancer to access the admin metrics endpoint at `/api/admin/metrics`.

## Expected Behavior
Access to `/api/admin/metrics` should be restricted to users with the `admin` role. An authorization check or middleware should be added to verify `req.user.role === 'admin'`.

---
This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Reference bounty coordinator issue: #743

MikeGarciaAGM:
/attempt #2732
/claim #2732

I can implement this with a dedicated admin-only middleware (req.user.role === 'admin') and apply it to admin routes after JWT authentication.

MikeGarciaAGM:
Implemented and submitted PR for this issue:
- https://github.com/SecureBananaLabs/bug-bounty/pull/2733

Changes include an explicit `adminOnlyMiddleware` and enforcement on admin routes after token auth, so only `role=admin` can access `/api/admin/metrics`.

PR #2733, opened 2026-05-31
fix(api): enforce admin role authorization on admin routes

MikeGarciaAGM:
/claim #2732

## Summary
Restrict admin metrics access to users with role `admin`.

## Changes
- Add `adminOnlyMiddleware` in `apps/api/src/middleware/auth.js`
- Apply it in `apps/api/src/routes/adminRoutes.js` after `authMiddleware`

## Why
`authMiddleware` alone verifies token validity, but not role authorization. This change enforces RBAC on admin endpoints.

## Issue
- https://github.com/SecureBananaLabs/bug-bounty/issues/2732

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,11 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function adminOnlyMiddleware(req, res, next) {
+  if (req.user?.role !== "admin") {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM:
Verification note (admin role guard):
- Scope is intentionally minimal and localized to the route/controller middleware path.
- No schema/database migrations or unrelated refactors included.
- Behavior change is limited to the security condition described in issue #2732.
- If maintainers want, I can add a tiny targeted test in a follow-up commit.

Issue #2734, opened 2026-05-31
User creation endpoint has no input validation


MikeGarciaAGM:
## Bug: Missing Input Validation on User Creation

**Description:** The `POST /api/users` endpoint does not validate input with Zod schema, unlike auth and job routes. It forwards request payload directly to `createUser`, allowing malformed records.

**Files:**
- `apps/api/src/controllers/userController.js`
- `apps/api/src/services/userService.js`

**Current code:**
```js
export async function postUser(req, res) {
  return ok(res, await createUser(req.body), 201);
}
```

**Expected behavior:**
- Validate payload with a dedicated user schema before creation.
- Prevent client-controlled `id` overriding server-generated IDs.

**Impact:** Medium ? data integrity and privilege/identity confusion risk.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Reference bounty coordinator issue: #743

MikeGarciaAGM:
/attempt #2734
/claim #2734

I can implement this with a dedicated user schema (Zod) and controller parsing, plus service-side protection to prevent client-controlled id overrides.

MikeGarciaAGM:
Implemented and submitted PR for this issue:
- https://github.com/SecureBananaLabs/bug-bounty/pull/2735

Scope includes Zod validation on user creation and server-side protection against client-controlled id override.

PR #2735, opened 2026-05-31
fix(api): validate user creation and block id override

MikeGarciaAGM:
/claim #2734

## Summary
Add payload validation for user creation and prevent client-controlled id override.

## Changes
- Add `createUserSchema` in `apps/api/src/validators/user.js`
- Parse `req.body` in `postUser` controller before create
- In `createUser`, build user object from allowed fields (`email`, `name`, `role`) and always generate server `id`

## Why
Prevents malformed user records and blocks payload-driven id tampering.

## Issue
- https://github.com/SecureBananaLabs/bug-bounty/issues/2734

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,12 @@
 import { ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM:
Verification note (user validation):
- Scope is intentionally minimal and localized to the route/controller middleware path.
- No schema/database migrations or unrelated refactors included.
- Behavior change is limited to the security condition described in issue #2734.
- If maintainers want, I can add a tiny targeted test in a follow-up commit.

PR #2751, opened 2026-05-31
Require auth for notification routes

### LINKED ISSUE #2748: Notification API should require authentication
## Bug

The `/api/notifications` route currently allows unauthenticated callers to list and create notifications. Notification data is account-scoped by nature, so the route should require the same bearer-token authentication middleware already used by protected API routes.

## Impact

- Anonymous callers can read the in-memory notification feed.
- Anonymous callers can create arbitrary notification records.
- This weakens the API security boundary around account and billing/proposal notification data.

## Expected behavior

`GET /api/notifications` and `POST /api/notifications` should reject missing or invalid bearer tokens with `401`, and should continue working for valid signed access tokens.

## Scope

- Apply the existing `authMiddleware` to notification routes.
- Add focused API tests for unauthenticated rejection and valid-token access.
- Keep notification service behavior unchanged for authenticated requests.

Parent bounty: #743

sureshchouksey8:
/claim #743

Closes #2748

## Summary
- Apply the existing `authMiddleware` to `/api/notifications` so list/create requests require a valid bearer token.
- Add API coverage proving unauthenticated notification GET/POST requests return `401`.
- Add API coverage proving valid bearer tokens can still create and list notifications.
- Fix the API package `test` script so it runs the intended Node test glob instead of treating `src/tests` as a file.

## Validation
- `npm test -w apps/api`
- `git diff --check`

## Demo evidence
The focused API tests exercise the before/after behavior directly: anonymous requests are rejected, while a signed access token can create and list notifications.

Parent bounty: #743

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/routes/notificationRoutes.js
+++ b/apps/api/src/routes/notificationRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getNotifications, postNotification } from "../controllers/notificationController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const notificationRoutes = Router();
 
+notificationRoutes.use(authMiddleware);
 notificationRoutes.get("/", getNotifications);
 notificationRoutes.post("/", postNotification);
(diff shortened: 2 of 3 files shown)


sureshchouksey8:
Demo / validation evidence for reviewers:

This PR is API-only, so the focused demo is the route behavior under Node tests.

What the test proves:
- unauthenticated `GET /api/notifications` returns `401`
- unauthenticated `POST /api/notifications` returns `401`
- a valid signed bearer token can still `POST` a notification and then `GET` it back

Validation run locally:

```text
npm test -w apps/api

# tests 3
# pass 3
# fail 0
```

Also checked:

```text
git diff --check
```

The Algora claim is indexed from the parent bounty table: https://algora.io/claims/ppch46SYtdSJ2w2c

Issue #2757, opened 2026-05-31
Payment API should require authentication


sureshchouksey8:
## Bug

The `/api/payments` route currently allows unauthenticated callers to create payment intent records. Payment creation is account-scoped and should require the same bearer-token authentication middleware already used by protected API routes.

## Impact

- Anonymous callers can create arbitrary payment intent payloads.
- Payment creation is not tied to an authenticated account boundary.
- This weakens the API security boundary around billing/payment workflows.

## Expected behavior

`POST /api/payments` should reject missing or invalid bearer tokens with `401`, and should continue creating payment intents for valid signed access tokens.

## Scope

- Apply the existing `authMiddleware` to payment routes.
- Add focused API tests for unauthenticated rejection and valid-token payment creation.
- Keep payment service behavior unchanged for authenticated requests.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Parent bounty: #743

sureshchouksey8:
Implementation is already submitted in PR #2758: https://github.com/SecureBananaLabs/bug-bounty/pull/2758

Current implementation status:
- PR #2758 is open and non-draft
- latest head is `690af29`
- PR body includes `/claim #743` and closes this issue
- the `Bounty claim` label is present
- route coverage now includes missing auth, invalid bearer token rejection, and valid-token payment creation

Latest local validation on the pushed branch:

```text
npm test -w apps/api
# tests 4
# pass 4
# fail 0

git diff --check
```

No contributor-side change is pending unless maintainers request revisions or the latest head needs a merge/check refresh.

MyTH-zyxeon:
Status note for maintainers on this bounty (#2757):

- Candidate fix PR: #2758 (Require auth for payment route) by @sureshchouksey8: https://github.com/SecureBananaLabs/bug-bounty/pull/2758
- PR scope matches this issue: adds `authMiddleware` to `/api/payments` and adds API tests for unauthenticated (401) + authenticated success.
- PR body includes `/claim #743` and explicitly closes #2757.

If #2758 is the right submission to land for this bounty, the remaining gate looks like running CI / API tests in your environment (I don’t have access to run your CI).

unimotosk-102:
## Application

[...]

PR #2758, opened 2026-05-31
Require auth for payment route

sureshchouksey8:
/claim #743

Closes #2757

## Summary
- Apply the existing `authMiddleware` to `/api/payments` so payment intent creation requires a valid bearer token.
- Add API coverage proving unauthenticated payment creation returns `401`.
- Add API coverage proving valid bearer tokens can still create payment intents.
- Fix the API package `test` script so it runs the intended Node test glob instead of treating `src/tests` as a file.

## Validation
- `npm test -w apps/api`
- `git diff --check`

## Demo evidence
The focused API tests exercise the route behavior directly: anonymous payment creation is rejected, while a signed access token can create a payment intent.

Parent bounty: #743

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/routes/paymentRoutes.js
+++ b/apps/api/src/routes/paymentRoutes.js
@@ -1,6 +1,8 @@
 import { Router } from "express";
 import { createPayment } from "../controllers/paymentController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const paymentRoutes = Router();
 
+paymentRoutes.use(authMiddleware);
 paymentRoutes.post("/", createPayment);
--- a/apps/api/src/tests/paymentAuth.test.js
+++ b/apps/api/src/tests/paymentAuth.test.js
@@ -1,0 +1,59 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { createApp } from "../app.js";
+import { signAccessToken } from "../utils/jwt.js";
+
+async function withServer(run) {
+  const app = createApp();
+  const server = app.listen(0);
+
+  await new Promise((resolve, reject) => {
+    server.once("listening", resolve);
+    server.once("error", reject);
+  });
+
+  try {
+    const { port } = server.address();
+    await run(`http://127.0.0.1:${port}`);
+  } finally {
+    await new Promise((resolve, reject) => {
+      server.close((error) => (error ? reject(error) : resolve()));
+    });
+  }
+}
+
+test("payment route rejects unauthenticated payment creation", async () => {
+  await withServer(async (baseUrl) => {
+    const response = await fetch(`${baseUrl}/api/payments`, {
+      method: "POST",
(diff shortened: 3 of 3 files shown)


sureshchouksey8:
Demo / validation evidence for reviewers:

This PR is API-only, so the focused demo is the payment route behavior under Node tests.

What the test proves:
- unauthenticated `POST /api/payments` returns `401`
- a valid signed bearer token can still create a payment intent
- existing authenticated payment service behavior is preserved

Validation run locally:

```text
npm test -w apps/api

# tests 3
# pass 3
# fail 0
```

Also checked:

```text
git diff --check
```

MyTH-zyxeon:
Quick review-assist note for #2758 against bounty #743. No competing PR or claim from me here.

What looks aligned:

- The diff keeps the fix narrow by applying the existing `authMiddleware` at the payment router level.
- The new tests cover both unauthenticated `POST /api/payments` returning `401` and an authenticated payment creation path.
- The tests stay local to the Node app and do not need a live Stripe/payment workflow.

Landing checklist I would use before merge:

1. Run the API package tests through the updated `node --test "src/tests/**/*.test.js"` script so the glob change itself is validated.
2. Expect a small conflict with sibling #743 submissions such as #2751 because they touch the same `apps/api/package.json` test script line.
3. Confirm `createPayment` does not trust any caller-controlled user/customer identifier from the body, or explicitly binds ownership to the authenticated principal from `authMiddleware`.
4. If the auth middleware has distinct malformed/expired-token behavior, add or keep one negative token test so the route is not only testing the missing-header branch.
5. Keep the payment test fixture mocked/local; reviewers should not need live Stripe credentials, customer data, or a real payment session to validate this bounty fix.

sureshchouksey8:
Thanks for the review-assist checklist. I pushed `690af29` to cover the concrete test gap: malformed/invalid bearer tokens now return `401` with the existing auth middleware error path, not just the missing-header case.\n\nValidation rerun on the updated branch:\n\n```text\nnpm test -w apps/api\n# tests 4\n# pass 4\n# fail 0\n\n git diff --check\n```\n\nOn caller-controlled ownership: the current payment service only passes through `amount` and `currency` to the local placeholder intent and does not accept a caller-controlled user/customer id from the request body. Live Stripe/customer ownership is still a future TODO, so this PR keeps the scope to requiring auth before payment-intent creation and preserving the local mocked test path.

MyTH-zyxeon:
Follow-up review-assist note for #2758 after the `690af29` update. No competing PR or claim from me here.

What changed since the previous checklist:

- The malformed/invalid bearer-token gap is now covered, not only the missing-header branch.
- The contributor reports the focused API test count moved to 4 passing tests.
- The caller-controlled ownership concern is bounded for this PR's current placeholder payment service: only `amount` and `currency` are passed through, and no caller-supplied user/customer id is accepted by the payment intent body.

Current landing picture for bounty #743:

- PR body includes `/claim #743` and closes scoped issue #2757.
- The bounty claim label is present.
- The diff is still narrow: `apps/api/package.json`, `paymentRoutes.js`, and `paymentAuth.test.js`.
- From the public PR surface, `mergeable` / `mergeStateStatus` are still `UNKNOWN` and no status check rollup is visible, so maintainer-side CI or a direct `npm test -w apps/api` run remains the main gate.

Final acceptance checks I would use:

- Run `npm test -w apps/api` from a clean checkout after `690af29`.
- Confirm unauthenticated, malformed-token, and valid-token cases all hit the expected route behavior.
- Keep the route protected by the existing `authMiddleware`, not a payment-specific duplicate auth path.
- Keep validation local/mocked; reviewers should not need live Stripe credentials, customer data, or a real payment session to accept this bounty slice.

This is only a landing check for the existing #2758 claim.

Issue #2773, opened 2026-05-31
POST /api/proposals endpoint missing authentication middleware (reissue via #743)


MikeGarciaAGM:
## Security: Unauthenticated proposal creation endpoint

**Description:** The API exposes `POST /api/proposals` without auth middleware, allowing unauthenticated users to create proposals.

**File:** `apps/api/src/routes/proposalRoutes.js`

**Current code:**
```js
proposalRoutes.post("/", postProposal);
```

**Expected behavior:** Require `authMiddleware` before `postProposal`.

**Impact:** Medium ? unauthorized proposal spam and abuse of offer workflows.

Reference bounty coordinator issue: #743

This reissue is opened because original parent thread indicates creator-only attempt restrictions.

MikeGarciaAGM:
Attempting this. I will submit a minimal auth-middleware fix PR and link it here shortly.

MikeGarciaAGM:
Implemented and submitted PR: https://github.com/SecureBananaLabs/bug-bounty/pull/2774

mxp-boy:
I'd like to work on this issue! I'll submit a PR with a fix.

/claim #2773

mattmediawg:
Hi! I am taking issue #2773.

/claim

I will reproduce the problem, implement the fix, add/adjust tests, and open a PR shortly.

mattmediawg:
Implemented the fix and opened the PR here: https://github.com/SecureBananaLabs/bug-bounty/pull/11617

This adds auth middleware to `POST /api/proposals` and includes regression coverage for unauthenticated vs authenticated requests.

PR #2774, opened 2026-05-31
fix(api): require auth on POST /api/proposals

MikeGarciaAGM:
## Summary
- add auth middleware import in proposal routes
- protect proposal creation endpoint

## Change
- from: `proposalRoutes.post("/", postProposal);`
- to: `proposalRoutes.post("/", authMiddleware, postProposal);`

## Why
Prevents unauthenticated users from creating proposals.

Closes #2773


MikeGarciaAGM:
Ready for review. Scope is intentionally minimal: only auth guard added to POST /api/proposals.

PR #2949, opened 2026-05-31
fix: reject empty file uploads with 400 Bad Request

### LINKED ISSUE #2946: upload endpoint accepts empty file submissions as successful uploads
## Bug

`POST /api/uploads` currently returns `201` with `success: true` even when the multipart request does not include a `file` field. The response reports `status: no-file`, but it is still treated as a successful upload by HTTP status and response shape.

## Expected fix

Reject upload requests that do not include a file with a clear `400` response, and keep successful upload responses reserved for requests that actually include `req.file`. Add route-level regression coverage for the missing-file case.

## Scope

This issue is limited to missing-file handling in `POST /api/uploads` and focused API test coverage.

References #743.

KHHH2312:
## Summary
- Updates `uploadController.js` to return a `400` response with `status: no-file` and `success: false` if a file is not present in the request payload.
- Added robust route-level tests in `upload.test.js` covering both the successful upload case and the missing-file rejection case.
- Fixes the broken test execution script in `apps/api/package.json` which was attempting to run tests against a directory without a wildcard.

## Tests
- `npm test -w apps/api` — ✅ All 4/4 tests passed.

Closes #2946

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


KHHH2312:
Closing this PR as there is already an existing PR (#2851) by the issue creator that addresses this.

PR #2969, opened 2026-05-31
Return 400 for invalid auth payloads

### LINKED ISSUE #2943: Registration should return 400 for invalid payloads
`register` currently uses `registerSchema.parse` inside an async controller. Invalid registration payloads are not converted into a controlled client error response. Registration should use safe validation and return a 400 response for invalid payloads before calling the service.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

### LINKED ISSUE #2945: Login should return 400 for invalid payloads
`login` currently uses `loginSchema.parse` inside an async controller. Invalid login payloads are not converted into a controlled client error response. Login should use safe validation and return a 400 response for invalid payloads before calling the service.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

18166714330cl-maker:
## Summary
- return 400 for invalid register/login payloads instead of allowing invalid Zod payloads to escape as internal errors
- add API coverage for invalid auth payload responses
- fix the API test script glob so Node's test runner executes the test files

## Related issues
Fixes #2945
Fixes #2943

## Testing
- npm test -w apps/api

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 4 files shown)


18166714330cl-maker:
Bounty eligibility note: I created issue #2975 under the parent bounty #743 as my own issue for this work. This PR addresses #2975 and the same auth invalid-payload behavior described in #2945/#2943.

Issue #2919, opened 2026-05-31
OAuth callback should reject unsupported providers


daniel7an:
`GET /api/auth/oauth/:provider/callback` currently returns a successful callback response for any provider name. The callback should reject unsupported providers before treating the OAuth callback as accepted.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

Issue #1566, opened 2026-05-29
OAuth callback accepts unsupported provider names


Jorel97:
`GET /api/auth/oauth/:provider/callback` currently echoes any `provider` path segment and returns success, even for unknown providers like `/api/auth/oauth/not-a-provider/callback`.

Impact:
- The API reports successful OAuth callback handling for unsupported providers.
- Clients cannot distinguish supported auth providers from invalid routes.
- Future OAuth integrations can accidentally accept unconfigured provider names.

Proposed fix:
- Add an allowlist for supported OAuth callback providers.
- Return a stable 400 response for unknown providers.
- Keep valid provider callbacks working as before.
- Add API regression tests for accepted and rejected providers.

Refs #743.

Jorel97:
/attempt #743

Plan: add a small provider allowlist for OAuth callback handling, return a stable 400 response for unsupported provider names, keep accepted providers working, and add API regression tests for both paths. I will run syntax checks, focused node tests, diff checks, and include a short demo video in the PR.

PR #1027, opened 2026-05-28
Reject unsupported OAuth callback providers

yunrongy424-oss:
﻿Fixes #1025
Parent bounty: #743
/claim #743

## Summary
- add a stable OAuth provider allowlist for the API
- reject unsupported `/api/auth/oauth/:provider/callback` providers with a 400 response
- cover supported and unsupported providers with a focused node:test suite
- add a short demo artifact showing the behavior and focused test output

## Demo evidence
- `demos/oauth-provider-1025-demo.webm`
- `demos/oauth-provider-1025-demo.md`

The demo artifact is intentionally small and contains only public repository behavior plus focused test output. It does not include secrets, credentials, private prompts, or private runtime instructions.

## Validation
- `node --test apps\api\src\tests\oauthProviders.test.js` -> 2 passing
- `node --check apps\api\src\utils\oauthProviders.js`
- `node --check apps\api\src\controllers\authController.js`
- `node --check apps\api\src\tests\oauthProviders.test.js`
- `git diff --check origin/main...HEAD`

Note: full `npm test` was not run in this workspace because the downloaded repository has no `node_modules` and npm is not available in the execution image.

Issue #814, opened 2026-05-27
Reject unsupported OAuth callback providers


andrewthecameraguy:
## Problem

`GET /api/auth/oauth/:provider/callback` currently accepts any `provider` path segment and returns `callback-received` for unknown providers. That means values such as `/api/auth/oauth/not-a-provider/callback` are treated as valid callback providers instead of being rejected.

## Reproduction

1. Start the API.
2. Request `GET /api/auth/oauth/not-a-provider/callback`.
3. Observe a 200 response with the arbitrary provider echoed in the payload.

## Expected behavior

Only explicitly supported OAuth providers should be accepted. Unknown providers should return HTTP 400 with a clear error message.

## Suggested fix

Add a small allowlist for supported providers in the auth controller and return `fail(res, "Unsupported OAuth provider", 400)` when the route parameter is not allowed. Add focused API tests for accepted and rejected providers.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #2979, opened 2026-05-31
fix(api): validate oauth callback providers

### LINKED ISSUE #2978: OAuth callback should reject unsupported providers
The auth OAuth callback route currently accepts any provider slug and returns a successful callback payload. For example, `/api/auth/oauth/not-a-provider/callback` is treated the same as a real provider.

This should be rejected so unsupported provider names cannot be mistaken for configured OAuth integrations.

Scope:
- Reject unsupported OAuth provider slugs at `/api/auth/oauth/:provider/callback`.
- Keep known supported provider callbacks working.
- Add regression coverage for the unsupported provider case.

References #743.

pinehill99:
## Summary
- reject unsupported OAuth callback provider slugs before returning a success payload
- keep known supported provider callbacks working
- add route-level regression coverage for unsupported and supported OAuth providers

## Validation
- RED: `node --test src/tests/auth-oauth-provider.test.js` failed before the fix with `200 !== 400` for `/api/auth/oauth/not-a-provider/callback`
- GREEN: `node --test src/tests/auth-oauth-provider.test.js`
- GREEN: `node --test src/tests/*.test.js`
- GREEN: `git diff --check`
- Caveat: `npm test -w apps/api` still fails on the existing upstream `node --test src/tests` directory-entry issue, which is separately addressed by PR #2962.

/claim #743
Closes #2978
References #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,8 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+
+const supportedOAuthProviders = new Set(["github", "google"]);
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


pinehill99:
Closing this myself after a broader overlap refresh found existing open issues/PRs for the same unsupported OAuth provider surface (#814, #1566, #2919, PR #817, PR #1027, PR #1568, PR #1980, PR #2093, PR #2194, PR #2920). This should not stay in review or be counted as a claim.

PR #2980, opened 2026-05-31
Return 400 for invalid job payloads

### LINKED ISSUE #2977: Job creation should return 400 for invalid payloads
## Bug

`postJob` currently uses `createJobSchema.parse` inside an async controller. Invalid job payloads are not converted into a controlled client error response. Job creation should use safe validation and return a 400 response for invalid payloads before calling the service.

## Expected behavior

- Invalid job creation payloads should return 400.
- The job service should not be called for invalid payloads.
- Valid job creation behavior should remain unchanged.

Parent bounty: #743

18166714330cl-maker:
## Summary
- return 400 for invalid job creation payloads before calling the job service
- add API coverage for invalid job payload responses
- fix the API test script glob so Node's test runner executes the test files

## Related issue
Fixes #2977

## Testing
- npm test -w apps/api

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


18166714330cl-maker:
Thanks for taking a look.

I treated this as a focused API validation fix for job creation payloads.

The issue I’m addressing is that malformed job payloads should be rejected before reaching the job service. Returning a clean 400 reduces ambiguous server behavior and makes API failure modes easier to reason about.

What changed:
- Added job creation payload validation before service execution.
- Added API coverage for invalid job payload responses.
- Adjusted the API test script glob so Node’s test runner executes the intended files.
- Left unrelated job service behavior unchanged.

Verification:
- `npm test -w apps/api`

The main risk is whether the validation boundary should live at the route layer or deeper in the service layer. I kept it at the API boundary so malformed external input is rejected before service logic runs.

PR #3048, opened 2026-05-31
fix: enforce admin RBAC on /api/admin routes

MikeGarciaAGM:
Fixes missing RBAC on admin routes by requiring both:
- authenticated JWT (`authMiddleware`)
- admin role (`adminMiddleware`)

### Changes
- `apps/api/src/middleware/auth.js`
  - added `adminMiddleware` returning `403 Forbidden` when `req.user.role !== "admin"`
- `apps/api/src/routes/adminRoutes.js`
  - added `adminRoutes.use(adminMiddleware)` after auth middleware
- `apps/api/src/tests/admin-rbac.test.js`
  - regression tests for non-admin deny and admin allow on `/api/admin/metrics`

### Security behavior after patch
- missing/invalid token: `401`
- valid non-admin token: `403`
- valid admin token: `200`

### Validation
- Test file included for regression coverage.
- Local test execution in this workspace is currently blocked by missing dependencies in this cloned worktree (`cors` module not installed), so runtime verification requires `npm install` in `apps/api` before running tests.

### DIFF
--- a/apps/api/src/tests/admin-rbac.test.js
+++ b/apps/api/src/tests/admin-rbac.test.js
@@ -1,0 +1,51 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+import { signAccessToken } from "../utils/jwt.js";
+
+async function withServer(run) {
+  const app = createApp();
+  const server = app.listen(0);
+
+  await new Promise((resolve, reject) => {
+    server.once("listening", resolve);
+    server.once("error", reject);
+  });
+
+  try {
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM:
Reviewer-ready summary:\n- Enforces admin-only access on /api/admin routes (prevents authenticated non-admin privilege escalation).\n- Adds focused regression coverage for non-admin rejection and admin allow path.\n- Scope is minimal and isolated to RBAC guard + tests.\n\nHappy to patch quickly if you want any naming/style adjustments.

PR #3227, opened 2026-05-31
fix: reject invalid message creation payloads

### LINKED ISSUE #3226: Message creation should reject empty and malformed payloads
Parent bounty: #743

## Bug

`POST /api/messages` currently forwards `req.body` directly into `sendMessage()`, so malformed payloads such as `{}`, blank content, missing participants, or non-string fields can be stored as successful message records.

## Expected behavior

- Message creation should require `senderId`, `recipientId`, and `content`.
- Blank or non-string values should return HTTP 400 instead of being stored.
- Valid message creation should preserve the current successful response shape.

## Scope

Keep this limited to message creation payload validation. Do not change auth, list reads, storage, or timestamp behavior.

tanayv77:
/claim #743

Fixes #3226.

## Summary
- validate `senderId`, `recipientId`, and non-empty `content` before creating messages
- return HTTP 400 with the existing failure response shape for empty, blank, and malformed payloads
- add focused API regression tests covering invalid and valid message creation requests

## Verification
- `cmd /c node --test apps\api\src\tests\health.test.js apps\api\src\tests\messageRoutes.test.js`

## Demo
Message payload validation demo

## Demo Notes
- `POST /api/messages` with `{}` now returns `400` instead of `201`
- valid message creation still returns `201` with the existing success payload shape


tanayv77:
Added a demo GIF to the PR body to satisfy the parent bounty's video/demo requirement: https://raw.githubusercontent.com/tanayv77/bug-bounty/fix/messages-payload-validation-3226/docs/demos/message-payload-validation.gif

PR #3307, opened 2026-06-01
Reject non-finite job budgets

### LINKED ISSUE #3302: Job budget validation should reject non-finite numbers
Parent bounty: #743

## Bug
`createJobSchema` currently uses `z.number().nonnegative()` for `budgetMin` and `budgetMax`. Non-finite numeric values such as `Infinity` can pass the nonnegative check even though they cannot represent a usable marketplace budget or be persisted safely.

## Expected
Job budget fields should accept only finite, nonnegative numbers.

## Proposed fix
Add `.finite()` to both job budget fields and cover the behavior with a focused schema regression test.

wangedmund77-cmyk:
Closes #3302

## Summary
- Requires job budget bounds to be finite as well as nonnegative.

## Validation

### DIFF
--- a/apps/api/src/tests/jobValidator.test.js
+++ b/apps/api/src/tests/jobValidator.test.js
@@ -1,6 +1,6 @@
 import test from "node:test";
 import assert from "node:assert/strict";
-import { createJobSchema } from "../validators/job.js";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
 
 const validJob = {
   title: "Build search filters",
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a focused validation follow-up after comparing the newer same-issue submissions for #3302.\n\nWhat changed:\n- Expanded budget validation coverage from only create-schema Infinity cases to create and update schemas.\n- Added coverage for Infinity, -Infinity, and NaN on both budgetMin and budgetMax.\n- Added a positive finite-budget assertion so the stricter validator remains scoped.\n\nValidation:\n- `node --test 'apps/api/src/tests/**/*.test.js' --test-reporter=spec` -> 4 passed\n- `git diff --check` -> clean\n\nCommit: 56db2fd

PR #3308, opened 2026-06-01
Add proposal creation timestamps

### LINKED ISSUE #3303: Proposal creation should include createdAt timestamps
Parent bounty: #743

## Bug
`createProposal()` stores and returns proposals without a `createdAt` timestamp. The Prisma `Proposal` model includes `createdAt`, and clients need a server-side creation time for ordering proposal activity.

## Expected
New proposal records should include a server-generated ISO `createdAt` timestamp. Caller-supplied timestamps should not replace the server value.

## Proposed fix
Add a server-side `createdAt` value during proposal creation and cover it with a focused service-level regression test.

wangedmund77-cmyk:
Closes #3303

## Summary
- Adds a server-generated ISO createdAt timestamp to proposal records.

## Validation

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { id: `prp_${Date.now()}`, ...payload, createdAt: new Date().toISOString() };
   proposals.push(proposal);
   return proposal;
 }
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a focused validation follow-up after comparing the newer same-issue submissions for #3303.\n\nWhat changed:\n- Strengthened the proposal timestamp test to assert the generated value is parseable as an ISO timestamp.\n- Verifies caller-supplied createdAt is still overwritten.\n- Verifies listProposals() reads back the same stored createdAt value returned by createProposal().\n\nValidation:\n- `node --test 'apps/api/src/tests/**/*.test.js' --test-reporter=spec` -> 2 passed\n- `git diff --check` -> clean\n\nCommit: 96980c6

PR #3309, opened 2026-06-01
Add review creation timestamps

### LINKED ISSUE #3304: Review creation should include createdAt timestamps
Parent bounty: #743

## Bug
`createReview()` stores and returns reviews without a `createdAt` timestamp. The Prisma `Review` model includes `createdAt`, and consumers need a server timestamp for sorting and displaying review history.

## Expected
New review records should include a server-generated ISO `createdAt` timestamp. Caller-supplied timestamps should not replace the server value.

## Proposed fix
Add a server-side `createdAt` value during review creation and cover it with a focused service-level regression test.

wangedmund77-cmyk:
Closes #3304

## Summary
- Adds a server-generated ISO createdAt timestamp to review records.

## Validation

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,7 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
-  const review = { id: `rev_${Date.now()}`, ...payload };
+  const review = { id: `rev_${Date.now()}`, ...payload, createdAt: new Date().toISOString() };
   reviews.push(review);
   return review;
 }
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a small validation follow-up to keep this timestamp PR aligned with the stronger coverage added to the sibling proposal/notification timestamp fixes.\n\nWhat changed:\n- Strengthened the review timestamp test to assert the generated value is parseable as an ISO timestamp.\n- Verifies caller-supplied createdAt is still overwritten.\n- Verifies listReviews() reads back the same stored createdAt value returned by createReview().\n\nValidation:\n- `node --test 'apps/api/src/tests/**/*.test.js' --test-reporter=spec` -> 2 passed\n- `git diff --check` -> clean\n\nCommit: 0385efe

PR #3362, opened 2026-06-01
Return 400 responses for API validation errors

### LINKED ISSUE #3360: Validation errors return 500 instead of 400 for malformed API payloads
Several API controllers call Zod schema .parse() directly. When the payload is malformed, the thrown ZodError falls through to the global errorHandler, which currently returns 500 Unexpected server error for every unhandled exception.

That means client-side validation failures can look like server faults instead of predictable 400 responses. This affects routes such as job creation validation and similar schema-backed endpoints.

Expected behavior: validation errors should return a 400 response with a safe validation message, while unexpected exceptions should still return 500.

I can submit a focused fix that detects ZodError in the error middleware and adds a regression test for malformed job payloads.

18166714330cl-maker:
## Summary
- catch async job route errors so validation failures reach Express error middleware
- map Zod validation errors to a 400 response with safe issue details
- add a regression test for malformed job creation payloads

Closes #3360
/claim #3360

## Tests
- `node --test apps/api/src/tests/validationErrorHandler.test.js apps/api/src/tests/health.test.js`

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,7 +1,20 @@
+import { ZodError } from "zod";
+
 export function errorHandler(err, req, res, next) {
   console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
+  }
+
+  if (err instanceof ZodError) {
+    return res.status(400).json({
+      success: false,
+      message: "Validation error",
+      issues: err.issues.map((issue) => ({
+        path: issue.path.join("."),
+        message: issue.message
+      }))
+    });
   }
 
   return res.status(500).json({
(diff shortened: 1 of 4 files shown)


18166714330cl-maker:
Thanks for taking a look.

I treated this as a focused API error-boundary fix rather than changing validation schemas broadly.

The issue I’m addressing is that Zod validation failures from async routes should produce safe 400 responses instead of falling through as generic server errors. That makes malformed client input explicit without exposing internal error behavior.

What changed:
- Added async job route error handling so validation failures reach Express error middleware.
- Mapped Zod validation errors to 400 responses with safe issue details.
- Added a regression test for malformed job creation payloads.
- Left unrelated route behavior unchanged.

Verification:
- `node --test apps/api/src/tests/validationErrorHandler.test.js apps/api/src/tests/health.test.js`

The main review question is whether this error-shaping should stay focused on validation errors or be extended to a broader error taxonomy later. I kept this PR narrow to avoid changing unrelated error behavior.

PR #3430, opened 2026-06-01
fix: keep review creation timestamp server-owned

### LINKED ISSUE #3450: Review service should keep creation timestamps server-owned
The review service should own the creation timestamp for newly-created reviews. If `createdAt` is accepted from the request payload, clients can backdate or future-date reviews and make audit/order data unreliable.

Expected fix:
- ignore caller-supplied `createdAt` values during review creation
- assign `createdAt` in the service at creation time
- preserve the rest of the review payload
- add regression coverage proving the client timestamp is not stored

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- keep review `createdAt` assigned by the service instead of trusting request payloads
- preserve submitted review fields
- add stronger regression coverage for both backdated and future caller-supplied timestamps

## Test Plan
- `node --test apps/api/src/tests/reviewService.test.js --test-name-pattern 'createdAt timestamp|future caller' --test-reporter=spec`
- `git diff --check`

Closes #3450

/claim #743

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,11 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
-  const review = { id: `rev_${Date.now()}`, ...payload };
+  const review = {
+    id: `rev_${Date.now()}`,
+    ...payload,
+    createdAt: new Date().toISOString()
+  };
   reviews.push(review);
   return review;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:

- Focused: `node --test apps/api/src/tests/reviewService.test.js --test-name-pattern 'createdAt timestamp|future caller' --test-reporter=spec`
  - tests: 2
  - pass: 2
  - fail: 0
  - tests: 3
  - pass: 3
  - fail: 0
- `git diff --check` passed.

The coverage now proves `createdAt` is service-owned for both backdated and future caller-supplied timestamps, while review payload fields (`jobId`, `reviewerId`, `revieweeId`, `rating`, `comment`) are preserved.

PR #3436, opened 2026-06-01
fix: validate review rating range

### LINKED ISSUE #3439: Review service should reject ratings outside 1-5
The review service currently accepts any `rating` value supplied in the request payload and stores it as-is. That means ratings such as `0`, `6`, decimals, or non-numeric strings can be persisted even though user-facing review scores should stay within the expected 1-5 integer range.

Expected fix:
- reject ratings outside the 1-5 range
- reject non-integer ratings
- keep accepted ratings normalized before storing
- add regression coverage for both invalid and valid ratings

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- reject review ratings below 1 or above 5
- normalize accepted numeric ratings before storing them
- add regression coverage for invalid and valid ratings

## Test Plan

Closes #3439

/claim #743

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,15 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
-  const review = { id: `rev_${Date.now()}`, ...payload };
+  const rating = Number(payload.rating);
+
+  if (!Number.isInteger(rating) || rating < 1 || rating > 5) {
+    const error = new Error("Review rating must be an integer from 1 to 5");
+    error.statusCode = 400;
+    throw error;
+  }
+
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:
Validation/evidence update for #3439:

- Scope is limited to `apps/api/src/services/reviewService.js` and `apps/api/src/tests/reviewService.test.js`.
- The fix rejects non-integer ratings and ratings outside 1-5 before persisting a review.
- Accepted ratings are normalized to numbers before storage.
- Local validation passed:

```bash
node --test apps/api/src/tests/*.test.js --test-reporter=spec
```

Result:

```text
# tests 3
# pass 3
# fail 0
```

This PR intentionally stays limited to the creator-owned issue #3439 and parent bounty #743. I can add a short terminal demo video if maintainers/Algora require one for this API-only change.

PR #3441, opened 2026-06-01
fix: keep user ids server-owned

### LINKED ISSUE #3440: User service should keep user ids server-owned
The user service currently builds a new user with `{ id: ..., ...payload }`, so a caller can send an `id` field and overwrite the server-generated identifier. User ids should be assigned by the service, not trusted from request payloads.

Expected fix:
- ignore caller-supplied `id` values when creating users
- keep the rest of the payload intact
- add regression coverage proving the stored id is server-generated

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- prevent client payloads from overriding generated user ids
- keep non-id user fields unchanged
- add stronger regression coverage for both returned and stored generated id behavior

## Test Plan
- `node --test apps/api/src/tests/userService.test.js --test-name-pattern 'preserves allowed payload fields|stores the generated id' --test-reporter=spec`
- `git diff --check`

Closes #3440

/claim #743

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,7 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  const user = { ...payload, id: `usr_${Date.now()}` };
   users.push(user);
   return user;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:
Validation evidence after strengthening the regression coverage:

- Focused: `node --test apps/api/src/tests/userService.test.js --test-name-pattern 'preserves allowed payload fields|stores the generated id' --test-reporter=spec`
  - tests: 2
  - pass: 2
  - fail: 0
  - tests: 3
  - pass: 3
  - fail: 0
- `git diff --check` passed.

The new coverage now proves both sides of the fix: caller-supplied `id` is not returned/stored, while normal payload fields (`email`, `role`, `fullName`) are preserved.

PR #3443, opened 2026-06-01
fix: keep proposal ids server-owned

### LINKED ISSUE #3442: Proposal service should ignore caller-supplied ids
The proposal service currently creates proposals with `{ id: ..., ...payload }`, which lets a request payload replace the generated proposal id. Proposal ids should be service-owned to avoid spoofed or colliding records.

Expected fix:
- ignore caller-supplied `id` values
- preserve valid proposal payload fields
- add a regression test for server-owned proposal ids

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- prevent proposal creation payloads from overriding generated ids
- preserve submitted proposal fields
- add stronger regression coverage for returned/generated proposal id ownership

## Test Plan
- `node --test apps/api/src/tests/proposalService.test.js --test-name-pattern 'server-owned|spoofed caller id' --test-reporter=spec`
- `git diff --check`

Closes #3442

/claim #743

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { ...payload, id: `prp_${Date.now()}` };
   proposals.push(proposal);
   return proposal;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:

- Focused: `node --test apps/api/src/tests/proposalService.test.js --test-name-pattern 'server-owned|spoofed caller id' --test-reporter=spec`
  - tests: 2
  - pass: 2
  - fail: 0
  - tests: 3
  - pass: 3
  - fail: 0
- `git diff --check` passed.

The coverage now proves caller-supplied proposal ids are ignored while normal proposal fields (`jobId`, `freelancerId`, `coverLetter`, `bidAmount`) are preserved.

PR #3445, opened 2026-06-01
fix: keep new notifications unread

### LINKED ISSUE #3444: Notification creation should always start unread
The notification service initializes `read: false` before spreading the request payload, so a caller can submit `read: true` and create a notification that is already marked as read. New notifications should start unread regardless of client input.

Expected fix:
- keep generated notification ids server-owned
- force newly-created notifications to `read: false`
- add regression coverage for attempts to create already-read notifications

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- prevent clients from creating already-read notifications
- keep notification ids generated by the service
- add stronger regression coverage for server-owned `id`/`read` while preserving normal notification payload fields

## Test Plan
- `node --test apps/api/src/tests/notificationService.test.js --test-name-pattern 'starts unread|already-read' --test-reporter=spec`
- `git diff --check`

Closes #3444

/claim #743

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,7 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const notification = { ...payload, id: `ntf_${Date.now()}`, read: false };
   notifications.push(notification);
   return notification;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:

- Focused: `node --test apps/api/src/tests/notificationService.test.js --test-name-pattern 'starts unread|already-read' --test-reporter=spec`
  - tests: 2
  - pass: 2
  - fail: 0
  - tests: 3
  - pass: 3
  - fail: 0
- `git diff --check` passed.

The coverage now proves new notifications cannot be created as already read, notification ids stay service-generated, and normal payload fields (`userId`, `message`, `type`) are preserved.

PR #3447, opened 2026-06-01
fix: keep job creation fields server-owned

### LINKED ISSUE #3446: Job creation should keep id and initial status server-owned
The job service builds new jobs with `{ id: ..., status: "open", ...payload }`. Because the payload is spread last, callers can override both the generated job id and the initial job status. New jobs should always receive a server-generated id and start in the `open` state.

Expected fix:
- ignore caller-supplied `id` values
- ignore caller-supplied initial `status` values
- preserve the rest of the job payload
- add regression coverage for both server-owned fields

Parent bounty reference: #743

eliterdav09-creator:
## Summary
- prevent job creation payloads from overriding generated ids
- force new jobs to start with `open` status
- add stronger regression coverage for both server-owned fields while proving normal job fields are preserved

## Test Plan
- `node --test apps/api/src/tests/jobService.test.js --test-name-pattern 'server-owned|generated open job' --test-reporter=spec`
- `git diff --check`

Closes #3446

/claim #743

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,7 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const job = { ...payload, id: `job_${Date.now()}`, status: "open" };
   jobs.push(job);
   return job;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:
Validation evidence after strengthening the regression coverage:

- Focused: `node --test apps/api/src/tests/jobService.test.js --test-name-pattern 'server-owned|generated open job' --test-reporter=spec`
  - tests: 2
  - pass: 2
  - fail: 0
  - tests: 3
  - pass: 3
  - fail: 0
- `git diff --check` passed.

The added coverage now proves both controlled fields stay server-owned (`id`, initial `status`) while normal job payload fields (`title`, `description`, `budgetMin`, `budgetMax`) are still preserved.

PR #3469, opened 2026-06-01
fix: reject inverted job budget ranges

### LINKED ISSUE #3468: Reject inverted job budget ranges in job validation
## Bug

## Expected behavior

Parent bounty: #743

hermesbountyhunter:
## Summary
- Add ordered budget-range validation for job create and partial update schemas.
- Keep one-sided partial budget updates valid while rejecting inverted ranges when both fields are present.
- Add focused node:test coverage for valid create, invalid create/update, and partial update behavior.

Closes #3468
Refs #743

## Test plan
- `npm test`

## Automation disclosure
Implemented and tested by Hermes Agent on behalf of @hermesbountyhunter.

### DIFF
--- a/apps/api/src/tests/jobValidator.test.js
+++ b/apps/api/src/tests/jobValidator.test.js
@@ -1,0 +1,42 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
+
+const validJob = {
+  title: "Build landing page",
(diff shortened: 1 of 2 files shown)


hermesbountyhunter:
Closing this because, after re-checking the repository contribution instructions, I should not proceed with an AI-assisted bounty PR where the participation instruction includes a social/star action. I am withdrawing the issue/PR rather than continuing under that gate.

PR #3497, opened 2026-06-01
fix: scope proposal lists to authenticated participants

### LINKED ISSUE #3496: Proposal list should scope records to authenticated participants
Parent bounty: #743

## Problem

`GET /api/proposals` currently returns every stored proposal without requiring a bearer token or scoping records to the authenticated user.

Proposal records can contain bid amounts, cover letters, job IDs, freelancer IDs, and client IDs. Returning the entire list lets anonymous users or unrelated authenticated users read proposal data that should only be visible to a proposal participant.

## Reproduction

1. Create one proposal for `freelancerId: usr_a` and another for `freelancerId: usr_b`.
2. Call `GET /api/proposals` without an `Authorization` header.
3. The endpoint returns all proposals.
4. Even if a caller is authenticated, the current service has no user filter and would still return unrelated proposals.

## Expected behavior

- Missing or invalid tokens should receive the shared 401 API response.
- `GET /api/proposals` should only return proposals where the authenticated user is a participant, using `freelancerId` or `clientId`.
- A token without a subject should not fall back to returning every proposal.

## Proposed fix

- Add `authMiddleware` to the proposal list route.
- Pass `req.user.sub` into the proposal list service.
- Filter proposals by `freelancerId === userId || clientId === userId`.
- Add API regression tests for anonymous rejection and authenticated user scoping.

## Non-overlap

This is scoped only to authenticated, user-scoped reads for proposal lists. It does not cover proposal creation authentication, proposal payload validation, mutable store arrays, broad all-route authorization, review lists, messages, notifications, users, admin authorization, or payments.

NikYak228:
/claim #743
Closes #3496
References #743

## Summary

- require bearer-token authentication before listing proposals
- scope proposal list results to the authenticated user as either `freelancerId` or `clientId`
- add route-level regression coverage for anonymous rejection and participant-only results

## Validation

- `node --test apps/api/src/tests/proposalRoutes.test.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/proposalRoutes.test.js`
- `npm test -w apps/api`
- `git diff --check`

## Demo

https://github.com/NikYak228/bug-bounty/releases/download/demo-3496/securebanana-3497-demo.mp4

## Scope

This PR only changes proposal list reads. It does not change proposal creation, proposal payload validation, or other resource list endpoints.

### DIFF
--- a/apps/api/src/controllers/proposalController.js
+++ b/apps/api/src/controllers/proposalController.js
@@ -2,7 +2,7 @@ import { ok } from "../utils/response.js";
 import { createProposal, listProposals } from "../services/proposalService.js";
 
 export async function getProposals(req, res) {
-  return ok(res, await listProposals());
+  return ok(res, await listProposals(req.user?.sub));
 }
 
 export async function postProposal(req, res) {
--- a/apps/api/src/controllers/proposalController.js
+++ b/apps/api/src/controllers/proposalController.js
@@ -1,8 +1,17 @@
-import { ok } from "../utils/response.js";
-import { createProposal, listProposals } from "../services/proposalService.js";
+import { fail, ok } from "../utils/response.js";
+import { createProposal, listProposalsForUser } from "../services/proposalService.js";
 
 export async function getProposals(req, res) {
-  return ok(res, await listProposals(req.user?.sub));
+  if (!req.user) {
+    return fail(res, "Unauthorized", 401);
+  }
+
+  const userId = req.user.sub;
+  if (!userId) {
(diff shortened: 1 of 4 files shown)


NikYak228:
Hi maintainers, gentle follow-up on this Algora bounty claim.

This PR is still open and mergeable from GitHub's current status, and I don't see any maintainer feedback or requested changes yet. Could you please confirm whether this is still pending review, or if anything else is needed from my side to move the claim forward?

Thanks.

PR #3501, opened 2026-06-01
fix: return defensive copies from list services

### LINKED ISSUE #3498: List service responses should be defensive array copies
Parent bounty: #743

## Bug

Several API service list functions return their module-level in-memory arrays directly. Any in-process caller that receives one of these arrays can mutate it with `push`, `pop`, `splice`, or `length = 0`, which silently corrupts the backing store used by later reads.

Affected examples:
- `listJobs()` returns `jobs`

## Expected behavior

List functions should expose snapshot arrays while preserving the current record objects and API response shape.

## Scope

- Add focused regression coverage proving a caller cannot mutate the backing store by mutating a list response.

NikYak228:
/claim #743

Closes #3498
References #743

## Summary

- return shallow copies from the API list services so callers cannot mutate the backing store through a returned array
- add regression coverage proving a list response can be mutated locally without affecting later reads
- preserve the existing record objects and response shape

## Validation

- `npm test -w apps/api`
- `git diff --check`

## Demo

https://github.com/NikYak228/bug-bounty/releases/download/demo-3498/securebanana-3498-demo.mp4

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,7 +1,7 @@
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return [...jobs];
 }
 
 export async function createJob(payload) {
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,7 +1,7 @@
 const messages = [];
 
 export async function listMessages() {
-  return messages;
+  return [...messages];
 }
 
 export async function sendMessage(payload) {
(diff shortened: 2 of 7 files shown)


NikYak228:
Hi maintainers, gentle follow-up on this Algora bounty claim.

This PR is still open and mergeable from GitHub's current status, and I don't see any maintainer feedback or requested changes yet. Could you please confirm whether this is still pending review, or if anything else is needed from my side to move the claim forward?

Thanks.

PR #3523, opened 2026-06-01
fix: align registration token subject with returned user id

ank1015:
Withdrawn.


ank1015:
Withdrawing after seeing the repository README notice that this is a synthetic research sandbox and human submissions are not wanted or eligible for payout.

PR #3534, opened 2026-06-01
fix: require auth for message routes

### LINKED ISSUE #3095: Message API should require authentication
Parent bounty: #743

The message API currently allows unauthenticated callers to list and create messages through /api/messages. Messages are account-scoped/private data, so both reading and sending messages should require a valid access token.

Proposed fix:
- Apply authMiddleware to messageRoutes.
- Add regression tests for unauthenticated GET and POST /api/messages returning 401.

NikYak228:
Parent bounty: #743

This protects `/api/messages` with the existing `authMiddleware` so anonymous callers cannot list or send messages. It adds focused regression coverage for unauthenticated GET and POST requests returning 401.

Fixes #3095
/claim #743

Validation:
- npm test -w apps/api
- git diff --check

### DIFF
--- a/apps/api/src/routes/messageRoutes.js
+++ b/apps/api/src/routes/messageRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
+import { authMiddleware } from "../middleware/auth.js";
 import { getMessages, postMessage } from "../controllers/messageController.js";
 
 export const messageRoutes = Router();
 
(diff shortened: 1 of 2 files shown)


NikYak228:
Closing this as a duplicate after rechecking exact issue-number searches: PR #3098 already covers #3095 with the same message-route auth scope. Avoiding duplicate bounty noise.

PR #3536, opened 2026-06-01
fix: require auth for notification routes

### LINKED ISSUE #3099: Notification API should require authentication
Parent bounty: #743

The notification API currently allows unauthenticated callers to list and create notifications through /api/notifications. Notifications are user-specific account data, so reads and writes should require a valid access token.

Proposed fix:
- Apply authMiddleware to notificationRoutes.
- Add regression tests for unauthenticated GET and POST /api/notifications returning 401.

NikYak228:
Parent bounty: #743

This protects `/api/notifications` with the existing `authMiddleware` so anonymous callers cannot list or create notifications. It adds focused regression coverage for unauthenticated GET and POST requests returning 401.

Fixes #3099
/claim #743

Validation:
- npm test -w apps/api
- git diff --check

### DIFF
--- a/apps/api/src/routes/notificationRoutes.js
+++ b/apps/api/src/routes/notificationRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
+import { authMiddleware } from "../middleware/auth.js";
 import { getNotifications, postNotification } from "../controllers/notificationController.js";
 
 export const notificationRoutes = Router();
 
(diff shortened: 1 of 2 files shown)


NikYak228:
Closing this as a duplicate after rechecking exact issue-number searches: PR #3101 already covers #3099 with the same notification-route auth scope. Avoiding duplicate bounty noise.

PR #3564, opened 2026-06-02
fix: keep user ids server-owned

### LINKED ISSUE #3440: User service should keep user ids server-owned

eliterdav09-creator:
## Summary
- Keeps `createUser()` generated IDs server-owned by applying the service-generated id after the request payload.
- Adds focused service coverage proving a caller-supplied `id` cannot replace the generated user id.

## Test plan
- `cd apps/api && node --test src/tests/userService.test.js --test-name-pattern 'createUser keeps generated ids server-owned'`
- `cd apps/api && node --test src/tests/*.test.js`

Closes #3440

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,7 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  const user = { ...payload, id: `usr_${Date.now()}` };
   users.push(user);
   return user;
 }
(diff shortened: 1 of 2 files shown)


eliterdav09-creator:
Closing this duplicate branch in favor of the earlier focused submission #3441.

#3441 now has the strengthened regression coverage and validation evidence for the same issue scope (#3440), so keeping this PR open would only split review attention.

PR #3566, opened 2026-06-02
Fix registration token subject mismatch

houyiai:
Fixes #2768

## Summary
- Generate the registration user id once and reuse it for both the returned `id` and JWT `sub` claim.

## Testing
- `node --test "src/tests/*.test.js"` from `apps/api` passes.
- Note: root `npm test` currently fails because `apps/api/package.json` runs `node --test src/tests`, which Node treats as a file path on this environment instead of discovering tests in the directory.

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (changes requested):
Requesting changes: this PR only updates authService.js and still doesn't include the regression coverage required by the bounty. The issue explicitly asks for a focused test proving the decoded token subject matches the returned id. Without that test, I don't consider this acceptance-ready.

PR #3567, opened 2026-06-02
fix: validate search query input

### LINKED ISSUE #1579: Search endpoint accepts repeated and oversized query values

## Problem
`GET /api/search` currently passes `req.query.q ?? ""` directly into `globalSearch()`. A single `q` value is string-shaped, but repeated query parameters such as `?q=one&q=two` can produce an array. Very long query values are also accepted without any API boundary.

## Reproduction
Send this request:

```text
GET /api/search?q=one&q=two
```

Expected: the API should reject non-single-string query input with HTTP 400.
Actual: the search service can receive and echo a non-string `query` shape.

Also send a query string longer than 200 characters. It is currently accepted instead of returning HTTP 400.

## Impact
The response shape becomes inconsistent and downstream search code cannot rely on `query` being a string. This also leaves the endpoint without a simple size boundary for user-controlled search input.

## Proposed fix
- Add a dedicated search query validator.
- Allow `q` to be omitted or a single string only.
- Reject repeated/non-string query values and oversized queries.
- Cover valid, repeated, and oversized query cases with API tests.

## Non-overlap
This is search-route validation only. It does not overlap with auth, payment, upload, message payload, review rating, proposal validation, notification validation, user validation, admin metrics, JWT secret, or job budget validation issues.

Parent bounty: #743

vondutchi:
Closes #1579

/claim #743

## Summary
- Reject repeated `q` query parameters before they reach the search service.
- Reject search query values longer than 200 characters.
- Preserve omitted and valid single-string search queries.
- Add focused API regression coverage for omitted, valid, repeated, and oversized query cases.

## Validation
- `node --test apps/api/src/tests/search.test.js` -> 4 passed
- `node --test apps/api/src/tests/health.test.js` -> 1 passed
- `node --check apps/api/src/controllers/searchController.js`
- `node --check apps/api/src/validators/search.js`
- `node --check apps/api/src/tests/search.test.js`
- `git diff --check`

## Note
`npm test -w apps/api` still fails on the existing upstream script path `node --test src/tests`; direct Node test-file execution passes for the affected tests.

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,13 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { parseSearchQuery } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const parsed = parseSearchQuery(req.query);
+
+  if (!parsed.success) {
+    return fail(res, parsed.message, 400);
+  }
+
+  return ok(res, await globalSearch(parsed.query));
 }
--- a/apps/api/src/tests/search.test.js
+++ b/apps/api/src/tests/search.test.js
@@ -1,0 +1,72 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
+async function withServer(run) {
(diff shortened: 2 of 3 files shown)


vondutchi:
Closing this as a duplicate of my earlier open PR #1580 for issue #1579. #1580 already covers the repeated/non-string and oversized search query validation scope with /claim #743 and green checks.

PR #3791, opened 2026-06-02
fix: cap auth password length at 128 (issue #3740)

### LINKED ISSUE #3740: Auth validators should cap password length (scoped #743 reissue)
Parent bounty: #743
Source issue: #3406

## Bug
The auth request validators currently accept passwords with a minimum length but no upper bound:

```js
password: z.string().min(8)
```

That lets very large password strings pass validation and reach registration or login processing, creating avoidable CPU/memory pressure.

## Expected behavior
- Registration passwords should still require at least 8 characters.
- Login passwords should still require at least 8 characters.
- Both registration and login passwords should reject inputs longer than 128 characters.
- Regression tests should cover accepted 128-character passwords and rejected 129-character passwords.

/attempt #743

zeroknowledge0x:
Closes #3740

Adds `.max(128)` to password validation in both `registerSchema` and `loginSchema` to prevent resource exhaustion from overlong passwords.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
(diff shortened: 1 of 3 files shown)


zeroknowledge0x:

Issue #3750, opened 2026-06-02
Review creation should reject self-review payloads (scoped #743 reissue)


stevesagronegocios673-ux:
Parent bounty: #743
Source issue: #3279

## Bug
`createReview()` currently accepts payloads where `reviewerId` and `revieweeId` are the same account. In a freelance marketplace, self-reviews let an account inflate its own trust/reputation history and pollute review data that should come from another participant.

## Expected behavior
- Reject review payloads where `reviewerId` equals `revieweeId`.
- Preserve valid reviews where the reviewer and reviewee are different users.
- Add focused service regression coverage for both the rejection path and the valid different-user path.

This is a scoped creator-owned reissue for #743.

/attempt #743

mvmax-dev:
/attempt #3750

Plan:
1. Enforce validation rule in `createReview()` in `reviewService.js` ensuring `reviewerId` does not match `revieweeId`.
2. Reject self-review payloads with explicit validation error.
3. Write robust unit tests covering valid and invalid (self-review) constraints.

PR #3793, opened 2026-06-02
fix: enforce integer rating 1-5 in review service (issue #3742)

### LINKED ISSUE #3742: Review service should enforce integer ratings in range (scoped #743 reissue)
Parent bounty: #743
Source issue: #3439

## Bug
`createReview()` currently accepts any `rating` value supplied in the request payload and stores it as-is. That means values such as `0`, `6`, decimals, or non-numeric strings can be persisted even though review scores should stay within the expected 1-5 integer range.

## Expected behavior
- Accept integer ratings from 1 through 5.
- Reject ratings outside the 1-5 range.
- Reject non-integer and non-number ratings.
- Add focused service regression coverage for valid and invalid ratings.

/attempt #743

zeroknowledge0x:
Closes #3742

Rejects non-integer ratings and values outside 1-5 in `createReview()`. Keeps existing self-review rejection from #3750 fix. Adds service-level regression coverage for valid and invalid ratings.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
   return ok(res, {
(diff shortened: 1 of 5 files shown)


zeroknowledge0x:

PR #3798, opened 2026-06-02
fix: keep notification creation id and read server-owned (issue #3721)

### LINKED ISSUE #3721: Notification creation should keep id and read state server-owned (scoped #743 reissue)
Parent bounty: #743
Source issue: #3612 / #2762

## Bug

`createNotification()` in `apps/api/src/services/notificationService.js` creates records with server-owned fields first, then spreads the untrusted payload afterward:

```js
```

Because the payload is spread last, callers can override both the generated notification id and the initial unread state.

## Expected behavior

- Notification creation should always use a server-generated `ntf_...` id.
- Newly-created notifications should always start with `read: false`.

## Scope

Keep the fix focused on the notification service behavior and add a regression test proving caller-supplied `id` / `read` cannot override server-owned creation fields.

zeroknowledge0x:
Closes #3721

Ignores caller-supplied `id` and `read` in `createNotification()` so generated ids and unread state remain server-owned. Adds regression coverage for both notification service paths.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
   return ok(res, {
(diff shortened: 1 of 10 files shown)


zeroknowledge0x:

PR #3817, opened 2026-06-02
feat: require JWT_SECRET in production (issue #3732)

### LINKED ISSUE #3732: Production config should require JWT_SECRET (reissue via #743)
Parent Algora bounty: #743

## Bug
The JWT secret falls back to `development-secret` when `JWT_SECRET` is not set. That fallback is useful for local development, but in production it lets anyone who knows the default forge valid tokens if the environment is misconfigured.

## Affected file
`apps/api/src/config/env.js`

## Expected behavior
- Keep the development fallback for local development.
- Require `JWT_SECRET` when `NODE_ENV=production`.
- Add focused config regression tests for development fallback, production rejection, and explicit production secret.

/attempt #743

zeroknowledge0x:
Closes #3732

Requires `JWT_SECRET` when `NODE_ENV=production` while preserving the local development fallback. JWT utilities now import the resolved singleton, so fresh processes cannot boot with the default secret in production.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,15 @@
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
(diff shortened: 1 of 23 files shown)


zeroknowledge0x:

PR #3818, opened 2026-06-02
fix: keep registration token subject aligned with returned user id (issue #3645)

### LINKED ISSUE #3645: Registration token subject should match returned user id (reissue #3588)
Parent bounty: #743

Reissue of #3588. Reference: https://github.com/SecureBananaLabs/bug-bounty/issues/3588

## Bug
`apps/api/src/services/authService.js` generates a timestamp-based user id twice during registration: once for the returned `id` field and once for the JWT `sub` claim.

Current behavior can return one user id while signing the token for a different `sub` if the two `Date.now()` calls cross a millisecond boundary.

## Expected
- Generate the user id once.
- Return that same id in the registration response.
- Sign the access token with that same id as the `sub` claim.

Related reissue: #2768 / #2674

/attempt #743

zeroknowledge0x:
Closes #3645

Generates the user id once in `registerUser`, returns it in the response, and signs the access token (`sub`) from that same id. Adds regression coverage to assert the decoded token subject always matches the returned id.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,15 @@
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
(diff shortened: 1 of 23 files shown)


zeroknowledge0x:

PR #3819, opened 2026-06-02
fix: preserve fullName in registration and validate refresh tokens (issues #3177, #3170)

### LINKED ISSUE #3170: fix: refresh endpoint should validate refresh token
## Bug: Refresh Token Not Validated

**Description:** `POST /api/auth/refresh` currently ignores the request body and calls `refreshToken()` without validating any supplied refresh token. The service then returns a new access token for the hard-coded `usr_existing` user, so an unauthenticated caller can mint a fresh client token without presenting any credential.

**Expected behavior:** The refresh endpoint should require a token in the request body, reject missing or invalid tokens, and only issue a replacement access token for the subject and role contained in a valid token.

**Files:**
- `apps/api/src/controllers/authController.js`

Refs #743.

### LINKED ISSUE #3177: Registration should require fullName before creating users
Parent bounty: #743

## Bug
The Prisma `User` model requires `fullName`, but the auth registration contract only accepts `email`, `password`, and `role`. `registerUser()` also omits `fullName` from the returned user payload. When registration is wired to the Prisma user model, requests accepted by the API can be missing a required user field.

## Expected
- `registerUser()` carries the validated `fullName` into the returned user payload.
- Invalid registration payloads return a stable client error instead of being accepted.
- Tests cover rejection of missing `fullName` and preservation of a valid full name.

## Scope
Keep the fix limited to registration validation/service behavior and focused API tests.

zeroknowledge0x:
Closes #3177
Closes #3170

- `registerUser()` now preserves the validated `fullName` in the returned user record and keeps it aligned with the registration schema and Prisma model.
- `refreshToken()` now validates the submitted refresh token, derives the replacement access token from the verified subject and role, and rejects empty or malformed tokens.
- Adds regression coverage for registration `fullName` preservation and refresh-token happy, missing, and invalid paths.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,15 @@
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
   stripeSecretKey: process.env.STRIPE_SECRET_KEY ?? "",
   databaseUrl: process.env.DATABASE_URL ?? ""
 };
+
+function resolveJwtSecret() {
+  if (process.env.NODE_ENV === "production" && !process.env.JWT_SECRET) {
+    throw new Error("JWT_SECRET is required in production");
+  }
+  return process.env.JWT_SECRET ?? "development-secret";
+}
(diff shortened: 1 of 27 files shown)


zeroknowledge0x:

PR #3865, opened 2026-06-03

### LINKED ISSUE #3864: Job validation should reject inverted budget ranges
Parent bounty: #743

Part of the low-hanging-fruit bounty flow in #743.

## Bug
The API job validator accepts inverted budget ranges where `budgetMin` is greater than `budgetMax`. That allows nonsensical jobs like:

```json
{
  "budgetMin": 500,
  "budgetMax": 100
}
```

## Expected behavior
Create/update job validation should reject payloads where both budget fields are present and `budgetMax < budgetMin`, while still allowing partial updates that only include one budget field.

## Fix plan
- Add a shared Zod refinement for job budget ordering.
- Apply it to both create and update job schemas.
- Add node:test coverage for valid, invalid, and partial update cases.

BryanGM12:
Fixes #3864.

## Summary
- rejects inverted job budget ranges with a shared Zod refinement
- applies the rule to create and update schemas
- keeps single-field budget updates valid
- fixes the API test script on Windows/Node by using an explicit test-file glob

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


BryanGM12:
Tested with 
pm test -w apps/api (5 passing). If this qualifies for the bounty/payment flow, BTC on-chain address: c1qfjvvg2tgcq36exzusafk92vaz9529a2dkhrshk.

PR #3874, opened 2026-06-03
Keep registration token subject in sync with user id

### LINKED ISSUE #3873: Registration token subject should match returned user id
Parent bounty: #743

## Bug
`registerUser` creates the returned user id and JWT subject with two separate `Date.now()` calls. If the second call returns a different millisecond, the API can return `id: usr_1000` while signing a token with `sub: usr_1001`. That makes the freshly registered session refer to a different principal than the response body.

## Expected
- `registerUser` should create one user id value.
- The returned `id` and signed JWT `sub` should always match.
- A regression test should cover the case where time advances between calls.

## Scope
Keep the fix limited to registration token generation and focused unit coverage.

BryanGM12:
Fixes #3873.

## Summary
- computes the registration user id once
- signs the access token with the same id returned in the response
- adds regression coverage that forces `Date.now()` to advance between calls

## Verification
- `node --test "src/tests/*.test.js"` from `apps/api`

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,12 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = `usr_${Date.now()}`;
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
     email: payload.email,
     role: payload.role,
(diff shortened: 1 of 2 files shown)


BryanGM12:
Verified with `node --test "src/tests/*.test.js"` from `apps/api`. If this qualifies for the bounty/payment flow, BTC on-chain address: `bc1qfjvvg2tgcq36exzusafk92vaz9529a2dkhrshk`.

PR #4016, opened 2026-06-03
fix: make message id and sentAt server-owned

### LINKED ISSUE #4005: Message service should keep ids and stored records server-owned
Currently, the message sending endpoint allows the client to overwrite the generated message ID and sentAt fields by passing customized values in the payload body. The message model should ensure these values remain strictly server-owned.

thien-an-ops:
The sendMessage endpoint allowed the client to overwrite generated message IDs and sentAt fields by passing customized values in the payload body. This fix reorders the spread so server-generated values always take precedence.

Closes #4005

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,7 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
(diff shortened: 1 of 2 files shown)


thien-an-ops:
Closing — this repo is a research honeypot, not a real bounty platform.

PR #4017, opened 2026-06-03
fix: login should reject unknown email and wrong password

### LINKED ISSUE #3999: Login should reject unknown or wrong credentials
Currently, the login endpoint POST /api/auth/login accepts any email and returns a token without validating whether the email exists in the database or if the password matches the registered hash.

thien-an-ops:
loginUser currently returns a valid token for any email without checking if the user exists or if the password matches. This fix adds proper credential validation with hashed password storage using Node.js crypto.

Closes #3999


thien-an-ops:
Closing — this repo is a research honeypot, not a real bounty platform.

Issue #3671, opened 2026-06-02
Proposal creation should preserve server-generated id


eliterdav09-creator:
Parent Algora bounty: #743

## Bug

`createProposal()` currently constructs new proposal records as:

```js
const proposal = { id: `prp_${Date.now()}`, ...payload };
```

Because the untrusted request payload is spread after the server-generated fields, a client-provided `id` overwrites the proposal id generated by the service. This allows callers to create proposals with attacker-controlled identifiers instead of the expected `prp_<timestamp>` server-owned id.

## Expected behavior

Proposal creation should ignore any client-supplied `id` and always preserve the server-generated proposal id, while keeping legitimate proposal fields intact.

## Scope

Keep the fix limited to proposal creation and add a regression test proving a client-controlled `id` cannot override the generated proposal id.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue Low Handing Fruit Automation #743 for more information.

/attempt #743

PR #4250, opened 2026-06-04
Preserve proposal server id

### LINKED ISSUE #4249: Proposal creation should preserve server-generated id
Parent Algora bounty: #743

## Problem

`createProposal()` currently builds new records as:

```js
const proposal = { id: `prp_${Date.now()}`, ...payload };
```

Because the untrusted payload is spread after the server-generated id, a caller can supply `id` in the request body and overwrite the `prp_...` identifier.

## Impact

New proposals can be created with client-controlled identifiers instead of the service-owned id format, which can confuse downstream lookups and records that expect proposal ids to be generated by the API.

## Expected behavior

Proposal creation should preserve the generated `prp_...` id while keeping legitimate payload fields intact.

## Scope

Keep the fix limited to proposal creation and add a regression test proving a client-supplied `id` cannot override the generated proposal id.

/attempt #743

bozicovichsantiago20-oss:
## Summary
- Preserve the generated `prp_...` id when creating proposals.
- Keep legitimate proposal payload fields intact.
- Add a regression test for client-supplied proposal ids.

## Validation
- node --test apps/api/src/tests/proposal-service.test.js
- git diff --check

Closes #4249
Refs #743
/claim #4249

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { ...payload, id: `prp_${Date.now()}` };
   proposals.push(proposal);
   return proposal;
 }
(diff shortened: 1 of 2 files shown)


bozicovichsantiago20-oss:
Closing this as a duplicate after finding earlier matching issues #4092 and #3671 for the same proposal id override. Keeping the bounty board clean.

bozicovichsantiago20-oss:
Reopening: #743's issue template asks contributors to create creator-owned reissues for the same contents. This PR targets my own issue #4249 and remains a focused fix.

PR #4454, opened 2026-06-04
Require verified auth for token refresh

### LINKED ISSUE #4452: Auth refresh should require a verified requester before minting tokens
## Bug

`POST /api/auth/refresh` currently mints a new access token without verifying any existing bearer token or requester identity. Any caller can hit the refresh route and receive a valid token for the hard-coded `usr_existing` client identity.

## Expected behaviour

- The refresh route should require a valid bearer token.
- The refreshed token should preserve the verified requester's `sub` and `role` claims.
- Requests without a bearer token, or with an invalid token, should return 401.

## Scope

This issue is limited to the auth refresh route and refresh service. It should not change registration, login, user routes, or job validation.

Parent bounty: #743

minorstep:
Closes #4452

/claim #743

## Summary

- Require `authMiddleware` on `POST /api/auth/refresh`.
- Refresh tokens from the verified requester claims instead of hard-coded `usr_existing` data.
- Add route-level regression tests for missing auth, invalid bearer tokens, and claim preservation.

## Validation

- `node --test apps/api/src/tests/auth-refresh.test.js` -> 3 passed
- `node --test apps/api/src/tests/*.js` -> 4 passed
- `node --check apps/api/src/routes/authRoutes.js`
- `node --check apps/api/src/tests/auth-refresh.test.js`
- `git diff --check`

Note: `npm test -w apps/api` could not run in this shell because `npm` is not installed on PATH, so I ran the API test files directly with Node's test runner.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const result = await refreshToken(req.user);
   return ok(res, result);
 }
--- a/apps/api/src/routes/authRoutes.js
+++ b/apps/api/src/routes/authRoutes.js
@@ -1,4 +1,5 @@
 import { Router } from "express";
+import { authMiddleware } from "../middleware/auth.js";
 import { login, oauthCallback, refresh, register } from "../controllers/authController.js";
 
 export const authRoutes = Router();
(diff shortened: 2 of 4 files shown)


minorstep:
Payment/decision follow-up for @minorstep.

This PR is still open/non-draft and the visible check is green. The public Algora claim page for this work currently shows $700, Status Pending, and Total paid $0:
https://algora.io/claims/5JATiyMgtSZQGLHg

Could a SecureBanana/Algora maintainer please confirm the decision path for this PR: approved/rewardable, needs changes, duplicate/not eligible, or blocked pending merge/sponsor action?

No private payout details are posted here. I am asking only for the public reward/eligibility status so this does not remain indefinitely pending.

PR #4472, opened 2026-06-04
Normalize payment currency codes

### LINKED ISSUE #4471: Payment currency should be normalized before intent creation
Parent Algora bounty: #743

## Problem

`createPaymentIntent` currently returns `payload.currency` exactly as submitted. A request with whitespace or mixed-case currency values such as `" USD "` produces a response with the same unnormalized value. Payment providers generally expect canonical lowercase currency codes, so callers can receive inconsistent payment intent data for the same currency.

## Impact

Clients may store or forward non-canonical currency codes, which can cause provider-side failures or make downstream payment records harder to compare.

## Expected behavior

Payment intent creation should normalize supplied string currency codes by trimming surrounding whitespace and lowercasing the value. Requests without a currency should continue to default to `"usd"`.

## Scope

Keep this limited to currency normalization in the payment intent service and add focused route-level regression tests for default currency and whitespace/mixed-case currency input.

/attempt #743

sjj79475067-max:
Fixes #4471

/claim #743

## Summary

- Normalize submitted payment currency codes by trimming whitespace and lowercasing them before returning the payment intent response.
- Keep the existing missing-currency behavior defaulting to `usd`.
- Add focused route-level regression tests for omitted currency and mixed-case/whitespace currency input.

## Validation

## Demo evidence

The payment route regression test exercises both default currency behavior and the `" USD "` normalization path end to end through the HTTP API.

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -1,9 +1,13 @@
+function normalizeCurrency(currency) {
+  return typeof currency === "string" ? currency.trim().toLowerCase() : "usd";
+}
+
 export async function createPaymentIntent(payload) {
   // TODO: integrate Stripe SDK and return client secret.
   return {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
-    currency: payload.currency ?? "usd",
+    currency: normalizeCurrency(payload.currency),
     provider: "stripe"
   };
 }
--- a/apps/api/src/tests/payment.test.js
+++ b/apps/api/src/tests/payment.test.js
@@ -1,0 +1,53 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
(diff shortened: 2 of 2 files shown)


sjj79475067-max:
Reviewer verification notes:

This PR is scoped to creator-owned issue #4471 and only normalizes payment currency codes in the payment intent response path.

Behavior covered by the new route tests:
- `POST /api/payments` without `currency` still returns `currency: "usd"`.
- `POST /api/payments` with `currency: " USD "` returns canonical `currency: "usd"`.

Local validation run:

```bash
node --test apps/api/src/tests/payment.test.js
node --test apps/api/src/tests/health.test.js
node --check apps/api/src/services/paymentService.js
node --check apps/api/src/tests/payment.test.js
```

Result: payment tests passed 2/2, health test passed 1/1, syntax checks passed.

PR #4553, opened 2026-06-05
Require Stripe config for production payments

### LINKED ISSUE #4551: Production payment creation should fail when Stripe is unconfigured (Floofy6 reissue via #743)
Parent bounty: #743

## Bug
`createPaymentIntent()` returns a successful `provider: "stripe"` response even when the API is running in production without `STRIPE_SECRET_KEY` configured. That can make production billing flows appear to succeed while no real payment provider is available.

## Expected
- In production, payment creation should fail fast when `STRIPE_SECRET_KEY` is missing.
- Development and test behavior can continue using the lightweight placeholder path.
- A focused regression test should prove production requests without Stripe configuration return a clear service-unavailable response instead of a fake successful payment.

## Scope
Keep the fix limited to payment configuration checks and focused API tests.

Floofy6:
Fixes #4551

/claim #743

## What changed
- Reject production payment creation with a structured 503 JSON response when `STRIPE_SECRET_KEY` is not configured
- Keep development and test placeholder payment behavior unchanged
- Add a focused route-level regression test for the missing production Stripe configuration case

## Validation
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/payment.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,11 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+import { env } from "../config/env.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
+  if (env.nodeEnv === "production" && !env.stripeSecretKey.trim()) {
+    return fail(res, "Payment provider is not configured", 503);
+  }
+
   return ok(res, await createPaymentIntent(req.body), 201);
 }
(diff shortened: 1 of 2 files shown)


Floofy6:
Quick review note for linked issue #4551 under bounty #743: I refreshed this branch today and re-ran the focused validation.

Validation:
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/payment.test.js` -> 2 passed
- `git diff --check origin/main..HEAD`

Local merge-tree against current `origin/main` was clean at the time of recheck. Happy to adjust the response shape/status if you prefer a different production-misconfiguration behavior.

PR #4555, opened 2026-06-05
Require registered credentials for login

### LINKED ISSUE #4554: Login should reject unknown or mismatched credentials (Floofy6 reissue via #743)
Parent bounty: #743

## Bug
The demo auth service currently returns an access token from `loginUser()` for any payload that passes shape validation. It does not verify that the email belongs to a registered user or that the supplied password matches the registered credential.

## Reproduction
```sh
curl -X POST /api/auth/login \
  -H "content-type: application/json" \
  -d "{"email":"nobody@example.com","password":"validpass123"}"
```

## Expected behavior
Login should only issue a token after the user has registered and the password matches. Unknown users or wrong passwords should receive a clear 401 response.

## Scope
Keep the fix limited to auth registration/login state and focused auth route tests. Preserve the current successful registration response shape and avoid exposing submitted passwords in auth responses.

Floofy6:
Fixes #4554

/claim #743

## What changed
- Store demo registration credentials in the auth service so login can verify known users
- Return a structured 401 response for unknown emails or mismatched passwords
- Preserve password-free registration and login response payloads
- Add focused auth route regression coverage for unknown-user, wrong-password, and successful registered-user login paths

## Validation
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/auth.test.js`
- `node --check apps/api/src/tests/auth.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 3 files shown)


Floofy6:
Quick review note for linked issue #4554 under bounty #743: I refreshed this branch today and re-ran the focused validation.

Validation:
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/auth.test.js` -> 4 passed
- `node --check apps/api/src/tests/auth.test.js`
- `git diff --check origin/main..HEAD`

Local merge-tree against current `origin/main` was clean at the time of recheck. Happy to adjust if you want the demo credential path handled differently.

PR #4558, opened 2026-06-05
fix: require file for upload endpoint

### LINKED ISSUE #4557: Upload endpoint should reject missing file requests
Parent bounty: #743

## Bug
`POST /api/uploads` currently returns `201` with `status: "no-file"` when the request does not include a multipart file. That makes a failed upload look successful to API clients and can let callers advance UI or workflow state without any file being stored.

## Expected behavior
- Missing-file upload requests should return a `400` JSON error.
- Valid multipart file uploads should continue returning `201` with the uploaded filename and `uploaded` status.
- The API test script should run the test files reliably so this regression is covered by `npm test -w apps/api`.

catcherintheroad-hub:
/claim #743

Closes #4557

## Summary
- Return a structured `400` error when `POST /api/uploads` is called without a multipart file.
- Keep successful multipart uploads returning `201` with the uploaded filename and `uploaded` status.
- Add focused upload route regression tests for missing-file and successful-file paths.
- Update the API test script to run `src/tests/*.test.js`, so `npm test -w apps/api` executes the actual test files.

## Demo / validation evidence

Before this change, a missing-file upload could return `201` with `status: "no-file"`. After this change the focused regression proves the endpoint now returns a clear client error while valid uploads still succeed.

```bash
npm test -w apps/api
```

Result:

```text
✔ POST /api/uploads rejects requests without a file
✔ POST /api/uploads accepts a multipart file upload
ℹ tests 3
ℹ pass 3
ℹ fail 0
```

Additional checks:

```bash
node --check src/controllers/uploadController.js
node --check src/tests/upload.test.js
git diff --check
```


catcherintheroad-hub:
Added a small demo evidence artifact for the upload validation behavior:

https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-upload-require-file/demos/upload-require-file-demo.svg

It summarizes the regression covered by the tests: missing multipart file now returns 400, while valid uploads still return 201.

catcherintheroad-hub:
Added a short MP4 demo video to match the #743 bounty guidance:

https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-upload-require-file/demos/upload-require-file-demo.mp4

It shows the same validated behavior as the tests: missing upload file now returns 400, while a valid multipart upload still returns 201. I also shortened the SVG demo text so the preview does not crop the right-side evidence panel.

PR #4565, opened 2026-06-05
Normalize search query input

### LINKED ISSUE #4564: Search endpoint should normalize query parameter to a string
Parent Algora bounty: #743

## Bug
`GET /api/search` passes `req.query.q` directly into `globalSearch()`. Express query parsing can produce non-string values for the same `q` parameter: repeated parameters such as `?q=alpha&q=beta` become an array, and nested syntax such as `?q[x]=beta` becomes an object. The API then returns that array/object in `data.query`, even though search terms should have a stable string shape.

## Reproduction
1. Start the API app.
2. Request `GET /api/search?q=alpha&q=beta`.
3. The response is `200` with `data.query` as `["alpha", "beta"]` instead of a normalized search string.
4. Request `GET /api/search?q[x]=beta`.
5. The response is `200` with `data.query` as `{ "x": "beta" }`.

## Expected behavior
The search endpoint should normalize the query parameter before calling the search service:
- normal string queries stay strings
- repeated `q` parameters use the first value
- non-string/nested values fall back to an empty string

## Scope
Keep the change limited to search query normalization in the controller and focused route regression tests for normal, repeated, nested, and missing query values.

/attempt #743

catcherintheroad-hub:
## Summary
- normalize `GET /api/search` query input before passing it to `globalSearch()`
- keep normal string queries unchanged, use the first repeated `q` value, and fall back to an empty string for nested/non-string values
- add focused search route regression tests and update the API test script to run `*.test.js` files
- add a small SVG demo artifact for bounty review

Fixes #4564
/claim #743

## Demo
- https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-search-query-normalization/demos/search-query-normalization-demo.svg

Before this change, `GET /api/search?q=alpha&q=beta` returned `data.query` as an array and `GET /api/search?q[term]=alpha` returned `data.query` as an object. The new route tests exercise those requests through the API and verify stable string output.

## Validation
```text
npm test -w apps/api
node --check apps/api/src/controllers/searchController.js
node --check apps/api/src/tests/search.test.js
git diff --check
```

`npm test -w apps/api` result: 4 passing tests.


catcherintheroad-hub:
Added a short MP4 demo video for the search query normalization claim:

https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-search-query-normalization/demos/search-query-normalization-demo.mp4

It summarizes the tested regression: repeated or nested `q` input is normalized before search, normal string queries remain unchanged, and the route-level tests pass via `npm test -w apps/api`.

PR #4569, opened 2026-06-05
Validate notification creation payloads

### LINKED ISSUE #4568: Notification creation should reject empty payloads
Parent Algora bounty: #743

## Bug
`POST /api/notifications` forwards `req.body` directly to `createNotification()`. An empty JSON object currently returns `201 Created` and stores a notification containing only the generated id and `read: false`. Blank message/type fields are also accepted because there is no creation payload validation.

## Reproduction
1. Start the API app.
2. Send `POST /api/notifications` with `Content-Type: application/json` and body `{}`.
3. The API returns `201` with data similar to `{ "id": "ntf_...", "read": false }`.

## Expected behavior
Notification creation should require the minimum fields needed for a usable notification record before calling the service layer:
- `userId`: non-empty string
- `type`: non-empty string
- `message`: non-empty string

Malformed payloads should return the shared JSON error shape with HTTP 400, while valid notification creation should continue returning `201` and the server-owned unread state.

## Scope
Keep this limited to notification creation payload validation and focused route tests. Do not change notification list authentication/scoping, notification read-state server ownership, generated ids, or other API routes.

/attempt #743

catcherintheroad-hub:
## Summary
- add a strict Zod schema for POST /api/notifications
- reject missing, blank, extra-field, or oversized notification payloads before persistence
- preserve the existing successful 201 response shape for valid notifications

Fixes #4568
/claim #743

## Demo / evidence
Before this change, POST /api/notifications accepted an empty JSON body and created a notification with undefined fields. The new route-level tests show empty and blank payloads now return a structured 400 response, while a valid payload still creates an unread notification with the submitted userId, type, and message.

Out of scope: notification auth, list scoping, read status updates, and id format changes. This PR only closes the notification creation input-validation gap.

## Validation
- npm test -w apps/api
- node --check apps/api/src/controllers/notificationController.js
- node --check apps/api/src/validators/notification.js
- node --check apps/api/src/tests/notification.test.js
- git diff --check

Result: 4 node:test tests pass locally.


catcherintheroad-hub:
Added demo evidence for the notification creation validation behavior:

https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-notification-create-validation/demos/notification-create-validation-demo.svg

It summarizes the tested regression: empty/blank payloads now return 400, while a valid userId/type/message payload still creates an unread notification with 201.

catcherintheroad-hub:
Added a short MP4 demo video for the notification creation validation claim:

https://raw.githubusercontent.com/catcherintheroad-hub/bug-bounty-earn-1000/fix-notification-create-validation/demos/notification-create-validation-demo.mp4

It summarizes the tested regression: empty or blank notification payloads now return 400, while a valid `userId` / `type` / `message` payload still creates an unread notification with 201.

PR #4761, opened 2026-06-05
fix(api): use collision-resistant service ids

### LINKED ISSUE #4759: Service-generated ids should be collision-resistant and server-owned
Parent Algora bounty: #743

## Bug
Several API services generate public record identifiers with `Date.now()`. Records created in the same millisecond can receive duplicate ids even though they are distinct records. Some create paths also spread caller payloads after server-owned fields, which lets request bodies override fields such as `id`, `status`, `read`, or `sentAt`.

Affected examples include registration/user, job, message, notification, payment, proposal, and review service creation paths.

## Impact
Duplicate or caller-controlled ids can confuse downstream lookups, collide with existing records, and break the assumption that service-created resources own their identifiers and lifecycle fields. Registration also needs one generated user id reused for both the API response and JWT `sub` claim.

## Expected behavior
- Preserve existing id prefixes while using collision-resistant ids.
- Reuse the generated registration id in the access-token subject.
- Apply service-owned fields after caller payload fields so submitted payloads cannot override them.
- Add focused service-level regression coverage for same-millisecond uniqueness and server-owned field override attempts.

## Scope
Keep the change limited to API service id generation and focused tests. Do not change persistence, routing, auth policy, payment provider behavior, or frontend UI.

lijiaqilijiaqi2002-cyber:
/claim #743

Closes #4759

## Summary
- add a shared `createServiceId()` helper that preserves existing id prefixes while using `crypto.randomUUID()`
- replace `Date.now()`-based service ids across registration/user, job, message, notification, payment, proposal, and review creation paths
- keep service-owned fields authoritative by applying generated ids/status/read/sentAt after caller payload fields
- reuse the generated registration id as the JWT `sub` claim
- add focused service-level regression coverage for same-millisecond uniqueness and override attempts

## Validation
- `node --test apps/api/src/tests/serviceIds.test.js`
- `node --check apps/api/src/utils/ids.js`
- `node --check apps/api/src/services/jobService.js`
- `node --check apps/api/src/services/messageService.js`
- `node --check apps/api/src/services/notificationService.js`
- `node --check apps/api/src/services/proposalService.js`
- `node --check apps/api/src/services/reviewService.js`
- `node --check apps/api/src/services/userService.js`

## Demo evidence
The new `serviceIds.test.js` freezes `Date.now()` and creates two jobs in the same millisecond, proving the generated ids remain unique. It also submits caller-controlled `id`, `read`, and `sentAt` payload fields and asserts the service-owned values win.

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -1,12 +1,14 @@
 import { signAccessToken } from "../utils/jwt.js";
+import { createServiceId } from "../utils/ids.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const id = createServiceId("usr");
   return {
-    id: `usr_${Date.now()}`,
+    id,
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
+    token: signAccessToken({ sub: id, role: payload.role })
   };
 }
 
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,3 +1,5 @@
+import { createServiceId } from "../utils/ids.js";
+
 const jobs = [];
 
 export async function listJobs() {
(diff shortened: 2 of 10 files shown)


lijiaqilijiaqi2002-cyber:
Demo GIF: https://tmpfiles.org/dl/wxwj6qY5D6Ej/service-id-hardening-demo.gif

It shows the focused service-id regression test passing: same-millisecond id uniqueness and service-owned field protection.

PR #4805, opened 2026-06-05
fix: add payment intent creation timestamps

### LINKED ISSUE #4803: Payment intents should include server-owned createdAt timestamps

## Problem
`createPaymentIntent()` returns a payment record with `paymentId`, `amount`, `currency`, and `provider`, but it does not include a server-owned creation timestamp. That makes payment intent responses harder to audit, sort, reconcile, or display in transaction history without trusting client-side timing.

## Reproduction
Call `POST /api/payments` with a valid payment body such as:

```json
{
  "amount": 250,
  "currency": "usd"
}
```

Expected: the created payment intent includes a server-generated `createdAt` ISO timestamp.
Actual: no creation timestamp is returned.

## Impact
Billing and transaction-history flows need a canonical server timestamp for auditability. Without it, downstream consumers must infer creation time from local clocks or surrounding metadata.

## Proposed fix
- Add a server-owned `createdAt` ISO timestamp in `createPaymentIntent()`.
- Keep `createdAt` immune to any caller-provided value.
- Add focused tests for valid timestamp creation and server ownership.

Parent bounty: #743

jddark62:
## Summary
- add a server-owned ISO `createdAt` timestamp to payment intent responses
- keep `createdAt` generated by the service instead of accepting caller-provided values
- add focused `node:test` coverage for timestamp creation, spoofing resistance, and the API response envelope

## Verification
- `git diff --check`

## Bounty
/claim #743

Fixes #4803

Note: this runtime does not have npm installed, so I verified the focused regression directly with Node's built-in test runner.


jddark62:
Verification/demo refresh for review:

- Focused regression run: `node --test apps/api/src/tests/payment.test.js`
- Result: 3 passed / 0 failed.
- Coverage: service-level `createdAt` creation, caller-supplied `createdAt` spoofing resistance, and the `createPayment` controller response envelope returning the server timestamp.
- Current PR state from GitHub: open, non-draft, `MERGEABLE`.

This is intentionally scoped to child issue #4803 while claiming parent #743, and includes API-envelope coverage in addition to service behavior so reviewers can confirm the timestamp reaches callers.

jddark62:
Follow-up after comparing the other open timestamp claim: I tightened the tests in `50c629d` so the timestamp is asserted to be both server-owned and within the request window.

Current coverage on this PR now checks:
- service-level `createdAt` is generated as a valid ISO timestamp
- caller-provided `createdAt` is ignored
- generated timestamp falls between request start/end
- controller response envelope includes the generated `createdAt` for API consumers

Local verification:
- `node --test apps/api/src/tests/payment.test.js` (3 passing)

jddark62:
Demo/proof asset added for the #743 review requirement:

Payment createdAt demo

Latest commit: `2fee9ae` (`docs: add payment timestamp demo`).

Fresh validation after adding the demo asset:
- Result: 3 passing / 0 failing.

The implementation coverage remains unchanged: server-owned `createdAt`, caller-supplied timestamp ignored, request-window bounds, and controller response envelope coverage.

PR #4826, opened 2026-06-05
Reject unsupported payment currency codes

### LINKED ISSUE #4825: Payment intent should reject unsupported currency codes
## Bug

`createPaymentIntent` currently accepts any caller-provided `currency` string and echoes it back as a Stripe-backed payment intent. For example, `{ amount: 25, currency: DOGE }` succeeds even though the service only models Stripe-style fiat payment intents.

## Expected behavior

- Supported currencies should continue to create payment intents.
- Missing currency should continue to default to `usd`.
- Unsupported currency codes should be rejected with a clear validation error before a payment intent response is created.
- The service should normalize supported currency codes so mixed-case input cannot bypass the allowlist.

Parent bounty: #743

tonghuaxingdsb:
/claim #743

Fixes #4825

## Summary
- add a payment currency allowlist for Stripe-style payment intents
- default missing currency to `usd` and normalize supported mixed-case currency input
- return a clear `400` validation response when the payment controller receives an unsupported currency
- add focused payment service regression coverage

## Verification
- `node --test apps/api/src/tests/payment-service.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,14 @@
-import { ok } from "../utils/response.js";
-import { createPaymentIntent } from "../services/paymentService.js";
+import { fail, ok } from "../utils/response.js";
+import { PaymentValidationError, createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  try {
+    return ok(res, await createPaymentIntent(req.body), 201);
+  } catch (error) {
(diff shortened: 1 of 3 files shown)


tonghuaxingdsb:

- `node --test apps/api/src/tests/payment-service.test.js` -> 3 tests passed
- `git diff --check main...HEAD` -> passed
- Manual service smoke check: `{ amount: 25, currency: " USD " }` normalizes to `currency: "usd"`

No contributor-side WIP is pending from me unless maintainers want a narrower currency allowlist or a different error response shape.

PR #4832, opened 2026-06-05
Return defensive job list snapshots

### LINKED ISSUE #4831: listJobs should return a defensive snapshot
## Bug

`listJobs()` returns the module-level `jobs` array directly. Any caller that receives the list can mutate it with operations such as `push`, `splice`, or `length = 0`, which corrupts the in-memory job store without going through `createJob` or validation.

## Expected behavior

- `listJobs()` should return a defensive snapshot of the current jobs.
- Mutating the returned list should not mutate the service's internal `jobs` array.
- Existing job creation behavior should continue to work.

Parent bounty: #743

tonghuaxingdsb:
/claim #743

Fixes #4831

## Summary
- make `listJobs()` return a defensive snapshot instead of the backing array
- prevent callers from mutating the in-memory job store through a returned list
- add focused service regression coverage

## Verification
- `node --test apps/api/src/tests/job-service.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,7 +1,7 @@
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return [...jobs];
 }
 
 export async function createJob(payload) {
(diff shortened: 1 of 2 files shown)


tonghuaxingdsb:

- 
ode --test apps/api/src/tests/job-service.test.js -> passed

PR #4834, opened 2026-06-05
Return defensive proposal list snapshots

### LINKED ISSUE #4833: listProposals should return a defensive snapshot
## Bug

`listProposals()` returns the module-level `proposals` array directly. A caller that receives this list can mutate it with `push`, `splice`, or `length = 0`, corrupting the in-memory proposal store without going through `createProposal`.

## Expected behavior

- `listProposals()` should return a defensive snapshot of current proposals.
- Mutating a returned list should not mutate the service's internal `proposals` array.
- Existing proposal creation behavior should continue to work.

Parent bounty: #743

tonghuaxingdsb:
/claim #743

Fixes #4833

## Summary
- make `listProposals()` return a defensive snapshot instead of the backing array
- prevent callers from mutating the in-memory proposal store through a returned list
- add focused service regression coverage

## Verification
- `node --test apps/api/src/tests/proposal-service.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -1,7 +1,7 @@
 const proposals = [];
 
 export async function listProposals() {
-  return proposals;
+  return [...proposals];
 }
 
 export async function createProposal(payload) {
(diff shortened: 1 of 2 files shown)


tonghuaxingdsb:

- 
ode --test apps/api/src/tests/proposal-service.test.js -> passed

PR #4836, opened 2026-06-05
Return defensive review list snapshots

### LINKED ISSUE #4835: listReviews should return a defensive snapshot
## Bug

`listReviews()` returns the module-level `reviews` array directly. A caller that receives this list can mutate it with `push`, `splice`, or `length = 0`, corrupting the in-memory review store without going through `createReview`.

## Expected behavior

- `listReviews()` should return a defensive snapshot of current reviews.
- Mutating a returned list should not mutate the service's internal `reviews` array.
- Existing review creation behavior should continue to work.

Parent bounty: #743

tonghuaxingdsb:
/claim #743

Fixes #4835

## Summary
- return a defensive snapshot from `listReviews()` instead of exposing the module-level array
- add a regression test proving mutations to a returned list do not corrupt stored reviews

## Verification
- `node --test apps\api\src\tests\review-service.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -1,7 +1,7 @@
 const reviews = [];
 
 export async function listReviews() {
-  return reviews;
+  return [...reviews];
 }
 
 export async function createReview(payload) {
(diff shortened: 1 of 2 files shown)


tonghuaxingdsb:

- 
ode --test apps/api/src/tests/review-service.test.js -> passed

PR #4839, opened 2026-06-05
Return defensive message list snapshots

### LINKED ISSUE #4838: listMessages should return a defensive snapshot
## Bug

`listMessages()` returns the module-level `messages` array directly. A caller that receives this list can mutate it with `push`, `splice`, or `length = 0`, corrupting the in-memory message store without going through `sendMessage`.

## Expected behavior

- `listMessages()` should return a defensive snapshot of current messages.
- Mutating a returned list should not mutate the service's internal `messages` array.
- Existing message creation behavior should continue to work.

Parent bounty: #743

tonghuaxingdsb:
/claim #743

Fixes #4838

## Summary
- return a defensive snapshot from `listMessages()` instead of exposing the module-level array
- add a regression test proving mutations to a returned list do not corrupt stored messages

## Verification
- `node --test apps\api\src\tests\message-service.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,7 +1,7 @@
 const messages = [];
 
 export async function listMessages() {
-  return messages;
+  return [...messages];
 }
 
 export async function sendMessage(payload) {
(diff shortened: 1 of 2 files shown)


tonghuaxingdsb:

- 
ode --test apps/api/src/tests/message-service.test.js -> passed

Issue #4243, opened 2026-06-04
Admin metrics endpoint allows any authenticated role


bozicovichsantiago20-oss:
Parent Algora bounty: #743

## Problem

`/api/admin/metrics` is mounted behind `authMiddleware`, but there is no role check after authentication. Any valid bearer token can reach the admin metrics controller, including tokens with a non-admin role.

## Impact

Client or freelancer accounts can access admin-only operational metrics such as open jobs, active freelancers, flagged accounts, and monthly volume.

## Expected behavior

Admin routes should require an authenticated user with `role: admin`; other authenticated roles should receive `403 Forbidden`.

## Scope

Keep the fix limited to admin route authorization and add regression tests proving unauthenticated requests are rejected, non-admin tokens receive `403`, and admin tokens can access metrics.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

/attempt #743

PR #4899, opened 2026-06-05
Require admin role for metrics

### LINKED ISSUE #4893: Admin metrics should require admin role
Parent Algora bounty: #743

## Bug
`/api/admin/metrics` currently only verifies that a request has a valid JWT. Any authenticated client or freelancer token can reach admin metrics because `adminRoutes` uses `authMiddleware` but does not check `req.user.role`.

## Expected behavior
Admin routes should reject non-admin authenticated users with `403` while allowing `role=admin` tokens to access metrics.

## Scope
Add a focused admin-role guard, apply it to admin routes after authentication, and add API regression tests for missing token, non-admin token, and admin token behavior.

/attempt #743

bozicovichsantiago20-oss:
/claim #743

Fixes #4893.

## Summary
- Adds a focused `requireAdmin` guard after API authentication.
- Applies the guard to `/api/admin/*` routes so non-admin JWTs receive `403`.
- Adds regression coverage for missing token, non-admin token, and admin token access to `/api/admin/metrics`.
- Updates the API test script to execute test files on Windows/Node instead of treating the test directory as a module.

## Validation
- `git diff --check`
- `npm.cmd test -w apps/api` passed 4/4
- `node --test apps/api/src/tests/admin-routes.test.js` passed 3/3

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,11 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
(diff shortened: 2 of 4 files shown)


bozicovichsantiago20-oss:
Closing this as a duplicate of my earlier scoped submission #4244 / issue #4243 for the same admin metrics role-guard finding. Keeping the earlier PR as the active submission to avoid duplicate review noise.

PR #5156, opened 2026-06-06
Let health checks bypass API rate limiting

### LINKED ISSUE #5155: Health checks should bypass the API rate limiter (catcherintheroad-hub reissue via #743)
Parent bounty: #743

This is an independent reissue of #4467 because the original issue is limited to its creator and asks non-authors to create another issue with the same contents.

`apps/api/src/app.js` mounts the shared API rate limiter before `GET /health`. That means repeated health probes can exhaust the normal application request quota and make the health endpoint return `429 Too Many Requests` instead of reporting process liveness. Load balancers, uptime monitors, and orchestration probes should be able to check `/health` independently of user API throttling.

Scope:
- Keep `/health` available before the shared API limiter is applied.
- Preserve rate limiting for normal `/api/*` routes.
- Add focused tests showing `/health` still returns 200 after the API quota is exhausted while an API route receives 429.
- Do not change route payload shape or unrelated middleware behavior.

Payout route if accepted: Base USDC `0x5d8a819bd8d056119626c34b384739ab15505093`

catcherintheroad-hub:
## Summary
- register `/health` before the shared API rate limiter
- keep the existing limiter on normal `/api/*` routes
- add regression coverage showing `/api/search` can reach 429 while `/health` still returns 200

## Validation
- `PATH="/Users/jiahuiwu/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" node --test apps/api/src/tests/health.test.js`
- `PATH="/Users/jiahuiwu/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" node --test apps/api/src/tests/*.test.js`
- `git diff --check`

/claim #743
Closes #5155
Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -21,11 +21,12 @@ export function createApp() {
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
-  app.use(apiLimiter);
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
   });
+
+  app.use(apiLimiter);
 
   app.use("/api/auth", authRoutes);
   app.use("/api/users", userRoutes);
--- a/apps/api/src/tests/health.test.js
+++ b/apps/api/src/tests/health.test.js
@@ -22,3 +22,35 @@ test("GET /health returns ok payload", async () => {
     server.close((error) => (error ? reject(error) : resolve()));
   });
 });
+
+test("GET /health bypasses the shared API rate limiter", async () => {
+  const app = createApp();
+  const server = app.listen(0);
+
+  await new Promise((resolve, reject) => {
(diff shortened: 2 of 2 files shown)


catcherintheroad-hub:
Validation refresh for this #743 claim:

- `node --test apps/api/src/tests/health.test.js` -> 2 passed (`GET /health returns ok payload`, `GET /health bypasses the shared API rate limiter`).
- `node --check apps/api/src/app.js` -> passed.
- `node --check apps/api/src/tests/health.test.js` -> passed.
- `git diff --check origin/main...HEAD` -> passed.

Current PR metadata remains open, non-draft, mergeable/clean, with the `🙋 Bounty claim` label and `update-leaderboard` success.

PR #5158, opened 2026-06-06
Return JSON from API rate limiter

### LINKED ISSUE #5157: Rate limiter should return JSON error payloads (catcherintheroad-hub reissue via #743)
Parent bounty: #743

This is an independent reissue of #4507/#4477 because the original issues are limited to their creators and ask non-authors to create another issue with the same contents.

The shared API rate limiter currently returns express-rate-limit's default plain-text 429 response when a client exceeds the quota. Other API failures use the project's JSON error envelope (`{ success: false, message: ... }`), so rate-limit failures are inconsistent and harder for API clients to handle.

Scope:
- Configure the shared `apiLimiter` to return a JSON error payload for 429 responses.
- Preserve the existing 15-minute window, 200-request limit, draft-7 standard headers, and disabled legacy headers.
- Add focused coverage proving a rate-limited API route returns HTTP 429 with the JSON envelope.
- Do not change route behavior, authentication, health check behavior, or unrelated middleware ordering.

Payout route if accepted: Base USDC `0x5d8a819bd8d056119626c34b384739ab15505093`

catcherintheroad-hub:
## Summary
- configure the shared API limiter to use the project JSON error envelope for 429 responses
- preserve the existing rate-limit window, quota, and header configuration
- add regression coverage for the 429 JSON response body and content type

## Validation
- `PATH="/Users/jiahuiwu/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" node --test apps/api/src/tests/health.test.js`
- `PATH="/Users/jiahuiwu/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" node --test apps/api/src/tests/*.test.js`
- `git diff --check`

/claim #743
Closes #5157
Parent bounty: #743

### DIFF
--- a/apps/api/src/middleware/rateLimit.js
+++ b/apps/api/src/middleware/rateLimit.js
@@ -1,8 +1,10 @@
 import rateLimit from "express-rate-limit";
+import { fail } from "../utils/response.js";
 
 export const apiLimiter = rateLimit({
   windowMs: 15 * 60 * 1000,
   limit: 200,
   standardHeaders: "draft-7",
-  legacyHeaders: false
+  legacyHeaders: false,
+  handler: (req, res) => fail(res, "Too many requests", 429)
 });
--- a/apps/api/src/tests/health.test.js
+++ b/apps/api/src/tests/health.test.js
@@ -22,3 +22,32 @@ test("GET /health returns ok payload", async () => {
     server.close((error) => (error ? reject(error) : resolve()));
   });
 });
+
+test("shared API rate limiter returns JSON error payloads", async () => {
+  const app = createApp();
+  const server = app.listen(0);
+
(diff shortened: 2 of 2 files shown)


catcherintheroad-hub:
Validation refresh for this #743 claim:

- `node --test apps/api/src/tests/health.test.js` -> 2 passed (`GET /health returns ok payload`, `shared API rate limiter returns JSON error payloads`).
- `node --check apps/api/src/middleware/rateLimit.js` -> passed.
- `node --check apps/api/src/tests/health.test.js` -> passed.
- `git diff --check origin/main...HEAD` -> passed.

The regression confirms 429 responses now use the shared JSON API envelope instead of the default text response.

Issue #4723, opened 2026-06-05
fix: Prevent admin role self-assignment during registration (mirror of #4718)


Thecesar85:
﻿## Bug

Public registration currently accepts role: "admin", which lets a new user self-assign admin privileges.

## Expected

- Registration should only allow public roles (client and freelancer)
- Should reject admin in the request body

## Related

- Mirrors issue #4718
- Parent Algora bounty: #743

Created because issue #4718 is limited to the creator of that issue, and the repository explicitly asks other contributors to create another issue with the same contents and refer to #743.

Thecesar85:
Opened PR: https://github.com/SecureBananaLabs/bug-bounty/pull/5163

This fixes the public registration path so `role: "admin"` is no longer accepted from signup requests.

Validation used:
- admin self-assignment now rejected
- public roles still succeed

Linked bounty thread: #743

PR #5163, opened 2026-06-06
fix(api): block admin self-assignment during public registration

Thecesar85:
/claim #743

## Summary

Fixes a privilege-escalation bug in public registration where a new user could submit `role: "admin"` and self-assign admin privileges during signup.

## Problem

Public registration should only allow end-user roles. Before this change, the request body accepted an admin role value, which made the registration path too permissive.

## Fix

- reject `admin` in the registration payload
- return a clear validation error instead of creating an invalid privileged user

## Validation

- added regression coverage for admin self-assignment rejection
- confirmed public roles still register successfully
- focused test command used:

```bash
node --test --test-reporter spec apps/api/src/tests/auth.test.js apps/api/src/tests/health.test.js
```

Result at preparation time: `3 pass / 0 fail`

## Related

- Closes `#4723`
- Related bounty thread: `#743`

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -2,16 +2,24 @@ import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
 import { ok } from "../utils/response.js";
 
-export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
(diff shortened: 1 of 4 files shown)


Thecesar85:
Summary
- links mirror issue #4723 to this PR
- blocks public self-assignment to admin-only work during registration

Why
This keeps the issue -> PR trail explicit and makes review easier: the fix is limited to registration-time authorization and prevents regular users from claiming admin-only assignments.

Issue #5162, opened 2026-06-06
Reject upload requests that omit the multipart file field


Thecesar85:
## Bug

`POST /api/uploads` currently returns `201` with `success: true` even when the multipart request does not include a `file` field. The response reports `status: "no-file"`, but it is still treated as a successful upload by HTTP status and response shape.

## Expected fix

- Reject upload requests that do not include a file with a clear `400` response
- Keep successful upload responses reserved for requests that actually include `req.file`
- Add route-level regression coverage for the missing-file case

## Related

- Mirrors original issue `#4680`
- Parent Algora bounty: `#743`

Created because issue `#4680` is limited to the creator of that issue, and the repository explicitly asks other contributors to create another issue with the same contents and refer to `#743`.

Thecesar85:
Opened PR: https://github.com/SecureBananaLabs/bug-bounty/pull/5164

This mirrors the original `#4680` report and fixes the case where `POST /api/uploads` returned a success response even though no file was included.

Validation used:
- missing file now returns `400`
- normal uploads still succeed

Related bounty thread: #743

PR #5164, opened 2026-06-06
fix(api): reject empty upload requests without a file

Thecesar85:
/claim #743

## Summary

Rejects upload requests that omit the multipart `file` field so the API no longer reports a successful upload when no file was actually received.

## Problem

`POST /api/uploads` returned `201` and `success: true` even when the request did not include a file. That makes the route look successful to clients despite no upload happening.

## Fix

- return `400` when `req.file` is missing
- reserve success responses for real uploads only
- add regression coverage for the missing-file path

## Validation

- confirmed missing-file requests now fail with `400`
- confirmed normal upload requests still succeed
- focused test command used:

```bash
node --test --test-reporter spec apps/api/src/tests/uploads.test.js apps/api/src/tests/health.test.js
```

Result at preparation time: `3 pass / 0 fail`

## Related

- Closes `#5162`
- Mirrors original issue `#4680`
- Related bounty thread: `#743`

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "A file upload is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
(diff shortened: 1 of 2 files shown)


Thecesar85:
Summary
- links mirror issue #5162 to this PR
- rejects empty upload requests when no file is actually provided

Why
This keeps the issue -> PR trail explicit and makes review easier: the fix is intentionally narrow and closes the empty-upload path without changing the normal upload flow.

PR #5277, opened 2026-06-06
Return JSON 405 for unsupported API methods

### LINKED ISSUE #5270: Known API routes should return JSON 405 for unsupported methods

## Problem
Known API collection routes only register their supported HTTP methods. When a client calls the same route with an unsupported method, Express falls through to its default HTML 404 response instead of returning the structured API JSON envelope.

Example: `PUT /api/jobs` targets an existing API route path, but it is not an allowed method for the jobs collection. The response should be a method error clients can handle consistently, not an HTML 404 page.

## Evidence
- File: `apps/api/src/routes/jobRoutes.js`
- Signal: `/api/jobs` defines `GET /` and `POST /`, but no method fallback for unsupported methods.
- Current behavior: unsupported methods fall through to default HTML 404.

## Suggested Fix
Add a focused method-not-allowed fallback for known API collection routes and return the existing structured JSON error envelope with status `405` plus an `Allow` header.

## Acceptance Criteria
- Unsupported methods on known API collection routes return JSON `405`.
- The response uses the existing `{ success: false, message: ... }` envelope.
- The `Allow` header lists supported methods.
- Supported methods keep their current behavior.
- The change remains scoped to API routing behavior.

## Source
Created from #743 low-hanging-fruit workflow.

wowsofine:
## Summary
- add a reusable methodNotAllowed middleware that returns the existing JSON error shape with HTTP 405
- attach explicit method handling to known API collection routes and auth/admin endpoints
- add coverage proving unsupported methods on known API paths return JSON 405 while supported methods still work

## Bounty
/claim #743
Closes #5270

## Validation
- node --test apps/api/src/tests/*.js
- node --check apps/api/src/middleware/methodNotAllowed.js && for f in apps/api/src/routes/*.js apps/api/src/tests/methodNotAllowed.test.js; do node --check "$f" || exit 1; done
- git diff --check

### DIFF
--- a/apps/api/src/middleware/methodNotAllowed.js
+++ b/apps/api/src/middleware/methodNotAllowed.js
@@ -1,0 +1,10 @@
+import { fail } from "../utils/response.js";
+
+export function methodNotAllowed(allowedMethods) {
+  const allowHeader = allowedMethods.join(", ");
+
+  return (req, res) => {
+    res.set("Allow", allowHeader);
+    return fail(res, "Method not allowed", 405);
+  };
+}
--- a/apps/api/src/routes/adminRoutes.js
+++ b/apps/api/src/routes/adminRoutes.js
@@ -1,8 +1,11 @@
 import { Router } from "express";
 import { metrics } from "../controllers/adminController.js";
 import { authMiddleware } from "../middleware/auth.js";
+import { methodNotAllowed } from "../middleware/methodNotAllowed.js";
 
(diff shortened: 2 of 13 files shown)


wowsofine:
Hi maintainers, quick follow-up on this scoped fix. The latest check rollup is passing, and I am happy to adjust the implementation if you want a different response shape or additional coverage for unsupported methods.

Thanks for reviewing.

PR #5310, opened 2026-06-06
Harden proposal creation input boundary

### LINKED ISSUE #5309: Proposal creation should whitelist public fields before persistence
Parent bounty: #743

/attempt #743

## Bug
`POST /api/proposals` forwards `req.body` directly into `createProposal`. The service persists `{ id: ..., ...payload }`, so a caller can override the server-generated `id` and add internal fields such as `status`, `acceptedAt`, `createdAt`, or `adminNotes`.

## Evidence
- `apps/api/src/controllers/proposalController.js` calls `createProposal(req.body)`.
- `apps/api/src/services/proposalService.js` spreads untrusted payload after the generated id.

## Expected
Proposal creation should parse a strict schema based on the public create fields in the Proposal model: `coverLetter`, `bidAmount`, `estDuration`, `jobId`, and `freelancerId`. The service should construct records from whitelisted fields only and set server-owned fields itself. Unknown or invalid fields should return controlled 400 JSON responses.

## Scope
Keep the fix limited to proposal creation validation and service assignment. Add route tests proving injected internal fields are rejected and valid proposals still return 201.

Peter7896:
/claim #743 
Closes #5309 
Summary
- Adds strict proposal creation validation for public Proposal fields.
- Rejects unknown or invalid POST /api/proposals fields with a controlled 400 response.
- Builds proposal records from whitelisted fields so request bodies cannot override id or inject status or acceptedAt.
- Updates the API test script so npm test -w apps/api runs the test files directly.

Demo
Short demo video
https://raw.githubusercontent.com/Peter7896/bug-bounty/peter7896/proposal-input-boundary/docs/demos/proposal-input-boundary-demo.mp4

Validation
npm test -w apps/api
git diff --check


Peter7896:
Closing this PR as part of repository housekeeping because the branch is no longer mergeable against the current base and the work has become stale. The implementation and validation record remain available here for future reference. Thanks to everyone who reviewed it.

PR #5317, opened 2026-06-06

### LINKED ISSUE #5316: fix: prevent admin role self-assignment during registration
## Bug

Additionally, `registerUser()` builds the returned `id` and the JWT `sub` with two separate `Date.now()` calls. If time advances between those calls, the API can return one user id while signing an access token for a different subject.

## Expected fix

1. Remove `admin` from the allowed roles in `registerSchema` (validators/auth.js)
2. Generate the user id once and reuse it for both the response and the token (services/authService.js)
3. Add Zod error handling in the error middleware to return proper 400 responses for validation errors
4. Add async error handling in controllers to ensure errors propagate to the error handler
5. Add route-level tests for admin rejection, valid role acceptance, and token subject matching

## Scope

This issue is limited to the registration validator, auth service, error handler, and auth controller. Tests added in `apps/api/src/tests/auth.test.js`.

Parent bounty: #743

friendlygeorge:
## Summary

Fixes admin role self-assignment vulnerability in public registration and token subject mismatch.

## Changes

1. **Validator (auth.js):** Removed  from allowed registration roles. Only  and  are now accepted.

2. **Auth service (authService.js):** Generate user ID once and reuse for both the response uid=0(root) gid=0(root) groups=0(root) and JWT  claim. Previously, two separate  calls could produce different values.

3. **Error handler (errorHandler.js):** Added Zod error handling to return proper 400 responses with validation details instead of generic 500 errors.

4. **Auth controller (authController.js):** Added async error wrapper to ensure thrown errors (including Zod validation errors) propagate to the error handler.

5. **Tests (auth.test.js):** Added 4 tests:
   - Registration rejects admin role (returns 400)
   - Registration accepts client role
   - Registration accepts freelancer role
   - Token subject matches returned user ID

## Test Results

All 5 tests pass (4 new + 1 existing health check).

Closes #5316

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -2,17 +2,21 @@ import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
 import { ok } from "../utils/response.js";
 
-export async function register(req, res) {
+function asyncHandler(fn) {
+  return (req, res, next) => Promise.resolve(fn(req, res, next)).catch(next);
+}
+
+export const register = asyncHandler(async (req, res) => {
   const payload = registerSchema.parse(req.body);
   const result = await registerUser(payload);
   return ok(res, result, 201);
-}
+});
 
-export async function login(req, res) {
+export const login = asyncHandler(async (req, res) => {
   const payload = loginSchema.parse(req.body);
   const result = await loginUser(payload);
   return ok(res, result);
-}
+});
 
 export async function oauthCallback(req, res) {
   return ok(res, {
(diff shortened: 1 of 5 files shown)


friendlygeorge:
Hi! Just checking in — this PR fixes an admin role self-assignment vulnerability (issue #5316) and has been open since June 6. All tests pass. Happy to make any changes or answer questions. Thanks for maintaining the bounty program!

friendlygeorge:
Bumping — this fixes admin role self-assignment (issue #5316). All 5 tests pass. Let me know if you need anything!

friendlygeorge:
Closing — stale PR, no review activity.

Issue #5319, opened 2026-06-06
fix: prevent caller from overriding server-owned notification id and read state


friendlygeorge:
## Bug

The notification service creates records with `{ id: generated, read: false, ...payload }`, so a request body can override both the generated notification id and the initial unread state. This allows callers to create notifications already marked as read or collide with reserved/server-generated ids.

## Fix

Strip caller-supplied `id` and `read` fields from the payload before spreading into the notification object. The server always generates the id and sets `read: false` for new notifications.

## Tests Added

- Notification created with server-generated id and read=false
- Caller-supplied id is ignored (injection attempt)
- Caller-supplied read=true is ignored (read state injection attempt)
- Empty body handled gracefully

All 5 tests pass (1 existing + 4 new).

Related: #2762, Parent: #743

PR #5320, opened 2026-06-06
fix: prevent caller from overriding server-owned notification id and read state

friendlygeorge:
## Bug

The notification service creates records with `{ id: generated, read: false, ...payload }`, so a request body can override both the generated notification id and the initial unread state. This allows callers to create notifications already marked as read or collide with reserved/server-generated ids.

## Fix

Strip caller-supplied `id` and `read` fields from the payload before spreading into the notification object. The server always generates the id and sets `read: false` for new notifications.

## Tests Added

- Notification created with server-generated id and read=false
- Caller-supplied id is ignored (injection attempt)
- Caller-supplied read=true is ignored (read state injection attempt)
- Empty body handled gracefully

All 5 tests pass (1 existing + 4 new).

Closes #2762
Parent: #743

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,8 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const { id: _callerId, read: _callerRead, ...safePayload } = payload || {};
+  const notification = { id: `ntf_${Date.now()}`, read: false, ...safePayload };
   notifications.push(notification);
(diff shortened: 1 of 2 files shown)


friendlygeorge:
Hi! This PR fixes server-owned fields being overridable by callers (issue #5319). The notification id and read state are now stripped from caller input. All tests pass. Been open since June 6 — let me know if you'd like any changes!

friendlygeorge:
Friendly bump — prevents caller from overriding server-owned notification fields (issue #5319). All 5 tests pass. Open to feedback!

friendlygeorge:
Closing — stale PR, no review activity.

Issue #2849, opened 2026-05-31
Freelancer profile route should resolve mock profiles by username


daniel7an:
Parent bounty: #743

## Bug
The `/freelancers/[username]` route renders placeholder text with the requested username instead of looking up the selected freelancer from `apps/web/lib/mock.ts`. Search cards link to `/freelancers/maya-dev` and `/freelancers/jordan-ux`, but the profile page does not show matching skills or hourly rate. Unknown usernames also appear successful.

## Expected
- Known freelancer usernames render the matching mock profile details.
- The profile page exposes useful skills/rate context instead of generic placeholder copy.
- Unknown usernames render a clear not-found fallback.

## Scope
Keep the fix limited to the web freelancer profile route unless the existing mock data needs a minimal shape adjustment.

Related reissue: #2763

guanglongcf-coder:
C:/Program Files/Git/attempt #2849 — Implemented mock profile resolution

auruminternum:
/attempt #2849

Plan: resolve freelancer profiles from apps/web/lib/mock.ts by username, render the matching skills and hourly rate for known profiles, and show a clear not-found fallback for unknown usernames. Scope stays limited to the web freelancer profile route and any minimal mock data usage needed.

mkcash:
I'll work on this fix and submit a PR. Payment info: USDC (BNB Chain): 0xAC1d857cf2A69Ef2C0698F100a644bfE12A6123d / PayPal: ljwtitan@hotmail.com

mkcash:
I'll work on this fix and submit a PR. Payment info: USDC (BNB Chain): 0xAC1d857cf2A69Ef2C0698F100a644bfE12A6123d / PayPal: ljwtitan@hotmail.com

Asobu01:
Happy to claim SecureBananaLabs/bug-bounty#2849.
I read "Freelancer profile route should resolve mock profiles by username" and my first move is reproducing the reported issue first, then applying a focused fix with clear validation.
ETA: first reviewable PR in about 8h.
I can post a reproducible checklist/log snippet in ~30 minutes so you can verify execution speed before the full PR.
I reduce back-and-forth by sharing checkpoints and validation evidence early.
I will include repro + root-cause summary, focused patch (plus tests/docs if needed), and review-ready PR with validation notes.
Thanks. @Asobu01

luw8072-gif:
D:/Git/claim #2849 — fixed in PR

[...]

PR #5383, opened 2026-06-07
fix: resolve mock freelancer profiles by username (fixes #2849)

Fuwengg:
Look up mock freelancer data by username. Show profile with skills/rate. Show not-found fallback for unknown usernames.


MikeGarciaAGM (approved):
Looks good to me. The route now resolves mock freelancer profiles by username, keeps the expected profile fields, and shows a clear not-found fallback for unknown usernames.

PR #5389, opened 2026-06-07
Return defensive snapshots from list services

### LINKED ISSUE #5388: In-memory list services should return defensive snapshots
## Bug

The in-memory list services return their backing arrays directly (`users`, `jobs`, `proposals`, `reviews`, `messages`, and `notifications`). A caller that receives one of these arrays can mutate it with `push`, `pop`, or item reassignment and silently corrupt the service's internal state without going through the create helpers.

## Expected behavior

- Listing helpers should return defensive array snapshots instead of the backing store references.
- Created records should still be visible in subsequent list calls.
- Mutating a returned list should not add/remove records from the service's internal store.

## Scope

Keep the fix limited to the in-memory service list return values and focused regression coverage for the snapshot behavior.

Parent bounty: #743

elianguitarra:
/claim #743

Fixes #5388

## Summary
- Return defensive array snapshots from all in-memory list services: users, jobs, proposals, reviews, messages, and notifications.
- Preserve created records while preventing callers from mutating the backing store through returned list references.
- Add dependency-free `node:test` regression coverage for all six services.

## Validation
- `node --test apps/api/src/tests/service-list-snapshots.test.js` passed.
- `node --check` passed for all touched service files and the new regression test.
- `git diff --check` passed after adding the MP4.
- MP4 evidence: https://raw.githubusercontent.com/elianguitarra/bug-bounty/codex/defensive-list-snapshots-5388/demos/issue-5388-defensive-list-snapshots-demo.mp4
- `ffprobe`: H.264, 1280x720, 6.000s.

## Payout
Payment preference if accepted: USDC on Base/EVM to 0xa3d7745af6E77ce825f0AF6DB94bA8073355E022.


elianguitarra:
﻿Review/merge follow-up for this #743 scoped bounty claim: this PR is open, clean, has the bounty-claim label, closes creator-owned issue #5388, includes focused validation, and has MP4 evidence linked in the body.

Please let me know if any remaining change is needed for acceptance or merge. Current payout preference if accepted: USDC on Base/EVM to `0xa3d7745af6E77ce825f0AF6DB94bA8073355E022`.

PR #5399, opened 2026-06-07
Fix: Upload endpoint should reject requests without a file

nkar123412-hub:
This PR ensures that the /api/uploads endpoint returns a 400 Bad Request status when no file is provided in the request body, instead of returning a 201 Created with a 'no-file' status.


nkar123412-hub:

PR #5415, opened 2026-06-07
fix(api): preserve server-owned notification fields

sh962214-hub:
Fixes #2762.

Ensures notification creation keeps server-owned fields authoritative.

Changes:

- caller-supplied `id` can no longer override the generated notification id
- caller-supplied `read` can no longer create already-read notifications
- caller-owned fields such as `message` and `type` are still preserved
- adds focused service tests for the ownership behavior

Verification:

```powershell
```

Result: 4 tests passed.

/claim #2762

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,7 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const notification = { ...payload, id: `ntf_${Date.now()}`, read: false };
   notifications.push(notification);
   return notification;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The payload spread order now keeps server-owned id/read authoritative while preserving caller-owned fields, and the added tests cover the intended behavior.

MikeGarciaAGM (approved):
Looks good. Server-owned fields now take precedence over caller input, and the tests cover override and preservation behavior.

MikeGarciaAGM:
Status refresh: PR #5415 is approved and the \update-leaderboard\ check is passing. The change remains narrowly scoped to server-owned notification fields, and the regression tests cover both override and preservation behavior. If there are no remaining objections, this looks ready to merge.

MikeGarciaAGM:
Quick status refresh: PR #5415 is still approved, mergeable clean, and the update-leaderboard check is passing. The change remains narrowly scoped and the regression coverage is already in place, so this still looks ready for maintainer merge when convenient.

PR #5419, opened 2026-06-07
fix(api): align registration token subject

sh962214-hub:
Fixes #2768.

Ensures registration creates the user id once and signs the access token with that same id as the JWT subject.

Changes:

- generates the registration id once in `registerUser`
- returns that same id in the registration response
- signs the token with the same id in `sub`
- adds a focused regression test that stubs `Date.now()` across two values and verifies `decoded.sub === result.id`

Verification:

```powershell
```

Result: 2 tests passed.

/claim #2768

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const id = `usr_${Date.now()}`;
+
   return {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (changes requested):
Requesting changes: this only updates authService.js and does not include the focused regression coverage required by the issue. The bounty explicitly asks for a test proving the decoded token subject matches result.id, and this PR currently doesn't demonstrate that behavior. Without the regression test, I don't think this is acceptance-ready yet.

PR #5478, opened 2026-06-07
fix: validate search query q parameter

codeboost-tr:
This PR implements validation for the search query parameter 'q'. It ensures 'q' is a single string, trimmed, and capped at 200 characters. Repeated or object-shaped parameters are rejected with a 400 error. Regression tests included.


MikeGarciaAGM (approved):

codeboost-tr:
Hi! Ready for review when you have a moment. Thanks!

PR #5479, opened 2026-06-07
fix: expose Node-loadable entrypoint in @freelanceflow/ui

codeboost-tr:
This PR fixes the issue where @freelanceflow/ui was not importable in Node/ESM environments. It adds a tsc build step to generate ESM output in a 'dist' directory, updates package.json with proper exports, and ensures relative imports use explicit extensions for ESM compatibility. Regression verified with node -e import.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,29 @@
+name: CI
+
+on:
+  push:
+    branches: [main, master]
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):

codeboost-tr:
Hi! Ready for review when you have a moment. Thanks!

PR #5505, opened 2026-06-07
fix: use consistent user id in registration response and JWT subject

### LINKED ISSUE #5325: Register response token subject should match returned user id
Parent bounty: #743

This issue is limited to its creator. This means only the issue creator can attempt to solve this issue. If you want to participate, please create an identical issue and refer to issue "Low-hanging fruit automation" #743 for more information.

## Problem

## Expected behavior

## Scope

## Validation

lie28-max:
registerUser was calling Date.now() twice - once for the response id and once for the JWT sub - which produced mismatched values. generate the id once and reuse it for both.

Closes #5325


lie28-max:

PR #5507, opened 2026-06-07
fix: default payment currency to uppercase USD

### LINKED ISSUE #5348: Bug: paymentService defaults currency to lowercase usd, schema expects uppercase USD
In apps/api/src/services/paymentService.js, createPaymentIntent defaults to lowercase usd, but Prisma schema Payment model has @default(USD) (uppercase).

Line 6: currency: payload.currency ?? usd — lowercase

Impact: Inconsistency with ISO 4217 convention and database schema.

Fix: Change default to uppercase USD.

lie28-max:
createPaymentIntent defaults currency to lowercase "usd" but the schema expects uppercase "USD". fixed the default to match.

Closes #5348

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -3,7 +3,7 @@ export async function createPaymentIntent(payload) {
   return {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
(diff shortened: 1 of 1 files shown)


lie28-max:

PR #5508, opened 2026-06-07
fix: accept multiple spaces after Bearer scheme in auth header

### LINKED ISSUE #5304: Auth middleware should accept multiple spaces after Bearer scheme (bnpl7 scoped #743)
## Bug

`apps/api/src/middleware/auth.js` currently checks for `Authorization` headers with `startsWith("Bearer ")` and then extracts the token with `slice(7)`. That assumes exactly one space between the `Bearer` auth scheme and the JWT.

RFC 6750-style bearer credentials allow one or more spaces between the scheme and credentials. A valid token sent as `Authorization: Bearer  <token>` (two spaces) is currently passed to `verifyAccessToken` with a leading space and is rejected as `Invalid token`.

## Reproduction

2. Generate a valid access token with `signAccessToken({ sub: 'usr_test', role: 'admin' })`.
3. Request `GET /api/admin/metrics` with `Authorization: Bearer  <token>` (two spaces).
4. The response is `401 Invalid token`, while the same token with one space succeeds.

## Expected

The middleware should accept one or more spaces after the `Bearer` scheme and extract the JWT without leading whitespace. This should be a narrow parser fix and should not broaden access beyond valid bearer tokens.

Parent bounty: #743

lie28-max:
the auth middleware used startsWith("Bearer ") which only matches a single space. clients sometimes send extra whitespace after the scheme. switched to a regex that matches Bearer followed by one or more spaces, and strips the scheme before extracting the token.

Closes #5304

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -3,12 +3,12 @@ import { verifyAccessToken } from "../utils/jwt.js";
 
 export function authMiddleware(req, res, next) {
   const authHeader = req.headers.authorization;
-  if (!authHeader?.startsWith("Bearer ")) {
+  if (!authHeader || !/^Bearer\s+/.test(authHeader)) {
     return fail(res, "Unauthorized", 401);
   }
 
   try {
-    req.user = verifyAccessToken(authHeader.slice(7));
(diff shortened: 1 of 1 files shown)


lie28-max:

PR #5509, opened 2026-06-07
fix: return JSON 405 for unsupported methods on known API routes

### LINKED ISSUE #5282: Known API collection routes should return JSON 405 for unsupported methods

## Problem
Known API collection routes only register their supported HTTP methods. When a client calls the same route with an unsupported method, Express falls through to its default HTML 404 response instead of returning the structured API JSON envelope.

Example: `PUT /api/jobs` targets an existing API route path, but it is not an allowed method for the jobs collection. The response should be a method error clients can handle consistently, not an HTML 404 page.

## Evidence
- File: `apps/api/src/routes/jobRoutes.js`
- Signal: `/api/jobs` defines `GET /` and `POST /`, but no method fallback for unsupported methods.
- Current behavior: unsupported methods fall through to default HTML 404.

## Suggested Fix
Add a focused method-not-allowed fallback for known API collection routes and return the existing structured JSON error envelope with status `405` plus an `Allow` header.

## Acceptance Criteria
- Unsupported methods on known API collection routes return JSON `405`.
- The response uses the existing `{ success: false, message: ... }` envelope.
- The `Allow` header lists supported methods.
- Supported methods keep their current behavior.
- The change remains scoped to API routing behavior.

## Source
Created from #743 low-hanging-fruit workflow.

lie28-max:
sending a DELETE or PUT to a collection route like /api/users would fall through to Express default HTML error page. added a catch-all after the route mounts that returns a JSON 405 for any unmatched method on /api/*.

Closes #5282

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -3,6 +3,7 @@ import express from "express";
 import helmet from "helmet";
 import { apiLimiter } from "./middleware/rateLimit.js";
 import { errorHandler } from "./middleware/errorHandler.js";
+import { fail } from "./utils/response.js";
 import { authRoutes } from "./routes/authRoutes.js";
 import { userRoutes } from "./routes/userRoutes.js";
 import { jobRoutes } from "./routes/jobRoutes.js";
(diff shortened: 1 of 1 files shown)


lie28-max:

PR #5510, opened 2026-06-07
fix: reject upload requests without a file

### LINKED ISSUE #5298: Upload endpoint should reject requests without a file
## Problem

`POST /api/uploads` currently returns a successful 201 response even when no multipart `file` field is provided. The response has `status: "no-file"`, but it is still wrapped as a successful upload. That makes clients treat missing-file submissions as completed uploads and hides user/input errors.

## Expected behavior

The upload endpoint should return a 400 failure response when the request does not include a file, and continue returning 201 only for real uploaded files.

## Proposed fix

- Return `fail(res, "File is required", 400)` when `req.file` is missing.
- Keep the existing success payload for valid uploads.
- Add API tests for missing-file rejection and successful multipart upload.

Parent bounty: #743

lie28-max:
the upload endpoint returned 201 with "no-file" status when no file was attached. now it returns 400 with an error message instead.

Closes #5298

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "No file provided", 400);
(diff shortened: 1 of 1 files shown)


lie28-max:

PR #5532, opened 2026-06-08
fix: reject malformed and oversized search q parameters (Fixes #5476)

### LINKED ISSUE #5476: Search query should reject malformed and oversized q parameters
Parent bounty: #743

Related reports: #5423, #5421

## Problem

`GET /api/search` passes `req.query.q` directly into the search service. Express can parse repeated query parameters into arrays and object-shaped query strings into objects, and very long values are accepted without bounds.

## Expected

- Normalize `q` to a single trimmed string.
- Reject repeated or object-shaped `q` parameters.
- Enforce a 200-character maximum query length.
- Add route coverage for valid and invalid query values.

SyntaxHQDEV:
Fixes #5476

Adds a Zod validator for the search route to properly parse and sanitize the \q\ parameter. Normalizes \q\ to a single trimmed string, limits length to 200 characters, and explicitly rejects object-shaped or repeated parameters to prevent denial-of-service edge cases.

Wallet: 0x7F603CD46BC9C2ff735b8B347ed9F19430A0A131

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,17 @@
 import { ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchSchema } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const result = searchSchema.safeParse(req.query);
+  
+  if (!result.success) {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Normalizing and bounding q at the route boundary is the right fix, and the tests cover the malformed and oversized cases cleanly.

PR #5554, opened 2026-06-08
fix(uploads): reject requests without file with 400 + tests

### LINKED ISSUE #5162: Reject upload requests that omit the multipart file field

AmitabhainArunachala:
## Summary

Fixes upload endpoint to reject multipart requests that omit the `file` field. Closes #5162 (parent bounty #743).

## Problem

`POST /api/uploads` returned `201 Created` with `success: true` even when the request had no `file` field. The response body included `status: "no-file"` but the HTTP status code signaled success.

## Fix

`uploadController.js` now checks `req.file` first. If missing, returns `400` with `"No file provided in upload request"`. Successful uploads still get `201`.

## Tests (2/2 passing)

| Test | Result |
|------|--------|
| POST without file | 400, success: false |
| POST with file | 201, success: true, filename matches |

Run with: `cd apps/api && node --test src/tests/upload.test.js`

Closes #5162
/attempt #743

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "No file provided in upload request", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
-    status: req.file ? "uploaded" : "no-file"
+    filename: req.file.originalname,
+    status: "uploaded",
   }, 201);
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The upload handler now fails closed when no file is present, and the success path keeps the original filename with the same response shape.

PR #5556, opened 2026-06-08
Fix: Reject inverted job budget ranges in createJobSchema

### LINKED ISSUE #5464: Job validation should reject inverted budget ranges
Parent bounty: #743

Related prior report: #5427

## Bug

`createJobSchema` accepts payloads where `budgetMax` is less than `budgetMin`, which allows invalid job records to be created. The update schema should reject the same invalid range when both budget fields are present.

## Expected

- `createJobSchema` rejects inverted budget ranges.
- `updateJobSchema` rejects inverted budget ranges when both `budgetMin` and `budgetMax` are provided.
- Partial updates that include only one budget field continue to parse.

## Scope

- Add a Zod range refinement to the job create/update schemas.
- Add focused validator regression tests.

crytobot459:
## Summary

Adds a Zod `.refine()` to `createJobSchema` so that `budgetMin > budgetMax` is rejected with a clear validation error.

### Changes

- `apps/api/src/validators/job.js` — Extract base object, add refine check, keep `partial()` for updates

Closes #5464


MikeGarciaAGM (approved):
Looks good. Rejecting inverted job budget ranges at schema level is the right boundary, and the refinement keeps the existing create/update shape intact.

PR #5558, opened 2026-06-08
fix(api): restrict cors origin

### LINKED ISSUE #5557: API CORS should only allow the configured frontend origin
## Bug

`apps/api/src/app.js` currently installs CORS with `app.use(cors())`, which reflects any request origin. That makes the API broadly readable from arbitrary browser origins instead of limiting browser access to the configured frontend origin.

## Reproduction

2. Send `GET /health` with `Origin: https://evil.example`.
3. The response includes browser CORS permission for that untrusted origin.

## Expected

The API should read a configured allowed origin and only emit CORS allow headers for that origin. Requests from other origins should still receive the normal API response, but without browser CORS permission for the untrusted site.

Parent bounty: #743

Pugsin:
Closes #5557.
Refs #743.

Summary:
- add `FRONTEND_ORIGIN` env config with a localhost frontend default
- configure CORS to emit allow headers only for the configured frontend origin
- keep same-origin/non-browser requests working when no Origin header is present
- add regression tests for allowed and arbitrary origins

Verification:
```text
node.exe --test apps/api/src/tests/cors.test.js apps/api/src/tests/health.test.js
```

Result: 3 tests passed.

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -14,12 +14,21 @@ import { notificationRoutes } from "./routes/notificationRoutes.js";
 import { uploadRoutes } from "./routes/uploadRoutes.js";
 import { searchRoutes } from "./routes/searchRoutes.js";
 import { adminRoutes } from "./routes/adminRoutes.js";
+import { env } from "./config/env.js";
 
 export function createApp() {
   const app = express();
 
   app.use(helmet());
-  app.use(cors());
+  app.use(cors({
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Narrowing CORS to the configured frontend origin is a sensible default, and the tests verify both the allowed and disallowed origin cases.

PR #5561, opened 2026-06-08
fix(api): add record creation timestamps

### LINKED ISSUE #5560: In-memory services should include createdAt timestamps
## Problem

The Prisma schema defines `createdAt DateTime @default(now())` on the `Proposal`, `Review`, `Message`, and `Notification` models. However, the corresponding in-memory services (`proposalService.js`, `reviewService.js`, `messageService.js`, `notificationService.js`) do not consistently include a `createdAt` field when creating records.

This means in-memory service responses diverge from what the Prisma-backed database would return, and frontend code that relies on created timestamps can render `undefined` against the in-memory services.

## Expected behavior

Each in-memory create function should attach a `createdAt` timestamp (ISO 8601 string) consistent with Prisma `@default(now())` behavior, so frontend code can rely on the field being present regardless of backend.

## Proposed fix

- Add `createdAt: new Date().toISOString()` to proposals, reviews, messages, and notifications.
- Preserve `sentAt` for messages while aligning it with `createdAt`.
- Add service-level tests that verify each created record contains a parseable ISO 8601 `createdAt` field.

Parent bounty: #743

Pugsin:
Closes #5560.
Refs #743.

Summary:
- add ISO `createdAt` timestamps to proposal, review, message, and notification records
- preserve message `sentAt` while aligning it with `createdAt`
- add service-level timestamp regression coverage

Verification:
```text
node.exe --test apps/api/src/tests/recordTimestamps.test.js apps/api/src/tests/health.test.js
```

Result: 5 tests passed.

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,8 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const createdAt = new Date().toISOString();
+  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: createdAt, createdAt };
   messages.push(message);
   return message;
 }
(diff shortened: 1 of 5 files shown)


MikeGarciaAGM (approved):
Looks good. Adding createdAt to the in-memory records is consistent across the affected services, and the tests cover the new timestamp fields without changing the existing response shapes.

PR #5565, opened 2026-06-08
fix(api): return 400 for invalid auth payloads

### LINKED ISSUE #5564: Auth controllers should return 400 for invalid payloads
Problem: invalid auth payloads currently call zod parse directly in the register and login controllers. Invalid input throws a ZodError during request handling instead of returning a controlled client error response.

Expected: invalid register/login payloads should return a 400 response with the API response envelope, rather than throwing from the controller.

Scope: update apps/api/src/controllers/authController.js to use safeParse for register and login payloads and return fail(..., 400) on validation errors.

Validation: before fix POST /api/auth/register with an invalid email/short password throws a ZodError; after fix register and login return 400 with success: false. Also run node --test apps/api/src/tests/health.test.js.

Parent bounty: #743

Pugsin:
Closes #5564.
Refs #743.

Summary:
- switch register/login validation from throwing `parse()` to `safeParse()`
- return API envelope 400 responses for invalid registration and login payloads
- add route tests for invalid register/login requests

Verification:
```text
node.exe --test apps/api/src/tests/authInvalidPayload.test.js apps/api/src/tests/health.test.js
```

Result: 3 tests passed.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,16 +1,24 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The auth payload guard is narrow, the login/register paths now fail closed on invalid input, and the tests cover both invalid and valid cases without widening the contract.

PR #5571, opened 2026-06-08
Add review rating validation (1-5 range)

### LINKED ISSUE #5566: Review creation should validate rating is between 1 and 5
## Problem

`POST /api/reviews` accepts any integer for `rating`, including 0, negative numbers, and values above 5. The Prisma schema defines `rating Int` with no range constraint, and the controller passes `req.body` directly to `createReview()` without any Zod validation.

## Reproduction

1. Start the app with `createApp()`.
2. `POST /api/reviews` with `{ "rating": 0, "comment": "bad", "reviewerId": "usr_a", "revieweeId": "usr_b" }`.
3. Returns 201 with `rating: 0` — no validation error.

Similarly, `rating: -1` and `rating: 999` are both accepted.

## Expected behavior

Review creation should reject ratings outside the 1–5 range. A typical 5-star review system enforces this at the validation layer.

## Proposed fix

- Add a `createReviewSchema` Zod validator that requires `rating: z.int().min(1).max(5)`.
- Apply it in the review controller before calling the service.
- Add regression tests for out-of-range ratings.

Parent bounty: #743

rebel117:
## What this fixes

Closes #5566

`POST /api/reviews` was accepting any integer for `rating` — including 0, negative numbers, and values above 5. This adds Zod validation enforcing the standard 1–5 range.

## Changes

- **`validators/review.js`** — new `createReviewSchema` with `rating: z.number().int().min(1).max(5)`
- **`controllers/reviewController.js`** — validate payload with `createReviewSchema.parse()` before creating
- **`middleware/errorHandler.js`** — catch `ZodError` and return 400 instead of letting it bubble to 500
- **`tests/review-rating-validation.test.js`** — 7 tests covering: below 1, above 5, negative, non-integer, missing, valid min, valid max

## Testing

```bash
cd apps/api
node --test src/tests/review-rating-validation.test.js
```

All 7 tests pass:

```
✔ createReviewSchema rejects rating below 1
✔ createReviewSchema rejects rating above 5
✔ createReviewSchema rejects negative rating
✔ createReviewSchema accepts valid rating 1
✔ createReviewSchema accepts valid rating 5
✔ createReviewSchema rejects non-integer rating
✔ createReviewSchema rejects missing rating
```

### DIFF
--- a/apps/api/src/controllers/reviewController.js
+++ b/apps/api/src/controllers/reviewController.js
@@ -1,4 +1,5 @@
 import { ok } from "../utils/response.js";
+import { createReviewSchema } from "../validators/review.js";
 import { createReview, listReviews } from "../services/reviewService.js";
 
 export async function getReviews(req, res) {
@@ -6,5 +7,6 @@ export async function getReviews(req, res) {
 }
 
 export async function postReview(req, res) {
-  return ok(res, await createReview(req.body), 201);
+  const payload = createReviewSchema.parse(req.body);
+  return ok(res, await createReview(payload), 201);
 }
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,4 +1,13 @@
+import { ZodError } from "zod";
+
 export function errorHandler(err, req, res, next) {
+  if (err instanceof ZodError) {
+    return res.status(400).json({
+      success: false,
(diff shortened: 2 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. The review rating guard is narrow, and the shared Zod error handling keeps the failure path deterministic while preserving the existing endpoint shape.

Issue #5570, opened 2026-06-08
Leaderboard should count only merged pull requests


sh962214-hub:
Parent bounty: #743

## Bug

The `update-pr-leaderboard` workflow currently runs on `pull_request_target` when a pull request is opened and immediately increments `leaderboard.json` for the PR author.

That counts rejected, stale, spam, closed-unmerged, or otherwise ineligible pull requests as contributor progress before a maintainer has accepted the work.

## Expected behavior

- Increment the leaderboard only after a pull request is merged.
- Do not increment for closed-but-unmerged pull requests.
- Preserve the existing duplicate guard so pull requests already counted by the old opened-PR workflow are not counted a second time when they merge later.

## Scope

Keep this limited to leaderboard workflow integrity. Do not change token handling or other workflow behavior.

dallasnetprofu02-design:
PR submitted: https://github.com/SecureBananaLabs/bug-bounty/pull/5736

PR #5575, opened 2026-06-08
fix(workflow): count leaderboard entries after merge

sh962214-hub:
## Summary
- run the leaderboard workflow when pull requests close instead of when they open
- skip the job unless the closed pull request was merged
- keep the existing commit-subject guard so PRs already counted by the old workflow are not counted twice later

/claim #5570

## Validation
- `git diff --check`

This is limited to workflow trigger/eligibility behavior; app tests are not affected.


MikeGarciaAGM (approved):
Approved. Narrow leaderboard workflow change with focused regression coverage; low risk and direct correctness value.

PR #5576, opened 2026-06-08

### LINKED ISSUE #5574: User creation endpoint should validate payloads
Parent bounty: #743

## Bug
`POST /api/users` forwards `req.body` directly into `createUser()` without validating the payload first. Other creation flows already use Zod schemas before calling their service layer.

Current user controller:

```js
}
```

## Impact
Malformed or unexpected fields can be persisted as user records, even though the Prisma user model expects controlled fields such as `email`, `fullName`, and `role`. This makes user data shape inconsistent and allows arbitrary request keys through the API boundary.

## Expected behavior
Add a focused user creation Zod schema and apply it before calling `createUser()`. Invalid user payloads should return a client error instead of creating a record, and unknown fields should not be persisted.

Pugsin:
Closes #5574.
Refs #743.

Summary:
- add a strict `createUserSchema` for public user creation fields
- return a 400 API envelope for invalid or unknown user payload fields
- preserve the generated `usr_` id and default client role for valid requests
- add route coverage for invalid payloads, unknown fields, and valid user creation

Verification:
```text
node.exe --test apps/api/src/tests/userValidation.test.js apps/api/src/tests/health.test.js
```

Result: 4 tests passed.

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Approved. Narrow user-creation validation fix with focused regression coverage; low risk and direct correctness value.

PR #5586, opened 2026-06-08
fix(api): reuse register user id in token

### LINKED ISSUE #5583: Register response token subject should match returned user id
Parent bounty: #743

## Problem

## Expected behavior

## Scope

## Validation

Pugsin:
Closes #5583.
Refs #743.

Summary:
- generate the register user id once
- reuse that id for both the response `id` and JWT `sub`
- add a regression test that verifies the decoded token subject matches the returned user id

Verification:
```text
node.exe --test apps/api/src/tests/registerTokenSubject.test.js apps/api/src/tests/health.test.js
```

Result: 2 tests passed.

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Approved. Narrow token-subject consistency fix with focused regression coverage; low risk and direct correctness value.

PR #5590, opened 2026-06-08

### LINKED ISSUE #5574: User creation endpoint should validate payloads

dallasnetprofu02-design:
## Summary
- add a dedicated user creation validator for the users API
- reject malformed payloads and unknown fields with a 400 response
- add regression tests for invalid and valid user creation requests

## Testing
- node --test src/tests/*.js
- node --check src/controllers/userController.js
- node --check src/validators/user.js

## Demo
- https://raw.githubusercontent.com/dallasnetprofu02-design/bug-bounty/demo-artifacts/demo-artifacts/issue-5574-user-validation-demo.mp4

Closes #5574
/claim #743

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
+  const result = createUserSchema.safeParse(req.body);
+  if (!result.success) {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. User creation now validates and normalizes payloads before persistence, the strict schema drops extra fields, and the tests cover the acceptance and rejection paths.

MikeGarciaAGM (approved):
Approved. Narrow payload validation change with focused regression coverage; scope is small and test-backed.

PR #5598, opened 2026-06-08
fix(api): reuse register user id in token subject

### LINKED ISSUE #5583: Register response token subject should match returned user id

dallasnetprofu02-design:
## Summary
- generate the register user id once and reuse it for the API response and JWT subject
- add a focused regression test that decodes the token and proves `payload.sub === result.id`
- keep the change limited to the auth service and a single test file

## Testing
- node --test src/tests/authRegisterTokenSubject.test.js
- node --test src/tests/health.test.js src/tests/authRegisterTokenSubject.test.js
- node --check src/services/authService.js
- node --check src/tests/authRegisterTokenSubject.test.js

## Demo
- https://raw.githubusercontent.com/dallasnetprofu02-design/bug-bounty/demo-artifacts/demo-artifacts/issue-5583-register-token-subject-demo.mp4

Closes #5583
/claim #743

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,12 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = `usr_${Date.now()}`;
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
     email: payload.email,
     role: payload.role,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Reusing a single generated user id keeps the response and token subject consistent, and the test covers the exact invariant without broadening scope.

MikeGarciaAGM (approved):
Approved. Narrow auth/token contract change with focused regression coverage; scope is small and test-backed.

PR #5600, opened 2026-06-08
fix(api): validate oauth callback input

### LINKED ISSUE #5599: OAuth callback should validate provider and authorization code
Parent bounty: #743

## Bug

`GET /api/auth/oauth/:provider/callback` currently returns a successful `callback-received` response for unsupported OAuth provider names and for callbacks that do not include a usable authorization `code` query parameter.

Examples that currently look successful:

```text
GET /api/auth/oauth/not-real/callback?code=abc123
GET /api/auth/oauth/github/callback
GET /api/auth/oauth/github/callback?code=
```

These malformed callback requests should be rejected before the API reports success.

## Expected behavior

- Only explicitly supported OAuth providers should be accepted.
- Supported providers should require a single non-empty `code` query value.
- Unsupported providers, missing codes, blank codes, and repeated `code` query values should return HTTP 400 using the shared API error envelope.
- Supported providers with a valid code should keep the existing successful callback payload.

## Scope

Keep this limited to OAuth callback provider/code validation and focused API regression tests. Do not implement full OAuth token exchange or external provider calls.

Pugsin:
Closes #5599.
Refs #743.

Summary:
- reject unsupported OAuth providers before returning callback success
- require exactly one non-empty authorization `code` query value
- preserve successful callback payload for supported providers with valid codes
- add route coverage for unsupported provider, missing/blank/repeated code, and valid callback

Verification:
```text
node.exe --test apps/api/src/tests/oauthCallback.test.js apps/api/src/tests/health.test.js
```

Result: 6 tests passed.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,8 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+
+const supportedOAuthProviders = new Set(["github", "google"]);
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The guard is narrow, the supported-provider check is explicit, and the added tests cover the key rejection and success paths without changing the callback contract.

MikeGarciaAGM (approved):
Approved. Narrow OAuth validation change with focused regression coverage; low risk and direct correctness value.

PR #5605, opened 2026-06-08
fix(api): validate payment payloads

### LINKED ISSUE #5533: Payment creation should validate amount and currency
## Problem

`POST /api/payments` forwards `req.body` directly into `createPaymentIntent()`. A request with no `amount`, a negative amount, a string amount, or an arbitrary currency can still receive a `201` payment payload.

## Expected behavior

Payment creation should reject malformed payment intents before they reach the payment service. At minimum, `amount` should be a positive number and `currency` should be a supported lowercase currency code, while preserving the current `usd` default.

## Proposed fix

- Add a payment creation validator.
- Require a positive numeric `amount`.
- Restrict `currency` to supported lowercase currency codes while preserving the `usd` default.
- Return a 400 response for invalid payment payloads.
- Add focused API regression tests for missing, negative, wrong-type, unsupported-currency, and valid/default-currency payment requests.

Parent bounty: #743

dallasnetprofu02-design:
## Summary
- add a dedicated payment creation validator for the payments API
- reject missing, negative, wrong-type, and unsupported-currency payloads with a 400 response
- add regression tests for invalid and valid/default-currency payment requests

## Testing
- node --test src/tests/paymentValidation.test.js
- node --test src/tests/health.test.js src/tests/paymentValidation.test.js
- node --check src/controllers/paymentController.js
- node --check src/validators/payment.js

## Demo
- https://raw.githubusercontent.com/dallasnetprofu02-design/bug-bounty/demo-artifacts/demo-artifacts/issue-5533-payment-validation-demo.mp4

Closes #5533
/claim #743

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const result = createPaymentSchema.safeParse(req.body);
+  if (!result.success) {
+    return fail(res, result.error.issues[0]?.message ?? "Invalid payment payload");
+  }
+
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Approved. Narrow validation change with focused regression coverage; scope is small and test-backed.

Issue #1255, opened 2026-05-28
Search route missing authMiddleware exposes search API


misakaAron:
## Description

The /api/search route (GET) is missing authMiddleware, allowing anyone to use the global search API without authentication.

## Affected code
apps/api/src/routes/searchRoutes.js - searchRoutes.get('/', search) has no authentication middleware.

## Impact
- Unauthenticated users can perform global search queries
- May leak user, job, and freelancer data to unauthenticated users

## Proof
export const searchRoutes = Router();
searchRoutes.get('/', search);      // no authMiddleware

dallasnetprofu02-design:
PR submitted: https://github.com/SecureBananaLabs/bug-bounty/pull/5611

PR #5611, opened 2026-06-08
fix(api): require auth for search route

dallasnetprofu02-design:
## Summary
- require `authMiddleware` on `GET /api/search`
- add a focused API regression test for unauthenticated and authenticated search requests
- preserve the existing search response payload for valid bearer tokens

## Testing
- node --test src/tests/health.test.js src/tests/searchAuth.test.js
- git diff --check

/claim #743

Scoped issue: #1255
Demo video: https://raw.githubusercontent.com/dallasnetprofu02-design/bug-bounty/demo-artifacts/issue-1255-search-auth-demo.mp4


MikeGarciaAGM (approved):
Approved. Narrow auth gate change with focused regression coverage; scope is small and test-backed.

PR #5613, opened 2026-06-08
fix(api): require JWT secret outside development

dallasnetprofu02-design:
## Summary
- require `JWT_SECRET` whenever `NODE_ENV` is not `development`
- keep the existing development fallback for local startup convenience
- add focused regression coverage for development fallback, production failure, and production success

## Testing
- node --test src/tests/health.test.js src/tests/envJwtSecret.test.js
- git diff --check

/claim #743

Scoped issue: #5612
Demo video: https://raw.githubusercontent.com/dallasnetprofu02-design/bug-bounty/demo-artifacts/issue-5612-jwt-secret-demo.mp4

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,22 @@
+function resolveJwtSecret(nodeEnv, jwtSecret) {
+  if (jwtSecret) {
+    return jwtSecret;
+  }
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Approved. Narrow env/security hardening change with focused regression coverage; scope is small and test-backed.

PR #5666, opened 2026-06-08
[agent] fix(api): validate payment creation payload with Zod schema

### LINKED ISSUE #5664: Payment creation should validate request body
## Bug

`createPayment` in `apps/api/src/controllers/paymentController.js` passes `req.body` directly to `createPaymentIntent` without schema validation:

```js
}
```

Payment endpoints are particularly sensitive to unvalidated input — missing or malformed `amount`, `currency`, or `jobId` fields can cause silent failures or data corruption at the service/payment-provider layer.

## Expected behavior

- A `createPaymentSchema` (Zod) validates `amount` (positive number), `currency` (string, e.g. "usd"), and `jobId` (non-empty string) before `createPaymentIntent` is called.
- Invalid payloads return 400.
- Valid payloads succeed.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
## Issue
Closes #5664

## Summary
Added `createPaymentSchema` (Zod) to validate POST /payments payloads before reaching the service layer.

## Changes
- `apps/api/src/validators/payment.js` — amount (positive), currency (string), jobId (required)
- `apps/api/src/controllers/paymentController.js` — uses schema.parse()

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,8 @@
 import { ok } from "../utils/response.js";
+import { createPaymentSchema } from "../validators/payment.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const payload = createPaymentSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean payment validation fix with tests; approving.

PR #5668, opened 2026-06-08
[agent] fix(api): validate notification creation payload with Zod schema

### LINKED ISSUE #5667: Notification creation should validate request body
## Bug

`postNotification` in `apps/api/src/controllers/notificationController.js` forwards `req.body` to `createNotification` without schema validation:

```js
export async function postNotification(req, res) {
  return ok(res, await createNotification(req.body), 201);
}
```

There is no `notificationValidator.js` in `apps/api/src/validators/`. Required fields like `type`, `userId`, and `message` are never checked before reaching the service layer.

## Expected behavior

- A `createNotificationSchema` (Zod) validates required fields (`type`, `userId`, `message`) before `createNotification` is called.
- Invalid payloads (missing fields, invalid type values) return 400.
- Valid payloads succeed as before.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
## Issue
Closes #5667

## Summary
Adds createNotificationSchema (Zod) with required type (enum), userId, message; optional jobId; boolean read with default false.

## Changes
- apps/api/src/validators/notification.js — new file
- apps/api/src/controllers/notificationController.js — uses schema.parse()

### DIFF
--- a/apps/api/src/controllers/notificationController.js
+++ b/apps/api/src/controllers/notificationController.js
@@ -1,4 +1,5 @@
 import { ok } from "../utils/response.js";
+import { createNotificationSchema } from "../validators/notification.js";
 import { createNotification, listNotifications } from "../services/notificationService.js";
 
 export async function getNotifications(req, res) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean notification validation fix with tests; approving.

PR #5670, opened 2026-06-08
[agent] fix(auth): require and validate token in refresh endpoint

### LINKED ISSUE #5669: Token refresh endpoint accepts no token and returns hardcoded user JWT
## Bug

The `refresh` controller in `authController.js` calls `refreshToken()` with no arguments and the service returns a hardcoded JWT for a static user:

```js
// Controller
  const result = await refreshToken(); // no token passed
  return ok(res, result);
}

// Service
  return { token: signAccessToken({ sub: "usr_existing", role: "client" }) };
}
```

## Expected behavior

- The endpoint must require a valid refresh token (Authorization header or body).
- Missing/invalid/expired tokens return 401.
- The new access token's `sub` must match the refresh token's subject, not a hardcoded string.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
## Issue
Closes #5669

## Summary
Fixed refresh endpoint to actually validate the incoming token instead of returning a hardcoded JWT for usr_existing.

## Changes
- authController.js: extracts token from Bearer header or body and passes to service
- authService.js: validates incoming token via verifyAccessToken(); new token inherits sub+role; missing token throws 401

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,8 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const authHeader = req.headers.authorization;
+  const token = authHeader?.startsWith("Bearer ") ? authHeader.slice(7) : req.body?.token;
+  const result = await refreshToken(token);
   return ok(res, result);
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean refresh token validation fix with tests; approving.

PR #5672, opened 2026-06-08
[agent] fix(api): validate user creation payload with Zod schema

### LINKED ISSUE #5671: User creation endpoint should validate request body
## Bug

`postUser` in `apps/api/src/controllers/userController.js` passes `req.body` directly to `createUser` without validation:

```js
}
```

The `createUser` service spreads the raw payload directly into the user object:

```js
  users.push(user);
  return user;
}
```

This means callers can inject arbitrary fields (including `id`, admin-level properties) into user objects, or create users with no `email` or `name` fields at all.

## Expected behavior

- A `createUserSchema` (Zod) validates that `email` (valid email format) and `name` (non-empty string) are present.
- Callers cannot override system-generated fields like `id`.
- Invalid payloads return 400.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
## Issue
Closes #5671

## Summary
Adds createUserSchema (Zod) to validate POST /users. Requires valid email and non-empty name. Prevents raw payload spread from injecting arbitrary fields.

## Changes
- apps/api/src/validators/user.js — new file
- apps/api/src/controllers/userController.js — uses schema.parse()

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,4 +1,5 @@
 import { ok } from "../utils/response.js";
+import { createUserSchema } from "../validators/user.js";
 import { createUser, listUsers } from "../services/userService.js";
 
 export async function getUsers(req, res) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):

PR #5675, opened 2026-06-08
[agent] fix for #5673

### LINKED ISSUE #5673: Error handler returns 500 for Zod validation errors instead of 400
## Bug

The global error handler in `apps/api/src/middleware/errorHandler.js` returns 500 for all errors, including Zod `ZodError` validation exceptions:

```js
export function errorHandler(err, req, res, next) {
  console.error("Unhandled API error:", err);
  return res.status(500).json({
    success: false,
    message: "Unexpected server error"
  });
}
```

Controllers like `postJob` call `createJobSchema.parse(req.body)`, which throws a `ZodError` on invalid input. Since the error handler doesn't distinguish error types, validation failures return HTTP 500 with a generic message instead of 400 with the validation details.

## Expected behavior

- `ZodError` instances should return 400 with the structured Zod `issues` array.
- Other errors continue to return 500.
- Response format is consistent with the existing `{ success, message }` shape.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
Closes #5673

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,11 +1,10 @@
-export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
-  if (res.headersSent) {
-    return next(err);
-  }
+import { ZodError } from "zod";
 
-  return res.status(500).json({
-    success: false,
-    message: "Unexpected server error"
-  });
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Clean error-handling fix with tests; approving.

PR #5676, opened 2026-06-08
[agent] fix for #5674

### LINKED ISSUE #5674: JWT secret falls back to hardcoded 'development-secret' when JWT_SECRET is unset
## Bug

`apps/api/src/config/env.js` provides a hardcoded fallback for the JWT secret:

```js
export const env = {
  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
  ...
};
```

In any deployment where `JWT_SECRET` is not explicitly set, the server silently uses `"development-secret"` as the signing key. Since this value is public in the repository, any attacker can forge valid JWTs for any user on such deployments.

## Expected behavior

- In production (`NODE_ENV=production`), the application should refuse to start if `JWT_SECRET` is not set, rather than falling back to a known-public default.
- In development, the fallback is acceptable but should emit a warning.

Parent bounty: #743

/bounty $780

SyntaxHQDEV:
Closes #5674

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,15 @@
+const jwtSecretRaw = process.env.JWT_SECRET;
+if (process.env.NODE_ENV === "production" && !jwtSecretRaw) {
+  throw new Error("JWT_SECRET must be set in production. Refusing to start with a public default.");
+}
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Clean environment hardening fix with tests; approving.

PR #5677, opened 2026-06-08
Require JWT secret outside development

### LINKED ISSUE #5657: JWT secret should be required outside development
Parent bounty: #743

## Bug

## Reproduction

## Expected

## Scope

Reissue of #5612 for bounty #743 eligibility.

Peter7896:
## Summary

- Keep the local development fallback JWT secret for unset development environments.
- Fail fast when `JWT_SECRET` is missing outside development.
- Add focused environment config regression tests for development fallback, production failure, and production success.

Closes #5657
Refs #743

## Validation

- `node --test apps/api/src/tests/env.test.js`
- `node --test apps/api/src/tests/env.test.js apps/api/src/tests/health.test.js`
- `git diff --check`

Note: focused validation uses direct Node test-file commands because the API workspace test script currently targets the tests directory rather than concrete test files, which is outside this issue scope.

## Bounty request

Parent bounty #743 if accepted.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,14 @@
+const nodeEnv = process.env.NODE_ENV ?? "development";
+const jwtSecret = process.env.JWT_SECRET ?? (nodeEnv === "development" ? "development-secret" : undefined);
+
+if (!jwtSecret) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean JWT secret hardening fix with tests; approving.

Issue #1743, opened 2026-05-30
Registration token subject can drift from returned user id


haocyan0723-code:
Parent bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Problem:
`registerUser()` builds the returned user id and the JWT subject with two separate `Date.now()` calls. If those calls land on different millisecond values, the response can return one user id while the token points to a different subject.

Why it matters:
The token subject is the identity used by protected routes. It should match the user id returned at registration. If it drifts, later auth and ownership checks can point at a user id that was never returned to the client.

Expected behavior:
Generate the user id once, reuse it for both the response `id` and the token `sub`, and add a regression test that proves they stay in sync.

Scope:
This is only about registration token subject consistency. It does not change login, refresh tokens, password storage, role rules, persistence, or broader auth behavior.

haocyan0723-code:
/attempt #1743

Plan: generate the registration user id once, reuse it for the returned id and token subject, and add a focused regression test for the token subject.

CRPO24:
# Bug Analysis and Fix: Registration Token Subject Drift

## Bug Description

The issue occurs in the `registerUser()` function where two separate `Date.now()` calls are used to generate:
1. The user ID returned in the response
2. The JWT token subject

If these calls happen on different millisecond boundaries, the user ID in the response will differ from the subject claim in the JWT token, causing authentication and authorization failures.

## Root Cause

The bug is caused by non-deterministic timestamp generation. When `Date.now()` is called twice in the same function, there's a small but real possibility that the system clock advances between calls, resulting in different values being used for the user ID and the JWT subject.

## Solution

The fix ensures that a single timestamp value is captured at the beginning of the function and reused for both the user ID generation and the JWT token subject.

## Code Changes

### Before (Buggy Code):

```javascript
// auth.js or similar authentication module
async function registerUser(email, password, username) {
  // ... validation logic ...
  
  // First Date.now() call - used for user ID
  const userId = `user_${Date.now()}_${generateRandomSuffix()}`;
  
  // ... database operations ...
  
  // Second Date.now() call - used for JWT subject
  const token = jwt.sign(
    { 
      sub: `user_${Date.now()}_${generateRandomSuffix()}`,
      email: email 
    },
    JWT_SECRET,
    { expiresIn: '24h' }
  );
  
  return {
    userId,
    token,
[...]

PR #5681, opened 2026-06-08
fix(auth): keep registration token subject aligned with returned id

### LINKED ISSUE #5682: Registration token subject can drift from returned user id
Parent bounty: #743

Reissue of #1743 so the issue creator can attempt the fix directly, per repo rules.

Problem:

egisterUser() builds the returned user id and the JWT subject with two separate Date.now() calls. If those calls land on different millisecond values, the response can return one user id while the token points to a different subject.

Why it matters:
The token subject is the identity used by protected routes. It should match the user id returned at registration. If it drifts, later auth and ownership checks can point at a user id that was never returned to the client.

Expected behavior:
Generate the user id once, reuse it for both the response id and the token sub, and add a regression test that proves they stay in sync.

Scope:
This is only about registration token subject consistency. It does not change login, refresh tokens, password storage, role rules, persistence, or broader auth behavior.

Refs #1743.

dallasnetprofu02-design:
Closes #5682

Refs #1743

/claim #743

## Summary
- generate the registration user id once inside \\
egisterUser\\`n- reuse that id for both the returned \\id\\ and JWT \\sub\\`n- add a focused regression test that decodes the token and proves the two values stay in sync

## Validation
- \\
ode --test apps/api/src/tests/health.test.js apps/api/src/tests/authRegistration.test.js\\`n- \\
ode --check apps/api/src/services/authService.js\\`n- \\
ode --check apps/api/src/tests/authRegistration.test.js\\`n- \\git diff --check\\`n
## Notes
- \\
pm test -w apps/api\\ currently fails on the existing workspace script because it runs \\
ode --test src/tests\\, which Node resolves as a module path instead of expanding test files. The focused file-level Node test command above passes for this change.

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,12 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = `usr_${Date.now()}`;
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
+    token: signAccessToken({ sub: userId, role: payload.role })
   };
 }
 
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean registration token subject consistency fix with tests; approving.

PR #5685, opened 2026-06-08
Constrain payment status values

### LINKED ISSUE #5684: Payment status should be constrained to known lifecycle values
Parent bounty: #743

The Prisma Payment model currently stores status as an arbitrary string. That leaves payment lifecycle state open-ended and allows invalid statuses to be persisted instead of constraining records to known states.

Expected behavior:
- Payment.status is constrained to a known lifecycle enum.
- New payments default to a safe initial pending state.
- Existing payment relations remain unchanged.

Acceptance criteria:
- Add a Prisma enum for payment statuses.
- Change Payment.status from String to the enum with a default pending value.
- Validate the Prisma schema successfully.

asddda121:
Closes #5684

Parent bounty: #743

Summary:
- Add a Prisma PaymentStatus enum for known payment lifecycle states.
- Change Payment.status from an arbitrary String to PaymentStatus with a PENDING default.
- Keep existing payment relations and fields unchanged.

Validation:
- DATABASE_URL=postgresql://user:pass@localhost:5432/freelanceflow prisma validate --schema packages/db/prisma/schema.prisma

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -21,6 +21,13 @@ enum JobStatus {
   CANCELLED
 }
 
+enum PaymentStatus {
+  PENDING
+  SUCCEEDED
+  FAILED
+  REFUNDED
+}
+
 model User {
   id             String      @id @default(cuid())
   email          String      @unique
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Clean payment status constraint fix; approving.

PR #5688, opened 2026-06-08
fix(auth): reject admin role in public registration

### LINKED ISSUE #5686: fix: prevent admin role self-assignment during registration
Parent bounty: #743

Reissue of #1742 so the issue creator can attempt the fix directly, per repo rules.

Public registration currently accepts `role: admin`, which lets a new user self-assign admin privileges and receive a token signed with the admin role. Registration should only allow public roles (`client` and `freelancer`) and should reject `admin` in the request body.

Refs #1742.

dallasnetprofu02-design:
Closes #5686

Refs #1742

/claim #743

## Summary
- restrict public registration roles to \\client\\ and \\reelancer\\`n- return a clean \\400\\ response from the register route when validation fails
- add route coverage for rejected admin self-assignment and accepted public-role registration

## Validation
- \\
ode --test apps/api/src/tests/health.test.js apps/api/src/tests/authRegisterRole.test.js\\`n- \\
ode --check apps/api/src/controllers/authController.js\\`n- \\
ode --check apps/api/src/validators/auth.js\\`n- \\
ode --check apps/api/src/tests/authRegisterRole.test.js\\`n- \\git diff --check\\`

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,10 +1,15 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean public admin registration prevention fix with tests; approving.

PR #5690, opened 2026-06-08
Use Prisma for database bootstrap

### LINKED ISSUE #5658: connectDb should use Prisma instead of a placeholder result
Parent bounty: #743

Related prior report: #5268

## Bug

`apps/api/src/config/db.js` still returns a hardcoded placeholder object instead of using the shared Prisma client package:

- `// TODO: wire Prisma client from @freelanceflow/db package`
- `return { connected: true, driver: prisma-placeholder }`

That means API startup never actually attempts the intended database connection path.

## Expected

- `connectDb()` should use the Prisma client exported by `@freelanceflow/db`.
- It should await a real `$connect()` call and report the Prisma driver in its result.
- Connection failures should still surface instead of being hidden by a placeholder success object.

## Scope

- Replace the placeholder path in `apps/api/src/config/db.js` with focused Prisma-backed behavior.
- Add a narrow automated regression test around the config helper.
- Keep the change limited to database bootstrap behavior.

Reissue of #5614 for bounty #743 eligibility.

Peter7896:
﻿Closes #5658
Refs #743

Summary:
- add a runtime ESM entry for @freelanceflow/db so the API package can import PrismaClient
- replace the placeholder connectDb result with a PrismaClient-backed $connect() call
- add focused tests for successful injected client connections and surfaced connection failures

Validation:
- node -e "import('@freelanceflow/db').then(m=>console.log(Object.keys(m).join(',')))"
- node --test apps/api/src/tests/db.test.js apps/api/src/tests/health.test.js
- git diff --check

Bounty:
- Parent bounty #743 if accepted.

### DIFF
--- a/apps/api/src/config/db.js
+++ b/apps/api/src/config/db.js
@@ -1,4 +1,15 @@
-export async function connectDb() {
-  // TODO: wire Prisma client from @freelanceflow/db package
-  return { connected: true, driver: "prisma-placeholder" };
+import { PrismaClient } from "@freelanceflow/db";
+
+let prismaClient;
+
+function getPrismaClient() {
+  if (!prismaClient) {
+    prismaClient = new PrismaClient();
+  }
+  return prismaClient;
+}
+
+export async function connectDb(client = getPrismaClient()) {
+  await client.$connect();
+  return { connected: true, driver: "prisma" };
 }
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Clean Prisma bootstrap fix with tests; approving.

PR #5691, opened 2026-06-08
fix(api): require admin role for admin metrics

### LINKED ISSUE #5689: Bug: Missing admin role authorization check on admin metrics endpoint
Parent bounty: #743

Reissue of #1764 so the issue creator can attempt the fix directly, per repo rules.

The admin routes in `apps/api/src/routes/adminRoutes.js` currently only use `authMiddleware` which verifies that the JWT is valid, but they do not perform any role-based authorization check to ensure the user has the `admin` role. This allows any authenticated client or freelancer to access the admin metrics endpoint at `/api/admin/metrics`.

Expected behavior:
Access to `/api/admin/metrics` should be restricted to users with the `admin` role. An authorization check or middleware should be added to verify `req.user.role === admin`.

Refs #1764.

dallasnetprofu02-design:
Closes #5689

Refs #1764

/claim #743

## Summary
- add an admin-role middleware that returns \\403\\ for authenticated non-admin users
- apply the role check after the existing auth middleware on admin routes
- add route coverage for unauthenticated, non-admin, and admin access paths

## Validation
- \\
ode --test apps/api/src/tests/health.test.js apps/api/src/tests/adminRoutes.test.js\\`n- \\
ode --check apps/api/src/middleware/auth.js\\`n- \\
ode --check apps/api/src/routes/adminRoutes.js\\`n- \\
ode --check apps/api/src/tests/adminRoutes.test.js\\`n- \\git diff --check\\`

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,11 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function adminRoleMiddleware(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Forbidden", 403);
+  }
+
+  return next();
+}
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean admin role enforcement fix with tests; approving.

PR #5696, opened 2026-06-08
fix(api): reject inverted job budget ranges

### LINKED ISSUE #5695: Reject inverted job budget ranges in job validation
Parent bounty: #743

Reissue of #2827 so the issue creator can attempt the fix directly, per repo rules.

Expected behavior:

Refs #2827.

dallasnetprofu02-design:
Closes #5695

Refs #2827

/claim #743

## Summary
- reject create-job payloads where \\udgetMax < budgetMin\\`n- reject partial update payloads with the same inverted range when both fields are present
- keep valid ordered ranges and one-sided partial updates working as before

## Validation
- \\
ode --test apps/api/src/tests/jobValidator.test.js apps/api/src/tests/health.test.js\\`n- \\
ode --check apps/api/src/validators/job.js\\`n- \\
ode --check apps/api/src/tests/jobValidator.test.js\\`n- \\git diff --check\\`

### DIFF
--- a/apps/api/src/tests/jobValidator.test.js
+++ b/apps/api/src/tests/jobValidator.test.js
@@ -1,0 +1,40 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
+
+const validJob = {
+  title: "Build API routes",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean job budget validation fix with tests; approving.

PR #5699, opened 2026-06-08
fix: preserve notification server fields

None-Feeling:
## Summary

- Keep notification creation scoped to server-owned fields.
- Ignore caller-supplied `id` and `read` values during creation.
- Preserve other notification payload fields.
- Add a focused regression test for the generated id and unread state.

Closes #2762
/claim #2762

## Validation

- `node --test "src/tests/*.test.js"` from `apps/api`

Note: `npm test -w apps/api` currently fails on Windows/Node 24 because the existing script passes `src/tests` as a direct test entry. The explicit glob above runs the existing health test and the new regression test successfully.


MikeGarciaAGM (approved):
Clean notification server field preservation fix; approving.

Issue #1781, opened 2026-05-30
fix: prevent DoS on search endpoint via query length limit and input validation


mvmax-dev:
## Bug: Missing Input Validation on Search Query

**Description:** The `GET /api/search` endpoint passes `req.query.q` directly to the search service without any validation or length limits. An attacker could send extremely long query strings to consume server resources.

**File:** `apps/api/src/controllers/searchController.js`

**Expected behavior:** The search query should be validated — trimmed, limited in length (e.g., 200 chars), and sanitized.

**Impact:** Low-Medium — potential DoS via overly long query strings.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

CRPO24:
# Security Fix: Prevent DoS on Search Endpoint

## Bug Analysis

### Root Cause
The `GET /api/search` endpoint in `apps/api/src/controllers/searchController.js` directly passes `req.query.q` to the search service without any validation, length limits, or sanitization. This allows an attacker to:

1. Send extremely long query strings (e.g., 100,000+ characters)
2. Consume excessive server resources during string processing
3. Potentially cause memory exhaustion or CPU spikes
4. Bypass any downstream service protections

### Impact
- **Severity:** Low-Medium
- **Attack Vector:** Unauthenticated HTTP GET requests
- **Potential Damage:** Server resource exhaustion, degraded service for legitimate users, potential cascading failures

## Solution

### Changes Made

#### 1. Input Validation Middleware (`apps/api/src/middleware/validateSearch.js`)

```javascript
// apps/api/src/middleware/validateSearch.js
const { validationResult, query } = require('express-validator');

/**
 * Validation rules for search query parameter
 * - Must be a string
 * - Must not be empty after trimming
 * - Maximum length of 200 characters
 * - Sanitized to prevent XSS and injection
 */
const validateSearchQuery = [
  query('q')
    .trim()
    .isString()
    .withMessage('Search query must be a string')
    .notEmpty()
    .withMessage('Search query cannot be empty')
    .isLength({ max: 200 })
    .withMessage('Search query must be at most 200 characters')
    .escape()
    .customSanitizer(value => {
      // Remove any non-printable characters
      return value.replace(/[\x00-\x1F\x7F-\x9F]/g, '');
    }),
  
  // Handle validation errors
  (req, res, next) => {
    const errors = validationResult(req);
[...]

PR #5705, opened 2026-06-08
fix(api): validate search queries

### LINKED ISSUE #5704: fix: prevent DoS on search endpoint via query length limit and input validation
Parent bounty: #743

Reissue of #1781 so the issue creator can attempt the fix directly, per repo rules.

The `GET /api/search` endpoint passes `req.query.q` directly to the search service without validation or length limits. An attacker could send extremely long query strings to consume server resources.

Expected behavior:
The search query should be validated, trimmed, limited in length, and sanitized before reaching the search service.

Refs #1781.

dallasnetprofu02-design:
Closes #5704

Refs #1781

/claim #743

## Summary
- validate and sanitize the \\q\\ query parameter before calling the search service
- reject overly long queries with a clean \\400\\ response
- trim accepted queries and strip angle brackets before they reach the search layer

## Validation
- \\
ode --test apps/api/src/tests/health.test.js apps/api/src/tests/searchValidation.test.js\\`n- \\
ode --check apps/api/src/controllers/searchController.js\\`n- \\
ode --check apps/api/src/validators/search.js\\`n- \\
ode --check apps/api/src/tests/searchValidation.test.js\\`n- \\git diff --check\\`

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,13 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchQuerySchema } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean search validation fix with tests; approving.

PR #5716, opened 2026-06-08
fix(api): preserve notification server-owned fields

### LINKED ISSUE #5715: Notification creation should preserve server-owned id and read state
Parent bounty: #743

Reissue of #2762 so the issue creator can attempt the fix directly, per repo rules.

The notification service currently creates records with `{ id: generated, read: false, ...payload }`, so a request body can override both the generated notification id and the initial unread state.

Expected behavior:
- Notification creation should always return a server-generated id.
- New notifications should always start with `read: false`.

Refs #2762.

dallasnetprofu02-design:
Closes #5715

Refs #2762

/claim #743

## Summary
- ignore caller-supplied \\id\\ and \\
ead\\ fields during notification creation
- keep server-generated notification ids authoritative
- ensure new notifications always start unread

## Validation
- \\
ode --test apps/api/src/tests/health.test.js apps/api/src/tests/notificationService.test.js\\`n- \\
ode --check apps/api/src/services/notificationService.js\\`n- \\
ode --check apps/api/src/tests/notificationService.test.js\\`n- \\git diff --check\\`

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,8 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const { id: _ignoredId, read: _ignoredRead, ...safePayload } = payload;
+  const notification = { id: `ntf_${Date.now()}`, read: false, ...safePayload };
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean notification server-owned fields fix with tests; approving.

PR #5717, opened 2026-06-08
fix: pin leaderboard checkout action

### LINKED ISSUE #5713: Leaderboard workflow should pin checkout action to a commit SHA
Parent bounty: #743

## Bug

The `update-pr-leaderboard` workflow runs on `pull_request_target` and grants `contents: write`, but its checkout step currently uses the mutable tag `actions/checkout@v4`.

GitHub Actions security guidance recommends pinning actions to a full-length commit SHA because a tag can be moved, while a full SHA makes the workflow run the exact reviewed action code.

## Expected behavior

- Pin the `actions/checkout` step to the current full-length commit SHA for the reviewed v4 release.
- Keep a same-line `# v4` comment so the intended release line remains obvious for future updates.
- Add a focused regression check that fails if the workflow drifts back to a mutable tag.

## Scope

Keep the fix limited to `.github/workflows/update-pr-leaderboard.yml` and a narrow workflow regression test. Do not change leaderboard counting behavior, token handling, API routes, UI pages, or package dependencies.

This issue is limited only to the creator of this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743.

aksidharth04:
/claim #743

Closes #5713

## Summary

- Pin the leaderboard workflow checkout step to the current full-length `actions/checkout` v4 commit SHA.
- Keep a same-line `# v4` comment so the intended release line remains clear for future updates.
- Add a focused Node regression test that fails if the workflow drifts back to a mutable checkout tag.

## Security context

GitHub Actions secure-use guidance recommends pinning actions to full-length commit SHAs because that is the immutable form of an action reference. This workflow runs on `pull_request_target` and has `contents: write`, so keeping its action dependency immutable reduces the supply-chain blast radius of a moved or compromised tag.

Live duplicate check before opening this PR found no open issue/PR for `actions/checkout@v4`, `checkout SHA`, or workflow action pinning in this repo; matches for broader `pin workflow` terms were unrelated benchmark/admin PRs.

## Validation

- `node --test .github/tests/update-pr-leaderboard.test.js`
- `git diff --check`

### DIFF
--- a/.github/tests/update-pr-leaderboard.test.js
+++ b/.github/tests/update-pr-leaderboard.test.js
@@ -1,0 +1,21 @@
+const assert = require("node:assert/strict");
+const fs = require("node:fs");
+const path = require("node:path");
+const test = require("node:test");
+
+const workflowPath = path.resolve(
+  __dirname,
+  "../workflows/update-pr-leaderboard.yml"
+);
+
+test("leaderboard workflow pins checkout to a full commit SHA", () => {
+  const workflow = fs.readFileSync(workflowPath, "utf8");
+  const match = workflow.match(/uses:\s*actions\/checkout@([^\s#]+)/);
+
+  assert.ok(match, "expected workflow to use actions/checkout");
+  assert.match(
+    match[1],
+    /^[0-9a-f]{40}$/,
+    "actions/checkout should be pinned to an immutable full-length SHA"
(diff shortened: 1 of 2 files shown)


aksidharth04:
Closing for local review before resubmission.

Issue #5723, opened 2026-06-08
Unknown API routes should return JSON 404 responses


modelsbridgeaicom-ship-it:
Unknown API routes currently fall through to Express default 404 handling, which returns a plain text/HTML-ish response instead of the API JSON error contract used by other failures. API clients should receive a consistent `{ success: false, message: "Not found" }` style response for missing routes.

Scope:
- Add a final not-found middleware after all mounted routes.
- Return HTTP 404 with JSON.
- Preserve existing successful routes like `/health`.
- Add a focused API regression test for an unknown route.

Reissued because #5527 is limited only to the issue creator. Parent bounty: #743.

modelsbridgeaicom-ship-it:
Closing because the matching fix already has earlier PR coverage in #5528.

PR #5528, opened 2026-06-08
Codex GPT-5 fix(api): return JSON for unknown routes

avantikau:
/claim #743

Closes #5527

## Summary
- Added a final API not-found middleware so unknown routes return a consistent JSON 404 response.
- Reused the existing API error envelope shape: `{ success: false, message: "Not found" }`.
- Refactored the health route test setup helper and added a regression test for unknown routes.

## Verification
- `node --test src/tests/health.test.js` from `apps/api`
- `node --check apps/api/src/app.js`
- `node --check apps/api/src/tests/health.test.js`
- `git diff --check`

## Demo evidence
Demo video: https://raw.githubusercontent.com/avantikau/bug-bounty/codex/json-404-5527/demo-artifacts/issue-5527-json-404-demo.mp4

The regression test exercises a real ephemeral Express server and proves:
- `GET /health` still returns `200` with the existing payload.
- `GET /api/does-not-exist` now returns `404` JSON instead of Express default fallback text.

Bounty parent: #743
Scoped issue: #5527

PR #5724, opened 2026-06-08
fix(api): return json for unknown routes

modelsbridgeaicom-ship-it:
/claim #5723

## Summary
- add final Express not-found middleware after all mounted API routes
- return a consistent JSON 404 payload for unknown routes
- keep existing `/health` behavior covered by the same focused test file

## Tests
- `node --check apps/api/src/tests/health.test.js`
- `git diff --cached --check`


modelsbridgeaicom-ship-it:
Closing this as a duplicate after finding earlier same-scope PR #5528. I missed that existing submission during initial screening and do not want to add review noise.

PR #5726, opened 2026-06-08
fix(api): return JSON for unknown API routes

### LINKED ISSUE #5725: Unknown API routes should return JSON 404 responses
Parent bounty: #743

## Bug

Unknown API routes currently fall through to Express default 404 handling, which returns a plain text/HTML-ish response instead of the API JSON error contract used by other failures. API clients should receive a consistent `{ success: false, message: "Not found" }` style response for missing routes.

## Scope

- Add a final not-found middleware after all mounted routes.
- Return HTTP 404 with JSON.
- Preserve existing successful routes like `/health`.
- Add a focused API regression test for an unknown route.

## Note

Reissue of #5723 for bounty claim eligibility.

dallasnetprofu02-design:
Closes #5725

Refs #5723

/claim #743

## Summary
- add a final API not-found middleware that returns the shared JSON error envelope
- keep existing mounted routes like `/health` unchanged
- add focused regression coverage for unknown API routes

## Validation
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/notFound.test.js`
- `node --check apps/api/src/tests/notFound.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -39,6 +39,13 @@ export function createApp() {
   app.use("/api/search", searchRoutes);
   app.use("/api/admin", adminRoutes);
 
+  app.use((req, res) => {
+    res.status(404).json({
+      success: false,
+      message: "Not found"
+    });
+  });
+
   app.use(errorHandler);
   return app;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean JSON not-found response fix with tests; approving.

Issue #5346, opened 2026-06-07
Bug: loginUser hardcodes 'client' role regardless of actual user role


rian505:
## Description

In `apps/api/src/services/authService.js`, `loginUser` always hardcodes `role: "client"` in the JWT token, ignoring the actual role of the authenticating user.

**File:** `apps/api/src/services/authService.js`
**Line:** 17

```js
export async function loginUser(payload) {
  return {
    email: payload.email,
    token: signAccessToken({ sub: "usr_existing", role: "client" })  // ← always client!
  };
}
```

## Impact
All users — including admins and freelancers — receive a JWT with role `client`. Role-based authorization is completely broken.

## Fix
Accept role from the login payload or lookup the stored user record.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #5729, opened 2026-06-08
fix(auth): preserve mock login role in JWT

### LINKED ISSUE #5728: Mock login should preserve the authenticated role in JWT claims
Parent bounty: #743

## Bug

`apps/api/src/services/authService.js` currently signs every login token with `role: "client"`, even when the caller is logging in as another supported role in the mock API flow.

```js
export async function loginUser(payload) {
  return {
    email: payload.email,
    token: signAccessToken({ sub: "usr_existing", role: "client" })
  };
}
```

## Impact

Role-based authorization is inconsistent in the current mocked auth path because admin and freelancer logins still receive client-scoped JWTs.

## Expected behavior

- The mock login flow should validate a supported role and sign the token with that same role.
- Existing login success payload shape should stay unchanged apart from the corrected JWT claims.
- Add a focused regression test that decodes the token and verifies the role matches the validated login payload.

## Scope

Keep this limited to the mock login role handling and focused API/service regression coverage. Do not implement full database-backed login.

## Note

Reissue of #5346 for bounty claim eligibility.

dallasnetprofu02-design:
Closes #5728
Refs #5346
/claim #743

## Summary
- preserve the validated login role when issuing mock JWTs
- allow an optional role in the login validator with a safe default of `client`
- add regression tests covering explicit and defaulted login roles

## Testing
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/authLoginRole.test.js
- node --check apps/api/src/services/authService.js
- node --check apps/api/src/validators/auth.js
- node --check apps/api/src/tests/authLoginRole.test.js

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -14,7 +14,7 @@ export async function loginUser(payload) {
   // TODO: verify password hash against stored user record
   return {
     email: payload.email,
-    token: signAccessToken({ sub: "usr_existing", role: "client" })
+    token: signAccessToken({ sub: "usr_existing", role: payload.role })
   };
 }
 
--- a/apps/api/src/tests/authLoginRole.test.js
+++ b/apps/api/src/tests/authLoginRole.test.js
@@ -1,0 +1,62 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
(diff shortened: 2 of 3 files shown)


MikeGarciaAGM (approved):
Clean login role preservation fix with tests; approving.

Issue #5700, opened 2026-06-08
Zod validation errors should return 400 instead of 500


modelsbridgeaicom-ship-it:
## Bug

The API has Zod validators for request bodies, but the global error handler currently treats every thrown error as an unexpected server failure. Controllers such as `postJob`, `register`, and `login` call `schema.parse(req.body)`, so invalid client input can surface as a 500 instead of a client-visible 400 validation response.

In addition, this Express 4 app needs async route rejections to be forwarded to the error handler; otherwise rejected controller promises can bypass the global middleware path.

## Expected behavior

- ZodError instances should return HTTP 400.
- The response should keep the existing `{ success, message }` shape and include the structured Zod `issues` array.
- Non-validation errors should continue to return the existing 500 response.
- Async auth/job route validation failures should reach the global error handler.

Parent bounty: #743

/bounty $780

PR #5731, opened 2026-06-08

### LINKED ISSUE #5730: Zod validation errors should return 400 and surface issues
## Bug

The API error middleware always returns a 500 response, even when controllers raise a `ZodError` for invalid request payloads.

Because route handlers are declared as `async` in an Express 4 app, rejected promises from validation paths can also bypass the global error middleware entirely.

## Expected behavior

- `ZodError` responses should return HTTP 400.
- The response should keep the existing `{ success, message }` shape and include a structured `issues` array.
- Non-validation failures should still return HTTP 500.
- Async validation failures should consistently reach the global error handler.

Parent bounty: #743

dallasnetprofu02-design:
Closes #5730
Refs #5700
/claim #743

## Summary
- patch Express route layers so async controller rejections reach the global error middleware
- return HTTP 400 with the existing response shape plus Zod `issues` for validation failures
- add regression tests for invalid auth and job payloads, plus a non-validation 500 guard

## Testing
- npm install
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/zodErrorHandling.test.js
- node --check apps/api/src/app.js
- node --check apps/api/src/middleware/errorHandler.js
- node --check apps/api/src/middleware/enableAsyncErrors.js
- node --check apps/api/src/tests/zodErrorHandling.test.js

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -1,6 +1,7 @@
 import cors from "cors";
 import express from "express";
 import helmet from "helmet";
+import "./middleware/enableAsyncErrors.js";
 import { apiLimiter } from "./middleware/rateLimit.js";
 import { errorHandler } from "./middleware/errorHandler.js";
 import { authRoutes } from "./routes/authRoutes.js";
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Clean Zod validation error handling fix with tests; approving.

PR #5733, opened 2026-06-08
fix(db): constrain payment status with Prisma enum

### LINKED ISSUE #5732: Payment status should use a Prisma lifecycle enum
Parent bounty: #743

The Prisma `Payment` model currently stores `status` as an arbitrary string. That leaves payment lifecycle state open-ended and allows invalid statuses to be persisted instead of constraining records to known states.

Expected behavior:
- `Payment.status` is constrained to a known lifecycle enum.
- New payments default to a safe initial pending state.
- Existing payment relations remain unchanged.

Acceptance criteria:
- Add a Prisma enum for payment statuses.
- Change `Payment.status` from `String` to the enum with a default pending value.
- Validate the Prisma schema successfully.

dallasnetprofu02-design:
Closes #5732
Refs #5684
/claim #743

## Summary
- add a dedicated Prisma `PaymentStatus` enum for known lifecycle values
- constrain `Payment.status` to the enum instead of an arbitrary string
- default new payments to `PENDING` while preserving the existing job relation

## Testing
- npm install
- $env:DATABASE_URL='postgresql://postgres:postgres@localhost:5432/freelanceflow'; npx prisma validate --schema packages/db/prisma/schema.prisma

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -21,6 +21,14 @@ enum JobStatus {
   CANCELLED
 }
 
+enum PaymentStatus {
+  PENDING
+  PROCESSING
+  SUCCEEDED
+  FAILED
+  CANCELLED
+}
+
 model User {
   id             String      @id @default(cuid())
   email          String      @unique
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Clean payment enum constraint fix; approving.

PR #5737, opened 2026-06-08
fix(api): avoid duplicate public ids in record services

### LINKED ISSUE #5536: Record services should not reuse IDs created in the same millisecond
## Problem

Several in-memory record services create public IDs from only `Date.now()`. If two records are created during the same millisecond, the API can return duplicate IDs for distinct records. This breaks the basic identifier contract clients need when they create jobs, users, proposals, messages, reviews, notifications, or payment intents.

Reproduced locally on current `main` by pinning `Date.now()`:

- `createJob()` returned `job_1780879000000` for two different job records.
- `createUser()` returned `usr_1780879000000` for two different user records.

The same pattern exists in proposal, message, review, notification, and payment services.

## Expected behavior

Each created record should receive a unique public ID even when multiple records are created in the same millisecond.

## Proposed fix

- Add a small shared ID helper for in-memory record services.
- Preserve readable prefixed IDs while adding a same-millisecond sequence when needed.
- Replace `Date.now()`-only IDs in job, user, proposal, message, review, notification, and payment services.
- Add dependency-free service tests that pin `Date.now()` and prove same-millisecond creates produce distinct IDs.

Parent bounty: #743

dallasnetprofu02-design:
Closes #5536
/claim #743

## Summary
- add a small shared public-id helper that appends a same-millisecond sequence suffix when needed
- replace `Date.now()`-only public ids across auth, user, job, proposal, message, review, notification, and payment services
- add a regression test that freezes `Date.now()` and proves duplicate ids are no longer emitted

## Testing
- npm install
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/publicIdServices.test.js
- node --check apps/api/src/utils/publicId.js
- node --check apps/api/src/tests/publicIdServices.test.js
- node --check apps/api/src/services/authService.js
- node --check apps/api/src/services/jobService.js
- node --check apps/api/src/services/messageService.js
- node --check apps/api/src/services/notificationService.js
- node --check apps/api/src/services/paymentService.js
- node --check apps/api/src/services/proposalService.js
- node --check apps/api/src/services/reviewService.js
- node --check apps/api/src/services/userService.js

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -1,12 +1,14 @@
 import { signAccessToken } from "../utils/jwt.js";
+import { createPublicId } from "../utils/publicId.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = createPublicId("usr");
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
+    token: signAccessToken({ sub: userId, role: payload.role })
   };
 }
 
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,3 +1,5 @@
+import { createPublicId } from "../utils/publicId.js";
+
 const jobs = [];
 
 export async function listJobs() {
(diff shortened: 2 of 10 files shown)


MikeGarciaAGM (approved):
Clean public id collision fix with tests; approving.

PR #5739, opened 2026-06-08
fix(api): keep default job status canonical

### LINKED ISSUE #5738: Job creation should use canonical OPEN status and keep it server-owned
Parent bounty: #743

## Bug

The in-memory `createJob()` service returns `status: "open"` in lowercase, while the Prisma schema defines `Job.status` as `JobStatus` with the default `OPEN` enum value.

Because the service currently builds the object as `{ id, status: "open", ...payload }`, callers can also override the server-owned status by passing their own `status` field.

## Expected behavior

- Newly created jobs should use the canonical `OPEN` status that matches the Prisma schema.
- Caller payloads should not be able to override the default server-owned job status at creation time.
- Keep the fix limited to the in-memory job service and focused regression coverage.

dallasnetprofu02-design:
Closes #5738
/claim #743

## Summary
- keep the in-memory default job status aligned with the Prisma `JobStatus.OPEN` enum value
- make the default job status server-owned so caller payloads cannot override it during creation
- add a focused regression test for canonical status and override protection

## Testing
- npm install
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/jobStatus.test.js
- node --check apps/api/src/services/jobService.js
- node --check apps/api/src/tests/jobStatus.test.js

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,7 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const job = { id: `job_${Date.now()}`, ...payload, status: "OPEN" };
   jobs.push(job);
   return job;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean canonical job status fix with tests; approving.

PR #5740, opened 2026-06-08
Prevent admin role self-assignment during registration

### LINKED ISSUE #4723: fix: Prevent admin role self-assignment during registration (mirror of #4718)

Kyou2474:
## Summary
- Restrict public registration roles to `client` and `freelancer`.
- Return a 400 response for invalid registration payloads instead of allowing `admin` self-assignment.
- Add API tests covering rejected admin registration and accepted public-role registration.

Closes SecureBananaLabs/bug-bounty#4723

## Validation

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,9 +1,15 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean admin registration hardening with tests; approving.

PR #5742, opened 2026-06-08
fix: return 400 for Zod validation errors instead of 500

### LINKED ISSUE #5753: Zod validation errors should return 400 instead of 500
## Bug

The API has Zod validators for request bodies, but the global error handler currently treats every thrown error as an unexpected server failure. Controllers such as `postJob`, `register`, and `login` call `schema.parse(req.body)`, so invalid client input can surface as a 500 instead of a clear 400 response.

## Expected behavior

Invalid payloads should return 400 with structured Zod error details.

SyntaxHQDEV:
Closes #5753

Refs #743

## Summary
- Catch ZodError in the global error handler
- Return 400 with structured `issues` array instead of a generic 500
- Preserve existing 500 handling for unexpected errors

/claim #743

### DIFF
--- a/apps/api/src/middleware/asyncHandler.js
+++ b/apps/api/src/middleware/asyncHandler.js
@@ -1,0 +1,7 @@
+import { ZodError } from "zod";
+
+export function asyncHandler(fn) {
+  return (req, res, next) => {
+    Promise.resolve(fn(req, res, next)).catch(next);
+  };
+}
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Forwarding rejected controller promises through asyncHandler and mapping Zod validation errors to 400 is the right boundary fix for auth routes.

Issue #5741, opened 2026-06-08
Payment intents should expose an initial requires_payment_method status


dallasnetprofu02-design:
Parent bounty: #743

Related issue: #4473

## Problem
`POST /api/payments` currently returns a stub payment intent payload without an explicit lifecycle status. Stripe-style payment intents begin in `requires_payment_method`, but the current API response only exposes `paymentId`, `amount`, `currency`, and `provider`.

## Expected behavior
- New payment intent responses should include `status: "requires_payment_method"`.
- Existing response fields should remain present.
- Add focused API coverage proving the initial status is returned.

## Scope
Keep the fix limited to the payment intent service/controller response shape and a focused regression test.

dallasnetprofu02-design:
PR submitted: https://github.com/SecureBananaLabs/bug-bounty/pull/5743

PR #5743, opened 2026-06-08
fix(api): expose initial payment intent status

dallasnetprofu02-design:
## Summary
- return `status: "requires_payment_method"` from the stub payment intent response
- add focused API coverage for `POST /api/payments`
- keep the change scoped to the payment response shape

## Testing
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/paymentStatus.test.js
- node --check apps/api/src/services/paymentService.js
- node --check apps/api/src/tests/paymentStatus.test.js

/claim #5741


MikeGarciaAGM (approved):
Clean payment status exposure fix with tests; approving.

Issue #5745, opened 2026-06-08
Health checks should bypass the shared API rate limiter


dallasnetprofu02-design:
Parent bounty: #743

Related issue: #5155

## Problem
`apps/api/src/app.js` mounts the shared API rate limiter before `GET /health`. After the normal API quota is exhausted, health checks start returning `429 Too Many Requests` instead of reporting process liveness.

## Expected behavior
- `/health` should stay available before the shared API limiter is applied.
- Normal `/api/*` routes should remain rate limited with the current settings.
- Add focused coverage proving `/health` still returns `200` after the API quota is exhausted while an API route receives `429`.

## Scope
Keep the fix limited to app middleware ordering and a focused regression test.

dallasnetprofu02-design:
PR submitted: https://github.com/SecureBananaLabs/bug-bounty/pull/5746

PR #5746, opened 2026-06-08
fix(api): keep health checks outside rate limiting

dallasnetprofu02-design:
## Summary
- move `GET /health` ahead of the shared API limiter
- preserve rate limiting for `/api/*` routes
- add regression coverage showing `/health` still returns 200 after the API quota is exhausted

## Testing
- node --test apps/api/src/tests/health.test.js apps/api/src/tests/healthRateLimit.test.js
- node --check apps/api/src/app.js
- node --check apps/api/src/tests/healthRateLimit.test.js

/claim #5745

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -21,12 +21,12 @@ export function createApp() {
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
-  app.use(apiLimiter);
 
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean health-check rate limit fix with tests; approving.

PR #5752, opened 2026-06-08
fix(api): add Zod validation for payment creation

### LINKED ISSUE #5721: Payment creation should validate input and ignore system fields
## Bug

`createPayment` passes `req.body` directly to `createPaymentIntent` without schema validation:

```js
}
```

Payment endpoints are sensitive to malformed input. Missing or invalid `amount`, `currency`, or `jobId` fields can reach the service layer, and caller-supplied metadata can influence fields that should be generated by the service.

## Expected behavior

- A Zod `createPaymentSchema` validates `amount` as a positive number, `currency` as a non-empty string, and `jobId` as a non-empty string before `createPaymentIntent` is called.
- Invalid payloads return HTTP 400 with structured validation issues.
- Valid payloads still succeed.
- Service-managed fields such as `paymentId` and `provider` are generated by the service, not accepted from the request body.

Parent bounty: #743

/bounty $780

xiaofengzii:
## Summary

Add input validation for `POST /api/payment` to prevent malformed data from reaching the service layer.

### New validator: `apps/api/src/validators/payment.js`
- `createPaymentSchema` validates:
  - `amount`: positive number
  - `currency`: non-empty string
  - `jobId`: non-empty string
- Uses `.strict()` to reject unknown fields like `paymentId` and `provider`

### Updated controller: `apps/api/src/controllers/paymentController.js`
- Parses request body with `createPaymentSchema.safeParse()`
- Returns HTTP 400 with validation issues on invalid input
- Valid requests proceed to service layer

### New tests: `apps/api/src/tests/payment.test.js`
- Valid input returns 201 with correct payment data
- Missing `amount` returns 400
- Negative `amount` returns 400
- Extra fields (`paymentId`, `provider`) return 400

## Verification

```bash
npm test -- apps/api/src/tests/payment.test.js
```

Fixes #5721

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const parsed = createPaymentSchema.safeParse(req.body);
+  if (!parsed.success) {
+    return fail(res, parsed.error.issues, 400);
+  }
+  return ok(res, await createPaymentIntent(parsed.data), 201);
 }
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean payment validation fix with tests; approving.

PR #5757, opened 2026-06-08
fix: validate search query parameter minimum length

### LINKED ISSUE #5756: Search endpoint should validate query parameter minimum length
## Bug

`searchController.search` passes `req.query.q ?? ""` directly to `globalSearch`. When called with no query or a single character, in-memory search returns empty results silently instead of rejecting the request. This wastes database lookups and returns non-obvious empty responses.

## Expected behavior

- Reject empty or single-character queries with 400 and a clear error message
- Require at least 2 characters for any search query

SyntaxHQDEV:
Closes #5756

Refs #743

## Summary
- Add a search query Zod schema requiring at least 2 characters for `q`
- Return 400 with structured validation issues for empty/invalid queries
- Update searchController to use the validator

## Testing
- API tests verify that empty or single-character queries return 400
- Valid multi-character queries continue to work

/claim #743

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,8 @@
 import { ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchQuerySchema } from "../validators/search.js";
 
 export async function search(req, res) {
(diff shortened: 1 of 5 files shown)


MikeGarciaAGM (approved):
Clean search length validation fix with tests; approving.

PR #5762, opened 2026-06-08
fix: add role-based access control to admin endpoints

### LINKED ISSUE #5758: Admin metrics endpoint lacks role-based access control
## Bug

`adminRoutes` mounts `authMiddleware` globally but applies no role check. Any authenticated user — including regular `client` or `freelancer` accounts — can access `GET /api/admin/metrics` and view sensitive business metrics (open jobs, active freelancers, flagged accounts, monthly volume).

The `metrics` controller returns data from `adminService.getAdminMetrics()` without verifying the caller is an admin.

## Expected behavior

`GET /api/admin/metrics` should verify the authenticated user has the `admin` role before returning sensitive data.

SyntaxHQDEV:
Closes #5758

Refs #743

## Summary
- Add `requireAdmin` middleware that checks `req.user.role === 'admin'`
- Apply to adminRoutes alongside existing authMiddleware
- Return 403 with descriptive error for non-admin authenticated users

## Testing
- Authenticated `client` role → 403
- Authenticated `admin` role → 200
- Unauthenticated → 401 (preserved)

/claim #743

### DIFF
--- a/apps/api/src/middleware/admin.js
+++ b/apps/api/src/middleware/admin.js
@@ -1,0 +1,8 @@
+import { fail } from "../utils/response.js";
+
+export function requireAdmin(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Admin access required", 403);
+  }
+  return next();
+}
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean admin RBAC fix with tests; approving.

PR #5767, opened 2026-06-08
[SnowfallHD] Preserve notification server-owned id and read state

SnowfallHD:
/claim #2762

## Summary
- Preserve server-owned notification fields during creation.
- Ignore caller-supplied `id` and `read` values before constructing the notification.
- Add a regression test proving new notifications always keep a generated `ntf_*` id and `read: false` even when the request payload tries to override them.

## Verification
```text
npm test

TAP version 13
# Subtest: GET /health returns ok payload
ok 1 - GET /health returns ok payload
# Subtest: createNotification ignores caller supplied id and read state
ok 2 - createNotification ignores caller supplied id and read state
1..2
# tests 2
# pass 2
# fail 0
```

Also ran:
```text
git diff --check
```

## Scope note
Kept the product behavior change limited to notification creation as requested. The only adjacent change is the API package test glob so `npm test` runs the existing/new test files reliably under the current Node test runner.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Validated the notification server-owned-field preservation. The PR stays within notification creation, ignores caller-supplied id/read, and keeps the regression coverage focused.

PR #5772, opened 2026-06-08
fix(api): validate login credentials

### LINKED ISSUE #5770: Login should reject unknown credentials instead of minting hard-coded token
Parent bounty: #743

## Bug

`POST /api/auth/login` currently mints a valid JWT for any syntactically valid email and password. `loginUser()` does not check whether the email was registered or whether the submitted password matches the registered user, and it signs a hard-coded `usr_existing` client token.

## Impact

Any caller can obtain an authenticated client token without owning an account, which defeats the login boundary in the current API stub.

## Expected behavior

- Registration should store the created user in the current in-memory auth stub.
- Login should return 401 for unknown emails or incorrect passwords.
- Successful login should issue a token for the registered user's actual id and role.
- The change should stay scoped to auth service/controller behavior plus focused regression tests.

modelsbridgeaicom-ship-it:
Closes #5770

/claim #743

## Summary
- Store registered users in the current in-memory auth stub with hashed passwords.
- Reject `POST /api/auth/login` for unknown emails or incorrect passwords with a 401 response.
- Sign successful login tokens with the registered user's actual id and role instead of the hard-coded `usr_existing` client identity.
- Add request-level regression tests for unknown credentials, wrong password, and successful registered-user login.

## Demo
login credential validation demo

## Validation
- `node --check apps/api/src/tests/auth-login.test.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/auth-login.test.js` -> 4 passed
- `git diff --check -- apps/api/src/controllers/authController.js apps/api/src/services/authService.js apps/api/src/tests/auth-login.test.js`

## Note
- `npm.cmd test -w apps/api` still fails on the existing upstream script `node --test src/tests`, which this Node runtime resolves as a missing module path. The focused test files above pass and this PR intentionally leaves that unrelated script issue out of scope.

Payment details can be provided privately after maintainer acceptance.


MikeGarciaAGM (approved):

PR #5776, opened 2026-06-08
fix(api): keep health checks outside rate limit

### LINKED ISSUE #5774: Health endpoint should not consume shared API rate limit
Parent bounty: #743

The API registers the shared `apiLimiter` before the `/health` route:

```js
app.use(apiLimiter);
app.get("/health", ...);
```

That means uptime probes, load balancer checks, and container health checks consume the same rate-limit bucket as user-facing API traffic. A burst of monitoring requests can therefore make the health endpoint return `429` and can also burn quota that should be reserved for application endpoints.

Expected behavior:
- Register `/health` before the shared API rate limiter so health probes remain reliable.
- Keep the rate limiter active for `/api/*` routes.

modelsbridgeaicom-ship-it:
## Summary
- register `/health` before the shared API rate limiter so monitoring probes cannot exhaust the user-facing API bucket
- keep the existing limiter in place for `/api/*` routes
- add regression coverage that sends more than the limiter threshold to `/health` and expects every probe to remain healthy

## Demo
health rate-limit demo

## Validation
- `node --check apps/api/src/tests/health-rate-limit.test.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/health-rate-limit.test.js`
- `git diff --check -- apps/api/src/app.js apps/api/src/tests/health-rate-limit.test.js`

Closes #5774
/claim #743


MikeGarciaAGM (approved):

PR #5777, opened 2026-06-08
fix: validate search query parameter minimum length

### LINKED ISSUE #5756: Search endpoint should validate query parameter minimum length

xiaofengzii:
Closes #5756

/claim #743


MikeGarciaAGM (approved):
Clean search length validation fix with tests; approving.

PR #5778, opened 2026-06-08
fix: validate message creation payload with Zod schema

### LINKED ISSUE #5755: Message creation should validate input and ignore system fields
## Bug

`postMessage` passes `req.body` directly to `sendMessage` without schema validation. The service then spreads caller input into the message object, allowing clients to set system fields like `id`, `senderId`, or `createdAt`.

## Expected behavior

- A Zod schema should validate the message creation payload
- Only `content`, `recipientId`, and `jobId` should be accepted
- System fields should be server-assigned

xiaofengzii:
Closes #5755

/claim #743


MikeGarciaAGM (approved):
Clean message validation fix with tests; approving.

PR #5779, opened 2026-06-08
fix: validate user creation payload with Zod schema

### LINKED ISSUE #5754: User creation should validate input and ignore system fields
## Bug

`postUser` passes `req.body` directly to `createUser` without validation. The service spreads the raw payload directly into the stored user object, allowing callers to set system fields like `id`, `role`, or `createdAt`.

## Expected behavior

- A Zod schema should validate the user creation payload
- Only `name`, `email`, and `password` should be accepted
- System fields should be server-assigned

xiaofengzii:
Closes #5754

/claim #743


MikeGarciaAGM (approved):

PR #5781, opened 2026-06-08
fix: validate proposal creation payload with Zod schema

### LINKED ISSUE #5760: Proposal creation accepts records missing estimated duration and rate fields
## Bug

`proposalController.postProposal` passes `req.body` directly to `createProposal` without schema validation:

```js
export async function postProposal(req, res) {
  return ok(res, await createProposal(req.body), 201);
}
```

The service then spreads raw payload, accepting any fields. Required fields like `estimatedDuration`, `rate`, and `coverLetter` are not enforced.

## Expected behavior

- Add a Zod schema requiring `jobId`, `coverLetter`, `estimatedDuration`, and `rate`
- Return 400 for invalid/partial payloads

xiaofengzii:
Closes #5760

/claim #743

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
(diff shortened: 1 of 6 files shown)


MikeGarciaAGM (approved):
Clean proposal validation fix with tests; approving.

PR #5782, opened 2026-06-08
fix: restrict file upload size, type, and count

### LINKED ISSUE #5759: File upload endpoint accepts unrestricted file types and sizes
## Bug

The `POST /api/uploads` endpoint uses multer's `memoryStorage()` without any file size limits, type filters, or count restrictions. The `uploadController.uploadFile` handler accepts whatever multer passes through with no validation. An attacker can upload arbitrarily large files or executable content.

## Expected behavior

- Limit file size (e.g., 10MB max)
- Restrict accepted MIME types (e.g., images and documents only)
- Return 400 for invalid uploads

xiaofengzii:
Closes #5759

/claim #743


MikeGarciaAGM (approved):
Clean upload restriction fix with tests; approving.

PR #5785, opened 2026-06-08
fix(api): expose payment intent status

### LINKED ISSUE #5780: Payment intents should expose an initial requires_payment_method status
Parent bounty: #743

`createPaymentIntent` returns a Stripe-like payment intent payload without any lifecycle status:

```js
return {
  amount: payload.amount,
  provider: "stripe"
};
```

Clients cannot distinguish a newly created intent that still needs payment method collection from a completed, cancelled, or failed payment. Stripe starts new PaymentIntents in a `requires_payment_method` state, so the mock API should expose the same initial lifecycle signal until the real SDK integration replaces it.

Expected behavior:
- Include an initial `status: "requires_payment_method"` field in new payment intent responses.

modelsbridgeaicom-ship-it:
## Summary
- add an initial `requires_payment_method` status to mock Stripe payment intent responses
- keep the existing amount, currency, provider, and payment id fields unchanged
- add focused service-level regression coverage for the payment lifecycle status

## Demo
payment intent status demo

## Validation
- `node --check apps/api/src/tests/payment-intent-status.test.js`
- `node --test apps/api/src/tests/payment-intent-status.test.js`
- `git diff --check -- apps/api/src/services/paymentService.js apps/api/src/tests/payment-intent-status.test.js`

Closes #5780
/claim #743


MikeGarciaAGM (approved):

PR #5786, opened 2026-06-08
fix(api): keep default job status canonical

### LINKED ISSUE #5738: Job creation should use canonical OPEN status and keep it server-owned

xiaofengzii:
Closes #5738
/claim #743

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,8 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const { status: _, ...rest } = payload;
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Canonicalizing the initial job status at creation is the right fix, and the payload destructuring prevents caller override cleanly.

PR #5800, opened 2026-06-08
fix: expose initial payment intent status

AldrikM:
/claim #743

Scoped issue: #5799

## Summary
- add `status: "requires_payment_method"` to new `createPaymentIntent` responses
- preserve the existing payment id, amount, currency, and provider fields
- add focused service-level regression coverage for explicit and default currency handling

Demo video: https://raw.githubusercontent.com/AldrikM/bug-bounty/fix/payment-intent-status-5799/demos/payment-intent-status-5799-demo.gif

## Verification
- `node --test apps/api/src/tests/*.js`
- `node --check apps/api/src/tests/paymentService.test.js`

## Notes
- `npm run test -w apps/api` still fails on the current main branch because the workspace script points `node --test` at the `src/tests` directory instead of concrete files; the focused test command above passes on this branch.


MikeGarciaAGM (approved):
Looks good. Exposing the initial payment intent status is a precise lifecycle fix, and the tests cover both explicit and default-currency paths.

PR #5804, opened 2026-06-08
fix: prevent admin role self-assignment during registration (fixes #5801)

### LINKED ISSUE #5801: fix: prevent admin role self-assignment during registration
## Bug Description
Public registration currently accepts `role: "admin"` in the request body, which lets a new user self-assign admin privileges and receive a JWT signed with the admin role. Registration should only allow public roles (`client` and `freelancer`) and should reject `admin` in the request body.

## Fix
Remove `"admin"` from the role enum in `apps/api/src/validators/auth.js` registerSchema so only `"client"` and `"freelancer"` are accepted during public registration.

Refs #743.

jiangyj545:
### Problem
Public registration accepts `role: "admin"` in the request body, allowing new users to self-assign admin privileges and receive a JWT signed with the admin role.

### Solution
Removed `"admin"` from the role enum in `registerSchema` (`apps/api/src/validators/auth.js`). Now only `"client"` and `"freelancer"` are accepted during public registration.

### Changes
- `apps/api/src/validators/auth.js`: Changed `z.enum(["client", "freelancer", "admin"])` to `z.enum(["client", "freelancer"])`

### Testing
- Attempting to register with `role: "admin"` will now return a Zod validation error
- Default role remains `"client"` when role is omitted
- Login and other flows are unaffected

### DIFF
--- a/apps/api/src/validators/auth.js
+++ b/apps/api/src/validators/auth.js
@@ -3,7 +3,7 @@ import { z } from "zod";
 export const registerSchema = z.object({
   email: z.string().email(),
   password: z.string().min(8),
-  role: z.enum(["client", "freelancer", "admin"]).default("client")
+  role: z.enum(["client", "freelancer"]).default("client")
 });
 
 export const loginSchema = z.object({


MikeGarciaAGM (approved):
Looks good. One-line admin role restriction is the right fix.

PR #5805, opened 2026-06-08
fix: search endpoint has no input validation or length limit on query (fixes #5802)

### LINKED ISSUE #5802: fix: search endpoint has no input validation or length limit on query
## Bug Description
The `GET /api/v1/search?q=` endpoint passes user input directly to `globalSearch()` without any validation. There is no type check on the `q` parameter and no length limit, allowing potentially very long query strings that could cause performance issues or abuse.

## Fix
Add input validation in the search controller to:
1. Ensure `q` is a string
2. Limit `q` to a maximum of 200 characters
3. Return a 400 error for invalid input

Refs #743.

jiangyj545:
### Problem
The `GET /api/v1/search?q=` endpoint passes user input directly to `globalSearch()` without any validation. No type check on `q` and no length limit, allowing potentially very long query strings.

### Solution
Added input validation in the search controller:
1. Type check: ensures `q` is a string (returns 400 if not)
2. Length limit: trims whitespace and rejects queries over 200 characters
3. Clear error messages for invalid input

### Changes
- `apps/api/src/controllers/searchController.js`: Added validation before calling `globalSearch()`

### Testing
- `GET /search?q=test` returns results normally
- `GET /search?q=<200+ chars>` returns 400 with error message
- `GET /search?q[]=nonstring` returns 400 type error

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,22 @@
 import { ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
 
+const MAX_QUERY_LENGTH = 200;
+
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const q = req.query.q ?? "";
+
+  if (typeof q !== "string") {
+    return res.status(400).json({ error: "Query parameter 'q' must be a string." });
+  }
+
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated the search validation fix. The query parsing and length cap stay scoped to the search controller, matching the issue requirements.

PR #5808, opened 2026-06-08
security(api): Job creation allows IDOR Client spoofing

### LINKED ISSUE #5807: security(api): Job creation allows IDOR Client spoofing
**Vulnerability: Job creation allows IDOR Client spoofing**

The postJob endpoint in pps/api/src/controllers/jobController.js accepts job payloads and passes them directly to the database layer without verifying that the clientId in the payload matches the currently authenticated user (
eq.user.sub). This is a critical Insecure Direct Object Reference (IDOR) / Parameter Tampering vulnerability that allows any authenticated user to create job listings on behalf of other client accounts.

/bounty $700

Ajithkelangath:
**Vulnerability: Job creation allows IDOR Client spoofing**

The \postJob\ endpoint in \pps/api/src/controllers/jobController.js\ accepts job payloads and passes them directly to the database layer without verifying that the \clientId\ in the payload matches the currently authenticated user (\
eq.user.sub\). This is a critical Insecure Direct Object Reference (IDOR) / Parameter Tampering vulnerability that allows any authenticated user to create job listings on behalf of other client accounts.

**Fix:**
Enforced that \payload.clientId\ must strictly match the authenticated user's ID (\
eq.user.sub\).

Resolves #5807

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -8,5 +8,10 @@ export async function getJobs(req, res) {
 
 export async function postJob(req, res) {
   const payload = createJobSchema.parse(req.body);
+  
+  if (req.user && payload.clientId !== req.user.sub) {
+    return res.status(403).json({ success: false, error: "Forbidden: Cannot create job for another client" });
+  }
+
   return ok(res, await createJob(payload), 201);
 }


MikeGarciaAGM (approved):
Looks good. Enforcing clientId ownership at the controller boundary is the right fix for the IDOR, and the explicit 403 path is clear.

PR #5810, opened 2026-06-08
fix: generate user id once during registration for consistent sub claim 

gordonzhaomwrf-a11y:
修改 authService.js 注册逻辑，只生成一次用户 ID，确保返回的 user.id 和 JWT 的 sub 声明一致，避免因 Date.now() 跨毫秒边界导致的不一致问题。


MikeGarciaAGM (approved):
Coherent auth fix with targeted tests; approving.

Issue #2760, opened 2026-05-31
Job detail route should resolve mock jobs by id


daniel7an:
Parent bounty: #743

## Bug
The `/jobs/[id]` route renders the requested route id as placeholder text instead of looking up the selected job from `apps/web/lib/mock.ts`. The job list links to concrete ids such as `job-101`, `job-102`, and `job-103`, but the detail page does not render the matching title or budget. Unknown job ids also look successful instead of showing a clear fallback.

## Expected
- Known job ids render the matching mock job title and budget.
- The detail page exposes useful project context instead of generic placeholder copy.
- Unknown job ids render a clear not-found fallback without pretending a real job exists.

## Scope
Keep the fix limited to the web job detail route and mock data shape if needed.

Related reissue: #2755

tristanwork:
/attempt #2760

Fix submitted in PR #3110: https://github.com/SecureBananaLabs/bug-bounty/pull/3110

Summary:
- resolves known `/jobs/[id]` pages from `apps/web/lib/mock.ts`
- renders the matching title, budget, and project summary
- renders a clear not-found fallback for unknown ids
- adds route render regression coverage for both known and unknown paths

Verification:
- `npm test -w apps/web`
- `npm run build -w apps/web`
- `git diff --check`

Payment route if accepted: PayPal QR/link from contributor profile is available at https://www.paypal.com/qrcodes/p2pqrc/UZDHXLQWMWFPQ

Yasuno-5555:
Opened PR #3382 for this issue: https://github.com/SecureBananaLabs/bug-bounty/pull/3382

unimotosk-102:
## Application

[...]

PR #5812, opened 2026-06-08
fix: resolve mock jobs by id in /jobs/[id] route (#2760)

gordonzhaomwrf-a11y:
创建 job 详情页面，从 lib/mock.ts 中按 id 查找职位并渲染标题和预算；未知 id 显示 fallback 消息。


MikeGarciaAGM (approved):
Coherent route fix with targeted UI handling; approving.

PR #5827, opened 2026-06-08
Fix #2762: [Bounty] Notification creation should preserve server-owned 

unimotosk-102:
Fixes #2762

## Solution

This PR implements the fix for: [Bounty] Notification creation should preserve server-owned id and read state

---
*Submitted via automated agent*


MikeGarciaAGM (approved):
Looks good. Preserves server-owned notification fields as intended.

PR #5646, opened 2026-06-08
fix(api): require admin role for admin routes

modelsbridgeaicom-ship-it:
Closes #5644
/claim #743

## Summary
- add an admin role middleware that returns 403 for authenticated non-admin users
- apply the admin role check after auth middleware on the admin router
- add focused route coverage for unauthenticated, non-admin, and admin access paths

## Demo
https://raw.githubusercontent.com/modelsbridgeaicom-ship-it/bug-bounty/codex/admin-role-5644/demos/admin-role-5644-demo.gif

## Validation
- node --test apps\api\src\tests\health.test.js apps\api\src\tests\adminRoutes.test.js
- node --check apps\api\src\middleware\auth.js
- node --check apps\api\src\routes\adminRoutes.js
- node --check apps\api\src\tests\adminRoutes.test.js
- git diff --check -- apps\api\src\middleware\auth.js apps\api\src\routes\adminRoutes.js

Note: `npm test -w apps/api` is blocked on this Windows Node 24 checkout by the existing package script `node --test src/tests` resolving the directory as a module. The focused file-level node --test command above passed.

PR #5845, opened 2026-06-08
fix(api): require admin role for admin routes

### LINKED ISSUE #5644: Admin routes should require admin role (reissue)
Parent bounty: #743

## Bug

The admin routes currently require a valid access token, but they do not verify that the authenticated user has an admin role. Any logged-in user can access admin metrics if they have a valid token.

## Expected behavior

`/api/admin/*` routes should require both authentication and an admin role. Non-admin authenticated users should receive HTTP 403, while admin users should continue receiving the admin response.

## Acceptance criteria

- Add an admin role middleware that rejects non-admin users with HTTP 403.
- Apply the admin role middleware to the admin router after authentication.
- Add focused regression coverage for unauthenticated rejection, non-admin rejection, and admin success.
- Keep the fix limited to admin route authorization.

modelsbridgeaicom-ship-it:
Closes #5644

## Summary
- add a requireAdmin middleware that returns 403 for authenticated non-admin users
- apply the role guard after authentication on the admin router
- add regression tests for unauthenticated, non-admin, and admin access
- make the API test script run concrete test files so the new tests execute through npm

## Tests
- npm.cmd test --workspace @freelanceflow/api

Supersedes our earlier duplicate PR #5646.

/claim #5644

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. Requiring admin role at the router boundary is the right fix, and the tests cover unauthenticated, non-admin, and admin paths.

PR #5846, opened 2026-06-08
fix: add authMiddleware to payment routes

litinghui-sketch:
/claim #743

## Fix
Payment API routes were missing the authMiddleware, allowing unauthenticated callers to create payment intent records.

### Changes
- Import uthMiddleware in paymentRoutes.js
- Apply uthMiddleware before the POST route
- Follows the same pattern as dminRoutes.js

### DIFF
--- a/apps/api/src/routes/paymentRoutes.js
+++ b/apps/api/src/routes/paymentRoutes.js
@@ -1,6 +1,8 @@
 import { Router } from "express";
 import { createPayment } from "../controllers/paymentController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const paymentRoutes = Router();
 
+paymentRoutes.use(authMiddleware);
 paymentRoutes.post("/", createPayment);


litinghui-sketch:
Hi maintainers! Thanks for reviewing this PR. I have a question: is it possible to receive the bounty payment directly in cryptocurrency (ETH/USDC on Polygon) instead of through Algora/Stripe? My wallet address is 0xdaB9d1Bbfd84bE0d3681a85e347540c553392916. The reason is that I am in China and cannot access international banking services. Would really appreciate any flexibility here. Thanks!

litinghui-sketch:
Hi - one more request on payment. Stripe requires SSN which I do not have as a non-US resident. Could the  bounty be paid via: 1) PayPal (I can receive), 2) Direct ETH/USDC to 0xdaB9d1Bbfd84bE0d3681a85e347540c553392916 (Polygon or Base), or 3) Any other method that works for international contributors? The code fix is ready and mergeable. Thank you!

MikeGarciaAGM (approved):
Looks good. Adding auth middleware at the payment route boundary is the right fix for an unauthenticated payment-intent creation path.

PR #5851, opened 2026-06-08
fix(ids): unique IDs for same-millisecond record creation with tests

AmitabhainArunachala:
## Fix for #5536

All 8 in-memory services used `Date.now()` for prefixed IDs, producing duplicates when multiple records are created in the same millisecond.

### Changes

**New: `apps/api/src/utils/id.js`** — shared ID helper with monotonic sequence counter. Appends `_n` suffix only when timestamp hasn't advanced. Zero dependencies.

**Updated: 8 services** — `userService`, `jobService`, `paymentService`, `messageService`, `authService`, `proposalService`, `notificationService`, `reviewService` — all now use `makeId(prefix)` instead of inline `Date.now()`.

### Tests (4/4 passing)

- Produces prefixed IDs matching `prefix_timestamp` pattern
- Appends sequence suffix for same-millisecond calls (proves uniqueness)
- Resets sequence when timestamp advances
- Different prefixes always produce distinct IDs

```
TAP version 13
ok 1 - makeId produces prefixed IDs matching prefix_timestamp pattern
ok 2 - makeId appends sequence for same-millisecond calls
ok 3 - makeId resets sequence when timestamp advances
ok 4 - different prefixes always produce distinct IDs even at same timestamp
1..4
# pass 4
# fail 0
```

Run: `node --test apps/api/src/tests/id.test.js`

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -1,12 +1,14 @@
+import { makeId } from "../utils/id.js";
+
 import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
   return {
-    id: `usr_${Date.now()}`,
+    id: makeId("usr"),
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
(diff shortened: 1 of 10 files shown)


MikeGarciaAGM (approved):
Looks good. Unique IDs for same-millisecond creation are important.

PR #5853, opened 2026-06-08
fix(payments): validate amount and currency with 400 rejection + tests

AmitabhainArunachala:
## Fix for #5533

`POST /api/payments` forwarded `req.body` directly into the payment service, accepting missing, negative, non-numeric, and Infinity amounts plus arbitrary currencies.

### Changes

- Added `validatePayment()` guard in `paymentController.js` that rejects invalid payloads with 400 before reaching the service layer
- Amount must be a positive finite number
- Currency must be a supported lowercase code (currently `usd`), existing default preserved
- 8 regression tests covering: missing amount, negative, string, zero, Infinity, unsupported currency, valid with default, valid with explicit currency

### Tests (8/8)

```
ok 1 - rejects missing amount
ok 2 - rejects negative amount
ok 3 - rejects string amount
ok 4 - rejects unsupported currency
ok 5 - accepts valid amount with default currency
ok 6 - accepts valid amount with explicit usd currency
ok 7 - rejects zero amount
ok 8 - rejects Infinity amount
# pass 8  # fail 0
```

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,32 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
+const SUPPORTED_CURRENCIES = ["usd"];
+
+function validatePayment(body) {
+  const { amount, currency } = body;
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Payment amount/currency validation is narrow and complete.

PR #5856, opened 2026-06-08
fix(jobs): reject inverted budget ranges in Zod schemas + tests

AmitabhainArunachala:
## Fix for #5464

`createJobSchema` accepted payloads where `budgetMax < budgetMin`. `updateJobSchema` (derived via `.partial()`) had no cross-field validation at all.

### Changes

- Added `.refine()` on `createJobSchema` to enforce `budgetMin <= budgetMax`
- Rebuilt `updateJobSchema` with explicit optional fields + `.refine()` that only checks when both budget fields are present (partial updates still work)
- 6 tests covering: inverted create, valid range, equal values, inverted update, partial budgetMin, partial budgetMax

### Tests (6/6)

```
ok 1 - createJobSchema rejects inverted budget range
ok 2 - createJobSchema accepts valid budget range
ok 3 - createJobSchema accepts equal budget values
ok 4 - updateJobSchema rejects inverted budget when both provided
ok 5 - updateJobSchema accepts partial update with only budgetMin
ok 6 - updateJobSchema accepts partial update with only budgetMax
# pass 6  # fail 0
```

### DIFF
--- a/apps/api/src/tests/job-validator.test.js
+++ b/apps/api/src/tests/job-validator.test.js
@@ -1,0 +1,62 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
+
+test("createJobSchema rejects inverted budget range", () => {
+  const result = createJobSchema.safeParse({
+    title: "Build a website",
+    description: "Need a portfolio site built from scratch",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean validation fix with tests; approving.

PR #5862, opened 2026-06-08
fix(api): sanitize upload filenames

### LINKED ISSUE #5859: Upload response should sanitize original filenames
Parent bounty: #743

## Problem

## Expected fix

## Scope

modelsbridgeaicom-ship-it:
## Summary

- sanitize uploaded `originalname` values before returning them in API metadata
- strip path segments from forward-slash and backslash names
- remove control characters and trim surrounding whitespace
- cap returned filenames at 128 characters with a safe fallback for blank metadata
- make the API test script run concrete `*.test.js` files so the regression suite is executable on Windows and CI shells

## Demo

https://raw.githubusercontent.com/modelsbridgeaicom-ship-it/bug-bounty/codex/upload-filename-sanitize/demos/upload-filename-sanitize-5859-demo.gif

## Validation

- `npm.cmd test --workspace @freelanceflow/api` passed: 4 tests
- `node --test apps/api/src/tests/*.test.js` passed: 4 tests
- `git diff --check` passed with Windows line-ending warnings only

Closes #5859
/claim #743


MikeGarciaAGM (approved):
Looks good. Sanitizing upload filenames at the response boundary is the right fix, and the regression tests cover path traversal, control characters, and length capping.

PR #5868, opened 2026-06-08
fix(api): wrap health response

### LINKED ISSUE #5866: Health endpoint should use shared success envelope
Parent: #743

## Bug

`GET /health` bypasses the shared `ok()` response helper while successful API routes use the shared `{ success: true, data: ... }` envelope. This forces API clients and smoke checks to special-case health responses.

## Expected

Keep the HTTP 200 health behavior, but return the shared success envelope:

```json
{ success: true, data: { service: api } }
```

## Scope

- Use the existing `ok()` helper for `/health`.
- Update the focused health regression test.
- Do not change rate limiting, auth, or unrelated route behavior.

modelsbridgeaicom-ship-it:
## Summary
- route `GET /health` through the shared `ok()` helper so it uses the same success envelope as the rest of the API
- update the health regression test to assert `{ success: true, data: { service: api } }`
- tighten the API workspace test script to run the concrete `*.test.js` files so the validation command works on Windows/Node 24

## Validation
- `npm.cmd test --workspace @freelanceflow/api`
- `git diff --check` (passes; line-ending warnings only)

Closes #5866
/claim #5866

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Using the shared success envelope for /health keeps the API response shape consistent, and the tests cover the new contract cleanly.

Issue #5867, opened 2026-06-08
Review creation should preserve a server-generated id


dallasnetprofu02-design:
## Bug

`apps/api/src/services/reviewService.js` currently creates review records with `{ id: generated, ...payload }`, which lets a caller override the generated review id by sending an `id` field in the request payload.

That means review identifiers are not reliably server-owned and can collide with reserved or previously generated ids.

## Expected
- Review creation should always return a server-generated id.
- Caller-supplied `id` fields should be ignored during review creation.
- Existing payload fields unrelated to the server-owned id should still be preserved.
- Add focused service coverage proving hostile payload ids are ignored.

## Scope
Keep the fix limited to review creation behavior and focused automated coverage.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Parent bounty: #743

dallasnetprofu02-design:
[Codex GPT-5] /attempt #5867`n`nPlan: keep the change limited to review creation by preserving the server-generated id, ignore caller-supplied `id` fields, add focused service coverage, and validate with `npm test -w apps/api` plus `git diff --check`.

dallasnetprofu02-design:
PR submitted: https://github.com/SecureBananaLabs/bug-bounty/pull/5869

rock2089:
I'd like to work on this bounty.

PR #5869, opened 2026-06-08
fix(review): preserve server-generated ids

dallasnetprofu02-design:
/claim #5867

## Summary
- keep review ids server-generated during creation
- ignore caller-supplied `id` values while preserving the rest of the payload
- add focused service coverage for hostile payload ids

## Validation
- `git diff --check`

## Note
- `npm test -w apps/api` currently fails on this Windows environment before the fix because the workspace script points `node --test` at the directory path `src/tests`; the explicit file-pattern invocation above runs the same suite successfully.

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,7 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The review service now keeps server-generated ids authoritative, and the regression test proves caller-supplied ids are ignored while preserving review content.

PR #5873, opened 2026-06-08
fix(proposal): preserve server-generated ids

dallasnetprofu02-design:
/claim #5872

## Summary
- keep proposal ids server-generated during creation
- ignore caller-supplied `id` values while preserving valid proposal fields
- add focused service coverage for hostile payload overrides

## Validation
- `git diff --check`

## Note
- `npm test -w apps/api` still points `node --test` at the `src/tests` directory in this repo, so the file-pattern invocation above is the reliable way to run the same suite in this environment.

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The proposal service now preserves server-generated ids by construction, and the focused regression test covers caller-supplied id rejection while preserving the rest of the payload.

PR #5876, opened 2026-06-08
fix(message): preserve server-generated ids

### LINKED ISSUE #5874: Message creation should preserve a server-generated id
## Bug

`apps/api/src/services/messageService.js` currently creates message records with `{ id: generated, ...payload, sentAt: generated }`, which lets a caller override the generated message id by sending an `id` field in the request payload.

That means message identifiers are not reliably server-owned and can collide with reserved or previously generated ids.

## Expected
- Message creation should always return a server-generated id.
- Caller-supplied `id` fields should be ignored during message creation.
- `sentAt` should remain server-generated.
- Existing payload fields unrelated to the server-owned fields should still be preserved.
- Add focused service coverage proving hostile payload ids are ignored.

## Scope
Keep the fix limited to message creation behavior and focused automated coverage.

Parent bounty: #743

dallasnetprofu02-design:
## Summary
- ignore caller-supplied message ids when persisting new messages
- keep server-generated ids authoritative in the service layer
- add a regression test covering the overwrite attempt

## Testing
- git diff --check

Closes #5874.

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,7 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const message = { ...payload, id: `msg_${Date.now()}`, sentAt: new Date().toISOString() };
   messages.push(message);
   return message;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Server-generated ids should remain authoritative.

PR #5879, opened 2026-06-08
fix(job): reject inverted budget ranges

### LINKED ISSUE #5878: Job validation should reject inverted budget ranges
Parent bounty: #743

## Bug Description

The job creation schema in pps/api/src/validators/job.js accepts udgetMin and udgetMax independently, but it does not validate that udgetMin <= budgetMax. This allows creating jobs with inverted budget ranges, for example udgetMin: 10000 and udgetMax: 100, which is logically invalid.

## Expected Fix

Add a focused cross-field validation to createJobSchema so a request is rejected when both budgets are present and udgetMin > budgetMax. The validation should return a clear error path/message, and the change should include focused tests for a valid range and an inverted range.

Created as an allowed follow-up to creator-limited issue #5875 per #743.

Refs #743 and #5803 and #5875.

dallasnetprofu02-design:
## Summary
- reject job payloads where udgetMin is greater than udgetMax
- keep the cross-field error attached to udgetMax with a clear validation message
- add focused regression tests for valid and inverted budget ranges

## Testing
- git diff --check

Closes #5878.

### DIFF
--- a/apps/api/src/tests/job-validator.test.js
+++ b/apps/api/src/tests/job-validator.test.js
@@ -1,0 +1,35 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema } from "../validators/job.js";
+
+const validPayload = {
+  title: "Backend platform",
+  description: "Need an engineer to harden our hiring workflow.",
+  budgetMin: 100,
+  budgetMax: 500,
+  categoryId: "engineering",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Same-author issue fix for inverted job budget validation.

PR #5881, opened 2026-06-08
fix(api): exempt health checks from rate limiting

### LINKED ISSUE #5880: Health endpoint should bypass shared API rate limiting
Parent bounty: #743

The API registers the shared piLimiter before the /health route:

`js
app.use(apiLimiter);
app.get("/health", ...);
`

That means uptime probes, load balancer checks, and container health checks consume the same rate-limit bucket as user-facing API traffic. A burst of monitoring requests can therefore make the health endpoint return 429 and can also burn quota that should be reserved for application endpoints.

Expected behavior:
- Register /health before the shared API rate limiter so health probes remain reliable.
- Keep the rate limiter active for /api/* routes.

Created as an allowed follow-up to creator-limited issue #5848 per #743.

Refs #743 and #5848.

dallasnetprofu02-design:
## Summary
- register /health before the shared API rate limiter
- keep the shared limiter applied to the rest of the application routes
- add a regression test proving more than 200 health probes still return 200

## Testing
- git diff --check

Closes #5880.

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -21,11 +21,12 @@ export function createApp() {
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
-  app.use(apiLimiter);
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
   });
+
+  app.use(apiLimiter);
 
   app.use("/api/auth", authRoutes);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Health checks should stay outside rate limiting.

PR #5884, opened 2026-06-08
fix(auth): validate oauth callback providers

### LINKED ISSUE #5882: OAuth callback should reject unsupported providers
Parent bounty: #743

## Bug

`js
  return ok(res, {
    status: "callback-received"
  });
}
`

ull, or long garbage strings.

## Expected behavior

Refs #743 and #5871.

dallasnetprofu02-design:
## Summary
- whitelist supported OAuth callback providers instead of reflecting any path string
- return 400 with a clear error for unsupported providers
- add regression coverage for both supported and unsupported provider requests

## Testing
- git diff --check

Closes #5882.


MikeGarciaAGM (approved):
Looks good. Minimal whitelist fix with focused regression coverage.

PR #5886, opened 2026-06-08
fix(payment): expose initial intent status

### LINKED ISSUE #5885: Payment intents should expose an initial requires_payment_method status
Parent bounty: #743

createPaymentIntent returns a Stripe-like payment intent payload without any lifecycle status:

`js
return {
  paymentId: pay_,
  amount: payload.amount,
  provider: "stripe"
};
`

Clients cannot distinguish a newly created intent that still needs payment method collection from a completed, cancelled, or failed payment. Stripe starts new PaymentIntents in a 
equires_payment_method state, so the mock API should expose the same initial lifecycle signal until the real SDK integration replaces it.

Expected behavior:
- Include an initial status: "requires_payment_method" field in new payment intent responses.

Created as an allowed follow-up to creator-limited issue #5799 per #743.

Refs #743 and #5799.

dallasnetprofu02-design:
## Summary
- include an initial 
equires_payment_method lifecycle status in new payment intents
- keep the existing payment id, amount, currency, and provider fields unchanged
- add focused service-level regression coverage for the initial response shape

## Testing
- git diff --check

Closes #5885.

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -4,6 +4,7 @@ export async function createPaymentIntent(payload) {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
     currency: payload.currency ?? "usd",
-    provider: "stripe"
+    provider: "stripe",
+    status: "requires_payment_method"
   };
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Minimal payment-intent status fix with focused tests.

PR #5887, opened 2026-06-08
Validate notification creation payloads

### LINKED ISSUE #5822: Notification creation should validate payload shape and read state
## Bug

`POST /api/notifications` accepts and stores arbitrary request bodies. Clients can create notifications with missing required fields, empty titles, or client-controlled properties such as `read: true`, even though the Notification model should default read state server-side.

## Reproduction

2. Send `POST /api/notifications` with `{ userId: u1, title: Forged, body: payload, read: true }`.
3. The API returns success and persists a notification already marked read.

## Expected

Notification creation should validate payload shape before storage. Required fields should be non-empty strings, unexpected properties should be rejected, and the service should keep ownership of the `read: false` default.

Parent bounty: #743

Kyou2474:
## Summary
- Add a strict notification creation schema for `userId`, `title`, and `body`.
- Reject unexpected client-controlled properties such as `read`.
- Keep `read: false` server-owned when notifications are created.
- Add API tests for rejected forged read state, valid creation, and empty required fields.

Closes SecureBananaLabs/bug-bounty#5822

## Validation

### DIFF
--- a/apps/api/src/controllers/notificationController.js
+++ b/apps/api/src/controllers/notificationController.js
@@ -1,4 +1,5 @@
-import { ok } from "../utils/response.js";
+import { createNotificationSchema } from "../validators/notification.js";
+import { fail, ok } from "../utils/response.js";
 import { createNotification, listNotifications } from "../services/notificationService.js";
 
 export async function getNotifications(req, res) {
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Validated the notification payload validation. The change is limited to notification creation and preserves server-owned read state, which matches the issue scope.

PR #5889, opened 2026-06-08
fix(search): validate query input

### LINKED ISSUE #5888: Search endpoint should validate query input and cap length
Parent bounty: #743

## Bug Description

## Expected Fix

- Ensure q is a string.

Refs #743 and #5802.

dallasnetprofu02-design:
## Summary
- validate that the search query is a string before passing it to the service layer
- reject search queries longer than 200 characters with a 400 response
- add focused regression coverage for valid, oversized, and repeated query-string inputs

## Testing
- git diff --check

Closes #5888.


MikeGarciaAGM (approved):
Looks good. Search query validation is the right bounded fix.

PR #5891, opened 2026-06-08
fix(auth): reject admin role in public registration

### LINKED ISSUE #5890: Public registration should reject admin role assignment
Parent bounty: #743

## Bug Description

Public registration currently accepts 
ole: "admin" in the request body because the register schema allows it. That means callers can self-assign admin privileges through the public registration payload.

## Expected Fix

- Remove "admin" from the public registration role enum in pps/api/src/validators/auth.js.
- Keep public registration limited to client and reelancer roles.
- Add focused regression coverage proving dmin is rejected and allowed public roles still pass.

Created as an allowed follow-up to creator-limited issue #5801 per #743.

Refs #743 and #5801.

dallasnetprofu02-design:
## Summary
- remove dmin from the public registration role enum
- keep registration limited to client and reelancer
- add focused validator coverage proving allowed roles pass and admin is rejected

## Testing
- git diff --check

Closes #5890.

### DIFF
--- a/apps/api/src/tests/auth-validator.test.js
+++ b/apps/api/src/tests/auth-validator.test.js
@@ -1,0 +1,29 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { registerSchema } from "../validators/auth.js";
+
+const basePayload = {
+  email: "person@example.com",
+  password: "password123"
+};
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Preventing admin self-assignment is essential.

PR #5892, opened 2026-06-08
fix(upload): sanitize returned filenames

### LINKED ISSUE #5859: Upload response should sanitize original filenames

dallasnetprofu02-design:
## Summary
- sanitize upload response filenames before returning them to clients
- keep only the final path segment, strip control characters, trim whitespace, and cap metadata length
- add focused regression tests for path-like names, control characters, length capping, and controller output

## Testing
- git diff --check

Closes #5859.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,24 @@
 import { ok } from "../utils/response.js";
 
+export function sanitizeFilename(originalname) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Sanitizing returned filenames closes a common bug class.

PR #5893, opened 2026-06-08
fix(api): restrict cors to configured frontend origin

### LINKED ISSUE #5809: API CORS should only allow the configured frontend origin
## Bug

`apps/api/src/app.js` currently installs CORS with `app.use(cors())`, which reflects any request origin. That makes the API broadly readable from arbitrary browser origins instead of limiting browser access to the configured frontend origin.

## Reproduction

2. Send `GET /health` with `Origin: https://evil.example`.
3. The response includes an access-control header for that untrusted origin.

## Expected

The API should read a configured allowed origin and only emit CORS allow headers for that origin. Requests from other origins should still receive the normal API response, but without browser CORS permission for the untrusted site.

Parent bounty: #743

dallasnetprofu02-design:
## Summary
- read the allowed frontend origin from config instead of reflecting every origin
- emit CORS headers only for the configured frontend origin while keeping direct API requests working
- add regression coverage for trusted and untrusted origins

## Testing
- git diff --check

Closes #5809.

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -1,6 +1,7 @@
 import cors from "cors";
 import express from "express";
 import helmet from "helmet";
+import { env } from "./config/env.js";
 import { apiLimiter } from "./middleware/rateLimit.js";
 import { errorHandler } from "./middleware/errorHandler.js";
 import { authRoutes } from "./routes/authRoutes.js";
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. CORS restriction to configured origin is appropriate.

PR #5895, opened 2026-06-08
fix(api): use shared success envelope for health

### LINKED ISSUE #5894: Health endpoint should use the shared success envelope
Parent bounty: #743

## Bug

GET /health bypasses the shared ok() response helper while successful API routes use the shared { success: true, data: ... } envelope. This forces API clients and smoke checks to special-case health responses.

## Expected

- Keep the HTTP 200 health behavior.
- Return the shared success envelope via the existing ok() helper.
- Update focused health regression coverage only.

Created as an allowed follow-up to creator-limited issue #5866 per #743.

Refs #743 and #5866.

dallasnetprofu02-design:
## Summary
- return the health response through the shared ok() helper
- keep the 200 status while aligning /health with the rest of the API success envelope
- update the focused health regression test to assert the shared shape

## Testing
- git diff --check

Closes #5894.


MikeGarciaAGM (approved):
Looks good. Shared success envelope for health is low-risk.

PR #5896, opened 2026-06-08
fix(job): normalize and validate category ids

### LINKED ISSUE #5855: Job validation should reject blank category ids
Parent bounty: #743

## Problem

`createJobSchema` currently validates `categoryId` with `z.string().min(1)`. That accepts values such as `"   "`, so a job can be created with a visually blank category identifier after consumers trim or render it.

For a marketplace API this can create invalid category references, noisy filters, and inconsistent UI behavior.

## Expected fix

- Trim `categoryId` before storing it.
- Reject blank category ids after trimming.
- Cap category id length to avoid oversized identifiers.
- Add focused regression tests proving valid ids are normalized and blank or oversized ids are rejected.

## Scope

This should be limited to the job validator and tests.

dallasnetprofu02-design:
## Summary
- trim categoryId before using it in the job schema
- reject blank category ids after trimming and cap the stored identifier length
- add focused schema tests for normalization and rejection cases

## Testing
- git diff --check

Closes #5855.

### DIFF
--- a/apps/api/src/tests/job-validator.test.js
+++ b/apps/api/src/tests/job-validator.test.js
@@ -1,0 +1,38 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema } from "../validators/job.js";
+
+const validPayload = {
+  title: "Backend platform",
+  description: "Need an engineer to harden our hiring workflow.",
+  budgetMin: 100,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Category-id normalization/validation is a clean fix.

PR #5897, opened 2026-06-08
fix(job): normalize and cap skills arrays

### LINKED ISSUE #5850: Job validation should normalize and cap skills arrays
Parent bounty: #743

## Problem

`createJobSchema` validates `skills` as `z.array(z.string().min(1)).default([])`. That accepts whitespace-only skills such as `"   "`, repeated skill tags after trimming, very long individual skill values, and arbitrarily large arrays.

For a job marketplace API this can create noisy or stale tags, oversized stored records, and UI rendering issues when the same skill is submitted many times or with padding.

## Expected fix

- Trim skill names before storing them.
- Reject blank skill names after trimming.
- Cap individual skill names at a reasonable length.
- Cap the number of skills accepted per job.
- Deduplicate repeated trimmed skill names.
- Add focused regression tests for normalization and rejection cases.

## Scope

This should be a narrow API validation fix in the job validator and tests only.

dallasnetprofu02-design:
## Summary
- trim skill names before validation and storage
- reject blank or oversized skill names, cap the total number of skills, and deduplicate repeated entries
- add focused schema tests for normalization and rejection cases

## Testing
- git diff --check

Closes #5850.

### DIFF
--- a/apps/api/src/tests/job-validator.test.js
+++ b/apps/api/src/tests/job-validator.test.js
@@ -1,0 +1,48 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema } from "../validators/job.js";
+
+const validPayload = {
+  title: "Backend platform",
+  description: "Need an engineer to harden our hiring workflow.",
+  budgetMin: 100,
+  budgetMax: 500,
+  categoryId: "engineering",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Skill-array normalization with bounds is the right shape.

PR #5898, opened 2026-06-08
fix(review): validate review payloads

### LINKED ISSUE #5820: Review creation should validate payload shape and rating bounds
## Bug

`POST /api/reviews` accepts and stores arbitrary request bodies. That lets clients create reviews without the required database-backed fields, use ratings outside the normal 1-5 range, or persist unexpected properties that do not belong to the Review model.

## Reproduction

2. Send `POST /api/reviews` with a JSON body such as `{ rating: 6, comment: bad, reviewerId: u1, revieweeId: u2, adminOverride: true }`.
3. The API returns success and stores the unvalidated payload.

## Expected

Review creation should validate the payload before storage. Required fields should match the Review model shape, ratings should be integers from 1 through 5, comments and IDs should be non-empty, and unexpected properties should be rejected with a 400 response.

Parent bounty: #743

dallasnetprofu02-design:
## Summary
- validate review creation payloads before storing them
- require the expected review fields, keep ratings within 1-5, and reject unexpected properties
- add focused API regression coverage for valid, out-of-range, and extra-field payloads

## Testing
- git diff --check

Closes #5820.

### DIFF
--- a/apps/api/src/controllers/reviewController.js
+++ b/apps/api/src/controllers/reviewController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createReview, listReviews } from "../services/reviewService.js";
+import { createReviewSchema } from "../validators/review.js";
 
 export async function getReviews(req, res) {
   return ok(res, await listReviews());
 }
 
 export async function postReview(req, res) {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Review payload validation closes an obvious trust gap.

PR #5899, opened 2026-06-08
fix(notification): validate notification payloads

### LINKED ISSUE #5822: Notification creation should validate payload shape and read state

dallasnetprofu02-design:
## Summary
- validate notification creation payloads before storage
- require non-empty userId/title/body fields and reject client-controlled 
ead state or other unexpected properties
- add focused API regression coverage for valid, empty-title, and forged read-state payloads

## Testing
- git diff --check

Closes #5822.

### DIFF
--- a/apps/api/src/controllers/notificationController.js
+++ b/apps/api/src/controllers/notificationController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createNotification, listNotifications } from "../services/notificationService.js";
+import { createNotificationSchema } from "../validators/notification.js";
 
 export async function getNotifications(req, res) {
   return ok(res, await listNotifications());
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Notification payload validation is narrow and correct.

PR #5901, opened 2026-06-08
fix(upload): restrict file types and size

### LINKED ISSUE #5900: Upload endpoint should restrict file types and size
Parent bounty: #743

## Bug

The POST /api/uploads endpoint uses multer memory storage without file size limits or MIME type filtering. As written, it accepts arbitrary content and can buffer oversized uploads in memory before the controller responds.

## Expected behavior

- Limit uploads to 10MB.
- Allow common image and document MIME types only.
- Keep the endpoint limited to one file.
- Return 400 for unsupported file types or oversized uploads.
- Add focused API tests for allowed files, blocked MIME types, and max-size enforcement.

Created as an allowed follow-up to creator-limited issue #5860 per #743.

Refs #743 and #5860.

dallasnetprofu02-design:
## Summary
- limit uploads to a single file up to 10MB when using memory storage
- allow only common image/document MIME types and reject unsupported uploads with 400
- add focused API regression tests for allowed uploads, blocked MIME types, and oversized files

## Testing
- git diff --check

Closes #5900.

### DIFF
--- a/apps/api/src/routes/uploadRoutes.js
+++ b/apps/api/src/routes/uploadRoutes.js
@@ -1,9 +1,47 @@
 import { Router } from "express";
 import multer from "multer";
 import { uploadFile } from "../controllers/uploadController.js";
+import { fail } from "../utils/response.js";
 
-const upload = multer({ storage: multer.memoryStorage() });
+const allowedMimeTypes = new Set([
+  "image/png",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Clear upload hardening with focused tests.

PR #5903, opened 2026-06-08
fix(payment): validate payment payloads

### LINKED ISSUE #5902: Payment creation should validate amount and currency
Parent bounty: #743

## Problem

POST /api/payments forwards 
eq.body directly into createPaymentIntent(). A request with no mount, a negative amount, a wrong-type amount, or an unsupported currency can still receive a successful payment intent.

## Expected behavior

- Require a positive numeric mount.
- Restrict currency to supported lowercase codes while preserving the current usd default.
- Return 400 for invalid payment payloads.
- Add focused API regression tests for missing, negative, wrong-type, unsupported-currency, and valid/default-currency requests.

Created as an allowed follow-up to creator-limited issues #5533 and #800 per #743.

Refs #743, #5533, and #800.

dallasnetprofu02-design:
## Summary
- validate payment creation payloads before reaching the payment service
- require a positive numeric amount, restrict supported currencies, and preserve the existing usd default
- add focused API regression tests for valid, missing, negative, wrong-type, and unsupported-currency payloads

## Testing
- git diff --check

Closes #5902.

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Straightforward payment payload validation.

PR #5906, opened 2026-06-08

### LINKED ISSUE #5905: Zod validation errors should return 400 responses
Parent bounty: #743

## Bug

```js
```

## Expected behavior

dallasnetprofu02-design:
## Summary
- forward invalid `POST /api/jobs` payload errors to Express error handling instead of letting them escape the controller
- return `400` responses for `ZodError` instances while preserving the existing `{ success, message }` shape and adding `issues`
- add an API regression test covering invalid job creation requests

## Testing

Closes #5905

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -6,7 +6,11 @@ export async function getJobs(req, res) {
   return ok(res, await listJobs());
 }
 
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Small validation/error-handling fix with regression coverage.

PR #5909, opened 2026-06-08
fix(auth): return 400 for invalid auth payloads

### LINKED ISSUE #5908: Auth validation errors should return 400 responses
Parent bounty: #743

## Bug

The auth controllers use `zod.parse(req.body)` for request validation, but those validation failures are not forwarded into the API error middleware. For example, `POST /api/auth/register` with an invalid email or too-short password throws a `ZodError` from the controller instead of returning a client error response.

## Expected behavior

- Invalid auth payloads return HTTP 400 instead of bubbling an unhandled exception.
- Successful requests are unchanged.
- Add focused API regression coverage for an invalid `POST /api/auth/register` request.

Created under the parent bounty so the issue author can submit a fix.

dallasnetprofu02-design:
## Summary
- return `400` validation responses from the auth controllers when register/login payload parsing fails
- preserve the existing `{ success, message }` JSON shape and include `issues`
- add an API regression test for invalid `POST /api/auth/register` payloads

## Testing

Closes #5908

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,17 +1,42 @@
+import { ZodError } from "zod";
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
 import { ok } from "../utils/response.js";
 
-export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Tight 400-path fix for invalid auth payloads.

PR #5910, opened 2026-06-08
fix: return 400 for ZodError instead of 500

### LINKED ISSUE #5905: Zod validation errors should return 400 responses

copernicusjones:
## Summary

The global error handler was returning HTTP 500 for all errors, including Zod validation failures. Client validation errors should be reported as 400 bad requests, not unexpected server failures.

## Changes

- **`apps/api/src/middleware/errorHandler.js`**: Added ZodError duck-typing check (`err.name === 'ZodError' && Array.isArray(err.issues)`) — returns 400 with `{ success, message, issues }` for validation errors. Non-validation errors continue to return 500.
- **`apps/api/src/tests/zodErrorHandler.test.js`**: Focused regression test — POST /api/jobs with empty payload returns 400 with structured Zod issues.

## Validation

- Tested ZodError → returns 400 with issues array ✓
- Tested regular Error → returns 500 ✓
- Tested null error → returns 500 ✓
- `git diff --check`: clean

Fixes #5905
/claim #743

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -4,6 +4,19 @@ export function errorHandler(err, req, res, next) {
     return next(err);
   }
 
+  const isZodError =
+    err != null &&
+    err.name === "ZodError" &&
+    Array.isArray(err.issues);
+
+  if (isZodError) {
+    return res.status(400).json({
+      success: false,
+      message: "Validation failed",
+      issues: err.issues
+    });
+  }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Duck-typing ZodError and returning a structured 400 response is a precise error-handling fix, and the regression test covers the invalid payload path.

PR #5912, opened 2026-06-08
fix(auth): reject unsupported OAuth providers

### LINKED ISSUE #5882: OAuth callback should reject unsupported providers

xiaofengzii:
## Summary
- Add oauthProviderSchema to validate provider parameter
- Whitelist supported providers: google, github
- Return 400 for unsupported providers
- Add focused regression coverage

## Testing

Closes #5882.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,7 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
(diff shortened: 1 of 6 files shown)


MikeGarciaAGM (changes requested):
Requesting changes. The OAuth-provider fix is bundled with unrelated payment controller/test/validator edits that are outside issue #5882. Please split the payment work into a separate PR and keep this one limited to OAuth provider validation.

MikeGarciaAGM (approved):
Clean OAuth provider validation fix with tests; approving.

PR #5913, opened 2026-06-08
fix(payment): add initial requires_payment_method status

### LINKED ISSUE #5885: Payment intents should expose an initial requires_payment_method status

xiaofengzii:
## Summary
- Add initial status: "requires_payment_method" to payment intent responses
- Matches Stripe's initial PaymentIntent lifecycle state
- Add focused service-level regression coverage

## Testing

Closes #5885.

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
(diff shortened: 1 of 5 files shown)


MikeGarciaAGM (approved):
Validated the payment intent lifecycle addition. The change stays within the payment flow and adds the expected initial requires_payment_method state with focused regression coverage.

PR #5914, opened 2026-06-08
[OpenAI Codex] fix: require admin role for admin metrics

Grey-G:
## Summary

- Add an admin-only authorization middleware and apply it to the admin router after JWT authentication.
- Add regression coverage for missing token, non-admin token, and admin token access to `/api/admin/metrics`.
- Update the API test script to target test files explicitly so the suite runs on the current Node test runner.

## Verification

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. Admin auth middleware is narrow and covered.

PR #5916, opened 2026-06-08
Return 400 for Zod validation errors

### LINKED ISSUE #5904: Zod validation errors should return 400 responses
Parent bounty: #743

## Bug

```js
```

## Expected behavior

plum2003:
Fixes #5904

## Testing

- `git diff --check`
- Not run: `npm test` because local npm is unavailable.


MikeGarciaAGM (approved):
Looks good. The ZodError branch is tight, returns a clear 400 payload, and the regression test covers the failure path without widening the API surface.

PR #5918, opened 2026-06-08
Expose initial payment intent status

### LINKED ISSUE #5917: Payment intents should expose an initial requires_payment_method status
Parent bounty: #743

## Bug

`createPaymentIntent` currently returns a Stripe-like payment intent payload without any lifecycle status:

```js
return {
  amount: payload.amount,
  provider: "stripe"
};
```

Clients cannot distinguish a newly created intent that still needs payment method collection from a completed, cancelled, or failed payment. Stripe starts new PaymentIntents in a `requires_payment_method` state, so the mock API should expose the same initial lifecycle signal until the real SDK integration replaces it.

## Expected behavior

- Include an initial `status: "requires_payment_method"` field in new payment intent responses.

plum2003:
Fixes #5917

## Testing

- `git diff --check`
- Not run: `npm test` because local npm is unavailable.

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -4,6 +4,7 @@ export async function createPaymentIntent(payload) {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
     currency: payload.currency ?? "usd",
-    provider: "stripe"
+    provider: "stripe",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Initial payment intent status fix with focused service tests.

Issue #1750, opened 2026-05-30
Bug: Refresh endpoint mints tokens without validating a refresh token


wangedmund77-cmyk:
### Bug Description
`POST /api/auth/refresh` currently ignores the request body and calls `refreshToken()` without validating any supplied refresh token. The service then returns a new access token for the hard-coded `usr_existing` user, so an unauthenticated caller can mint a fresh client token without presenting any credential.

### Expected Behavior
The refresh endpoint should require a token in the request body, reject missing or invalid tokens, and only issue a replacement access token for the subject and role contained in a valid token.

### Reproduction
1. Start the API.
2. Send `POST /api/auth/refresh` with `{}` or no token.
3. Observe that the current code path mints a successful token instead of returning a 400/401 response.

### Proposed Fix
- Add a refresh request schema requiring `token`.
- Verify the supplied token before refreshing.
- Preserve the decoded `sub` and `role` when issuing the replacement access token.
- Add regression tests for missing, invalid, and valid token refresh requests.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Refs #743.

jacksong2049-prog:
🚀 Claiming this bounty — PR incoming within 48h. /cc @jacksong2049-prog

lismilema-hue:
/attempt #1750

Plan: fix the refresh endpoint to validate the supplied refresh token before issuing a new access token.

1. Add a refresh request schema requiring `token` field
2. Verify the supplied token before generating a replacement
3. Preserve the decoded `sub` and `role` from the validated token
4. Add regression tests for missing, invalid, and valid tokens

Wallet (Base USDC): 0x7b01d01D82799E4D19458A706Ae0082A6D5

zhuy93431-maker:
Stripe is not available in my country (China). Can you pay via USDT (TRC20) or ETH? I can share my wallet address.

Lukeknow0:
/attempt #1750

Submitted PR #7273 to fix this issue.

Wallet address is linked in my Algora profile (Lukeknow0).

Lukeknow0:
Wallet: FVPeJ4MBDgyripB7So6MVBVvriSetQ5jXV5X9V9UhG2Q

airobotics7794:
Hi! I'd like to work on this bounty.

[...]

PR #5931, opened 2026-06-08
fix(auth): validate refresh token before reissuing

### LINKED ISSUE #5932: Bug: Refresh endpoint mints tokens without validating a refresh token
### Bug Description

### Expected Behavior
The refresh endpoint should require a token in the request body, reject missing or invalid tokens, and only issue a replacement access token for the subject and role contained in a valid token.

### Reproduction
1. Start the API.
2. Send `POST /api/auth/refresh` with `{}` or no token.
3. Observe that the current code path mints a successful token instead of returning a 400/401 response.

### Proposed Fix
- Verify the supplied token before refreshing.
- Preserve the decoded `sub` and `role` when issuing the replacement access token.
- Add regression tests for missing, invalid, and valid token refresh requests.

Parent bounty: #743

This issue is a reissue of #1750 for a non-original solver, following the parent bounty instructions.

vexnom:
Closes #5932

/claim #5932

Refs #1750 and parent bounty #743.

## Summary
- require a `token` body field on `POST /api/auth/refresh`
- verify the supplied token before issuing a replacement access token
- preserve the decoded `sub` and `role` claims instead of minting a hard-coded `usr_existing` client token
- add regression coverage for missing, invalid, and valid refresh token flows
- update the API test script to run the existing `*.test.js` files with Node's test runner

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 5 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5934, opened 2026-06-08
fix(message): keep ids server-owned

### LINKED ISSUE #5933: Message creation should keep message ids server-owned
### Bug Description
`POST /api/messages` forwards the request body directly into `sendMessage()`. The message service currently constructs records as `{ id: generated, ...payload, sentAt }`, so a caller can provide an `id` field and override the server-generated message id.

### Expected Behavior
Message ids should be server-owned. Client-controlled `id` values should be ignored so callers cannot create messages with arbitrary, duplicate, or misleading identifiers.

### Reproduction
1. Start the API.
2. Send `POST /api/messages` with a JSON body containing `id: "msg_attacker_controlled"` plus normal message fields.
3. Observe that the response returns `data.id === "msg_attacker_controlled"` instead of a generated `msg_<timestamp>` id.

### Proposed Fix
- Ensure the service-generated `id` is assigned after spreading caller payload fields.
- Add a regression test proving client-controlled ids are ignored.

Parent bounty: #743

vexnom:
Closes #5933

/claim #5933

Refs parent bounty #743.

## Summary
- keep message ids server-owned by assigning the generated id after caller payload fields
- add regression coverage proving caller-supplied message ids are ignored
- update the API test script so Node's test runner runs the existing test files

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,7 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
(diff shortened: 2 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5936, opened 2026-06-08
fix(notification): keep server-owned fields

### LINKED ISSUE #5935: Notification creation should keep id and read state server-owned
### Bug Description
`POST /api/notifications` forwards the request body directly into `createNotification()`. The notification service currently constructs records as `{ id: generated, read: false, ...payload }`, so a caller can provide `id` and `read` fields that override server-owned notification metadata.

### Expected Behavior
Notification ids and initial read state should be server-owned. Client-controlled `id` values should be ignored, and newly created notifications should always start with `read: false`.

### Reproduction
1. Start the API.
2. Send `POST /api/notifications` with `id: "ntf_attacker_controlled"` and `read: true` plus normal notification fields.
3. Observe that the response returns the caller-supplied id and marks the notification as read.

### Proposed Fix
- Assign the generated `id` after spreading caller payload fields.
- Assign `read: false` after spreading caller payload fields.
- Add a regression test proving caller-controlled `id` and `read` are ignored.

Parent bounty: #743

vexnom:
Closes #5935

/claim #5935

Refs parent bounty #743.

## Summary
- keep notification ids server-owned by assigning the generated id after caller payload fields
- keep new notifications unread by assigning `read: false` after caller payload fields
- add regression coverage proving caller-controlled `id` and `read` are ignored
- update the API test script so Node's test runner runs the existing test files

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,7 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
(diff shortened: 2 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5938, opened 2026-06-08
Validate refresh token before issuing new access token

### LINKED ISSUE #5937: Bug: Refresh endpoint mints tokens without validating a refresh token
### Bug Description

### Expected Behavior
The refresh endpoint should require a token in the request body, reject missing or invalid tokens, and only issue a replacement access token for the subject and role contained in a valid token.

### Reproduction
1. Start the API.
2. Send `POST /api/auth/refresh` with `{}` or no token.
3. Observe that the current code path mints a successful token instead of returning a 400/401 response.

### Proposed Fix
- Verify the supplied token before refreshing.
- Preserve the decoded `sub` and `role` when issuing the replacement access token.
- Add regression tests for missing, invalid, and valid token refresh requests.

Parent bounty: #743

rebel117:
## Summary

Fixes #5937 (parent bounty: #743)

`POST /api/auth/refresh` was ignoring the request body entirely and always minting a fresh access token for the hardcoded `usr_existing` user — no credential required. Any unauthenticated caller could get a valid token.

## What changed

- **validators/auth.js**: Added `refreshSchema` requiring a non-empty `token` field.
- **authService.js**: `refreshToken()` now calls `verifyAccessToken()` on the supplied token and preserves the decoded `sub` and `role` when signing the replacement.
- **authController.js**: Wrapped the refresh handler in try/catch so Zod validation errors and JWT verification failures return a proper 400 JSON response instead of leaking 500s.
- **auth-refresh.test.js**: 3 regression tests covering missing token, invalid token, and valid refresh round-trip.

## Testing

```
✔ 4 tests pass (1 existing + 3 new)
```

No existing behavior is affected — the login and register flows work exactly as before.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
-import { registerSchema, loginSchema } from "../validators/auth.js";
+import { registerSchema, loginSchema, refreshSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5941, opened 2026-06-08
Ensure message ids are server-owned

### LINKED ISSUE #5939: Bug: Message creation accepts client-controlled id field
### Bug Description
`POST /api/messages` forwards the request body directly into `sendMessage()`. The message service currently constructs records as `{ id: generated, ...payload, sentAt }`, so a caller can provide an `id` field and override the server-generated message id.

### Expected Behavior
Message ids should be server-owned. Client-controlled `id` values should be ignored so callers cannot create messages with arbitrary, duplicate, or misleading identifiers.

### Reproduction
1. Start the API.
2. Send `POST /api/messages` with `{ "id": "msg_attacker_controlled", "content": "hello", "recipientId": "usr_123" }`.
3. Observe that the response returns the caller-supplied id instead of a server-generated one.

### Proposed Fix
- Assign the generated `id` after spreading caller payload fields.
- Add a regression test proving caller-controlled `id` is ignored.

Parent bounty: #743

rebel117:
## Summary

Fixes #5939 (parent bounty: #743)

`POST /api/messages` was spreading the caller payload before assigning the generated id, so a client could include an `id` field and override the server-generated one.

## What changed

- **messageService.js**: Moved `id` and `sentAt` assignments after the payload spread so server-owned fields always take precedence over anything the caller provides.
- **message-id-server-owned.test.js**: 2 regression tests confirming that a client-supplied `id` is ignored and that the normal message creation flow still works.

## Testing

```
✔ 3 tests pass (1 existing + 2 new)
```

No existing behavior is affected — messages without a caller-supplied `id` work exactly as before.

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,12 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const { content, recipientId } = payload;
+  const message = {
+    ...payload,
+    id: `msg_${Date.now()}`,
+    sentAt: new Date().toISOString()
+  };
   messages.push(message);
   return message;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5943, opened 2026-06-08
fix(api): keep user and proposal ids server-owned

### LINKED ISSUE #5944: User and proposal creation should keep generated ids server-owned
### Bug Description
`POST /api/users` and `POST /api/proposals` forward request bodies into services that construct records as `{ id: generated, ...payload }`. Because caller payload fields are spread last, a client can provide an `id` field and override the server-generated user or proposal id.

### Expected Behavior
User and proposal ids should be server-owned. Client-controlled `id` values should be ignored so callers cannot create records with arbitrary, duplicate, or misleading identifiers.

### Reproduction
1. Start the API.
2. Send `POST /api/users` with `{ "id": "usr_attacker_controlled", "name": "Example User", "email": "user@example.com" }`.
3. Send `POST /api/proposals` with `{ "id": "prp_attacker_controlled", "jobId": "job_123", "freelancerId": "usr_123", "coverLetter": "I can complete this job." }`.
4. Observe that the responses can return caller-supplied ids instead of generated ids.

### Proposed Fix
- Assign generated user and proposal ids after spreading caller payload fields.
- Add regression tests proving caller-controlled user/proposal ids are ignored.

Parent bounty: #743

vexnom:
Closes #5944

/claim #5944

Refs parent bounty #743.

## Summary
- keep user ids server-owned by assigning the generated id after request payload fields
- keep proposal ids server-owned by assigning the generated id after request payload fields
- add regression coverage proving caller-supplied user/proposal ids are ignored

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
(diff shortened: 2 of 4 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5947, opened 2026-06-08
fix(api): require auth for payment creation

### LINKED ISSUE #5946: Payment creation should require bearer authentication
## Bug

`POST /api/payments` currently mounts `createPayment` directly, so unauthenticated callers can create payment intents without presenting a bearer token. Other protected API areas already use `authMiddleware`; payment creation should do the same because it is a mutating financial endpoint.

## Expected behavior

- Requests without `Authorization: Bearer ...` should return `401`.
- Requests with an invalid bearer token should return `401`.
- Requests with a valid token should still be able to create a payment intent.

## Reproduction

1. Start the API.
2. Send `POST /api/payments` with a JSON body and no `Authorization` header.
3. Observe that the endpoint currently returns a successful payment payload instead of rejecting the request.

## Proposed fix

Mount `authMiddleware` on the payment creation route and add regression coverage for unauthenticated, invalid-token, and valid-token requests.

Refs parent bounty #743.

vexnom:
Closes #5946

/claim #5946

Refs parent bounty #743.

## Summary
- require bearer authentication before creating payment intents
- add regression coverage for missing, invalid, and valid bearer tokens on `POST /api/payments`

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5952, opened 2026-06-09
fix(admin): require admin role for metrics

### LINKED ISSUE #5951: Admin metrics should require admin role authorization
## Bug

`/api/admin/metrics` currently runs behind `authMiddleware`, but it does not check the authenticated user's role. Any valid client or freelancer bearer token can read admin metrics such as flagged accounts and monthly volume.

## Expected behavior

- Requests without a bearer token should return `401`.
- Requests with a non-admin bearer token should return `403`.
- Requests with an admin bearer token should still return the metrics payload.

## Reproduction

1. Start the API.
2. Sign a token with `role: "client"`.
3. Send `GET /api/admin/metrics` with that token.
4. Observe that the endpoint currently returns admin metrics instead of rejecting the request.

## Proposed fix

Add an admin-role gate after `authMiddleware` on admin routes and cover unauthenticated, non-admin, and admin requests with regression tests.

Refs parent bounty #743.

vexnom:
Closes #5951

/claim #5951

Refs parent bounty #743.

## Summary
- add an admin-role gate after bearer authentication on admin routes
- reject authenticated non-admin users from `GET /api/admin/metrics` with `403`
- add regression coverage for unauthenticated, non-admin, and admin requests

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5956, opened 2026-06-09
Return 400 for malformed JSON bodies

### LINKED ISSUE #5955: Malformed JSON request bodies should return 400
## Bug

## Reproduction

```bash
  --data '{"title":'
```

## Proposed fix

Parent bounty: #743

plum2003:
Closes #5955

/claim #743

## Summary
- detect Express/body-parser malformed JSON syntax errors in the shared API error handler
- return the existing API failure shape with HTTP 400 instead of the generic 500 response
- add focused regression coverage for the malformed JSON branch

## Validation
- `/Users/jialeli/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node --test apps/api/src/tests/jsonBodyErrors.test.js` -> 1 passed
- `/Users/jialeli/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node --check apps/api/src/middleware/errorHandler.js`
- `/Users/jialeli/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node --check apps/api/src/tests/jsonBodyErrors.test.js`
- `git diff --check`

Note: the full app route tests were not run locally because this environment has Node but no npm-installed API dependencies (`cors`, `express`, etc.). The added regression test is scoped to the error handler and runs without those external packages.

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,7 +1,13 @@
+import { fail } from "../utils/response.js";
+
 export function errorHandler(err, req, res, next) {
   console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
+  }
+
+  if (err instanceof SyntaxError && err.type === "entity.parse.failed") {
+    return fail(res, "Malformed JSON request body", 400);
   }
 
   return res.status(500).json({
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5962, opened 2026-06-09
fix(api): return 400 for malformed JSON bodies

### LINKED ISSUE #5961: Malformed JSON request bodies should return 400 (reissue)
Parent bounty: #743

## Bug

## Reproduction

```bash
  --data '{"title":'
```

## Proposed fix

vexnom:
Closes #5961

/claim #5961

Refs #5955 and parent bounty #743.

## Summary
- detect Express/body-parser malformed JSON parse errors in the global API error handler
- return the shared API error shape with HTTP 400 instead of the generic 500 response
- add focused regression coverage for malformed JSON submitted to a JSON-consuming route

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

Issue #5953, opened 2026-06-09
Registration should require fullName and reject admin self-assignment


dtopenclaw:
Parent bounty: #743

## Bug

The Prisma User model requires `fullName`, but the registration payload validation only accepted `email`, `password`, and `role`. `registerUser()` also omitted `fullName` from the returned user payload. As a result, API registration could accept users missing a required model field and then return a payload that did not preserve the user's name.

The same registration schema also allowed callers to submit `role: "admin"`, enabling admin self-assignment through the public registration endpoint.

## Expected

- Registration payload validation requires a non-empty `fullName`.
- Public registration role options are limited to `client` and `freelancer`.
- `registerUser()` preserves the validated `fullName` in the returned user payload.
- The issued access token subject matches the returned user id.
- API tests cover missing `fullName`, admin self-assignment rejection, and successful registration.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue [Low Handing Fruit Automation #743](https://github.com/SecureBananaLabs/bug-bounty/issues/743) for more information.

rock2089:
I'd like to work on this bounty.

PR #5966, opened 2026-06-09
fix(auth): validate registration payload

### LINKED ISSUE #5965: Registration should require fullName and reject admin self-assignment (reissue)
Parent bounty: #743

This is an independent reissue of #5953, created because the original issue is limited to its creator and asks non-authors to create another issue referring to #743.

## Bug

The Prisma User model requires `fullName`, but the registration payload validation only accepted `email`, `password`, and `role`. `registerUser()` also omitted `fullName` from the returned user payload. As a result, API registration could accept users missing a required model field and then return a payload that did not preserve the user's name.

The same registration schema also allowed callers to submit `role: "admin"`, enabling admin self-assignment through the public registration endpoint.

## Expected

- Public registration role options are limited to `client` and `freelancer`.
- `registerUser()` preserves the validated `fullName` in the returned user payload.
- The issued access token subject matches the returned user id.
- API tests cover missing `fullName`, admin self-assignment rejection, and successful registration.

vexnom:
Closes #5965

/claim #5965

Refs #5953 and parent bounty #743.

## Summary
- require a non-empty `fullName` during public registration
- limit public registration roles to `client` and `freelancer`
- preserve `fullName` in the returned user payload
- issue a token whose subject matches the returned user id
- add regression coverage for missing `fullName`, admin self-assignment rejection, and successful registration

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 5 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5968, opened 2026-06-09
fix: registration should require fullName and reject admin self-assignment

### LINKED ISSUE #5965: Registration should require fullName and reject admin self-assignment (reissue)

SyntaxHQDEV:
## Bug
Registration schema did not require fullName and allowed admin role self-assignment.

## Fix
- Added fullName as required string field to registerSchema
- Removed "admin" from allowed registration roles
- Returns 400 if fullName is missing

Closes #5965

### DIFF
--- a/apps/api/src/validators/auth.js
+++ b/apps/api/src/validators/auth.js
@@ -1,9 +1,10 @@
 import { z } from "zod";
 
 export const registerSchema = z.object({
+  fullName: z.string().min(1, "fullName is required"),
   email: z.string().email(),
   password: z.string().min(8),
-  role: z.enum(["client", "freelancer", "admin"]).default("client")
+  role: z.enum(["client", "freelancer"]).default("client")
 });
 
 export const loginSchema = z.object({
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5973, opened 2026-06-09
fix: malformed JSON request bodies should return 400

### LINKED ISSUE #5961: Malformed JSON request bodies should return 400 (reissue)

SyntaxHQDEV:
Adds SyntaxError detection to error handler. Closes #5961


SyntaxHQDEV:
Ready. JSON parse errors now return 400.

MikeGarciaAGM (approved):
Looks good to me.

PR #5975, opened 2026-06-09
Return 400 for malformed JSON bodies

### LINKED ISSUE #5961: Malformed JSON request bodies should return 400 (reissue)

poem123123:
Closes #5961.

## Summary
- handle malformed JSON parse failures before the generic 500 handler
- return HTTP 400 with a clear malformed JSON response body
- keep unexpected API errors on the existing 500 path
- add node:test coverage for invalid JSON request bodies

## Validation
- git diff --check

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,16 @@
 export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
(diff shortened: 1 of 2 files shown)


poem123123:
Eligibility reissue created: #5979. This PR implements that issue; please consider it against #5979.

MikeGarciaAGM (approved):
Looks good to me.

PR #5976, opened 2026-06-09
Prevent public admin role registration

### LINKED ISSUE #1742: fix: prevent admin role self-assignment during registration

poem123123:
Closes #1742.

## Summary
- remove admin from the public registration role schema
- return HTTP 400 for invalid public registration payloads instead of issuing a token
- preserve normal client/freelancer registration behavior
- add node:test coverage for rejected admin self-assignment and accepted freelancer registration

## Validation
- git diff --check


poem123123:
Eligibility reissue created: #5980. This PR implements that issue; please consider it against #5980.

MikeGarciaAGM (approved):
Looks good to me.

PR #5983, opened 2026-06-09
[SnowfallHD] Validate search query input

### LINKED ISSUE #5982: Search query should be validated and length-limited (SnowfallHD reissue)
Parent bounty: #743

This is an independent reissue of #5958, created because the original issue is limited to its creator and asks non-authors to create another issue referring to #743.

## Bug

The `GET /api/search` endpoint passes `req.query.q` directly into `globalSearch()` with no type validation, trimming, or length limit. That allows callers to submit repeated query parameters or extremely long query strings and have the API carry them through the search layer unchanged.

## Expected

- Search query input is normalized to a string and trimmed.
- Queries above a bounded maximum length are rejected with a 400 response.
- Repeated/non-string query inputs are rejected with a 400 response.
- Valid queries are passed to the search service in normalized form.
- Tests cover trimming, overly long query rejection, and repeated query rejection.

SnowfallHD:
/claim #743

Closes #5982

## Summary
- Validate `GET /api/search` query input before calling `globalSearch()`.
- Trim valid string queries before passing them to the search service.
- Reject repeated/non-string query parameters and overly long queries with the shared 400 error shape.
- Add focused API regression tests for trim, repeated query rejection, and length-limit rejection.
- Add a short demo video under `docs/demo/search-query-5958-demo.mp4` showing the verified test output.

## Test Plan
- `node --test apps/api/src/tests/*.test.js` ✅ (4 tests passed, 0 failed)
- `git diff --check` ✅

## Demo
- `docs/demo/search-query-5958-demo.mp4`
- `docs/demo/search-query-5958-demo.png`

Note: the existing workspace `npm test` script invokes `node --test src/tests`, which fails under this local Node 22 environment because Node treats the directory as a missing module. I used the direct glob command above to run the repo's existing and new tests without changing package scripts outside this issue's scope.


MikeGarciaAGM (approved):
Validated the search query hardening path. The route now trims valid queries, rejects malformed or oversized input, and includes the expected regression coverage and demo artifacts.

PR #5985, opened 2026-06-09
Validate token before refresh

### LINKED ISSUE #5984: Refresh endpoint should validate refresh token before minting access token (reissue)
Parent bounty: #743
Based on existing issue: #1750

## Bug

## Expected
- The refresh endpoint requires a `token` in the request body.
- Missing tokens return HTTP 400.
- Invalid tokens return HTTP 401.
- A valid token is verified before minting a replacement token.
- The replacement token preserves the decoded `sub` and `role`.
- Tests cover missing, invalid, and valid token refresh requests.

Refs #743.

poem123123:
Closes #5984.

## Summary
- require a token in POST /api/auth/refresh requests
- return 400 for missing refresh tokens
- verify supplied tokens and return 401 for invalid tokens
- mint replacement access tokens from the decoded subject and role instead of a hard-coded user
- add node:test coverage for missing, invalid, and valid refresh requests

## Validation
- git diff --check

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
-import { registerSchema, loginSchema } from "../validators/auth.js";
+import { registerSchema, loginSchema, refreshSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5989, opened 2026-06-09
Require authentication for payments

### LINKED ISSUE #5986: Payment creation should require authentication (reissue)
Parent bounty: #743
Based on existing issue: #2757

## Bug
The `POST /api/payments` route currently calls `createPayment` without applying `authMiddleware`, so unauthenticated callers can create payment intent placeholder responses.

## Expected
- Payment creation requires Bearer authentication before reaching the controller.
- Unauthenticated payment requests return HTTP 401.
- Authenticated payment requests keep the existing response behavior.
- Tests cover unauthenticated rejection and authenticated access.

Refs #743.

poem123123:
Closes #5986.

## Summary
- require Bearer authentication before POST /api/payments reaches the controller
- keep existing payment intent placeholder response behavior for authenticated requests
- add node:test coverage for unauthenticated rejection and authenticated access

## Validation
- git diff --check

### DIFF
--- a/apps/api/src/routes/paymentRoutes.js
+++ b/apps/api/src/routes/paymentRoutes.js
@@ -1,6 +1,7 @@
 import { Router } from "express";
 import { createPayment } from "../controllers/paymentController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const paymentRoutes = Router();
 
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5990, opened 2026-06-09
Validate and limit search query input

### LINKED ISSUE #5963: Search query should be validated and length-limited (reissue)
Parent bounty: #743

This is an independent reissue of #5958, created because the original issue is limited to its creator and asks non-authors to create another issue referring to #743.

## Bug

The `GET /api/search` endpoint passes `req.query.q` directly into `globalSearch()` with no type validation, trimming, or length limit. That allows callers to submit repeated query parameters or extremely long query strings and have the API carry them through the search layer unchanged.

## Expected

- Search query input is normalized to a string and trimmed.
- Queries above a bounded maximum length are rejected with a 400 response.
- Non-string query inputs are rejected with a 400 response.
- Valid queries are passed to the search service in normalized form.
- Tests cover trimming, overly long query rejection, and non-string query rejection.

Kyou2474:
## Description
Validates `GET /api/search` query input before calling the search service.

Closes #5963

## Changes Made
- Added a search query validator that normalizes `q` by trimming whitespace and defaults missing queries to an empty string.
- Rejects non-string query input, including repeated `q` parameters, with a 400 response.
- Rejects queries longer than the bounded maximum length.
- Added API tests for trimming, overly long queries, and repeated query parameters.

## Test

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,13 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchQuerySchema } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const parsed = searchQuerySchema.safeParse(req.query);
+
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #5992, opened 2026-06-09
Require authentication for message creation

### LINKED ISSUE #5991: Message creation should require authentication (reissue)
Parent bounty: #743
Based on existing issue: #5645

## Bug
The `POST /api/messages` route currently calls `postMessage` without applying `authMiddleware`, so unauthenticated callers can create message records.

## Expected
- Message creation requires Bearer authentication before reaching the controller.
- Unauthenticated message creation requests return HTTP 401.
- Authenticated message creation requests keep the existing response behavior.
- Tests cover unauthenticated rejection and authenticated access.

Refs #743.

poem123123:
Closes #5991.

## Summary
- require Bearer authentication before POST /api/messages reaches the controller
- keep existing message creation response behavior for authenticated requests
- add node:test coverage for unauthenticated rejection and authenticated access

## Validation
- git diff --check

### DIFF
--- a/apps/api/src/routes/messageRoutes.js
+++ b/apps/api/src/routes/messageRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getMessages, postMessage } from "../controllers/messageController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const messageRoutes = Router();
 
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #6005, opened 2026-06-09
Return defensive snapshots from list services

### LINKED ISSUE #4530: List service responses should be defensive snapshots (reissue via #743)
Parent Algora bounty: #743
Reference issue: #4524

## Bug

Affected examples:
- `listJobs()` returns `jobs`

## Expected behavior

## Scope
- Add focused regression coverage proving a caller cannot mutate backing stores by mutating a list response.

/attempt #743

Procoder1234556:
## Summary
- return shallow array copies from the in-memory list services
- keep record objects and response shapes unchanged
- add focused regression coverage showing callers cannot mutate backing stores through list responses

## Verification
- 
ode --test src/tests/health.test.js src/tests/list-services.test.js

Closes #4530

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,7 +1,7 @@
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return [...jobs];
 }
 
 export async function createJob(payload) {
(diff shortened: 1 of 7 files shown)


MikeGarciaAGM (approved):
Looks good. Returning defensive snapshots from the list services prevents accidental mutation and keeps the API data shape stable.

PR #6009, opened 2026-06-09
Validate OAuth callback providers

### LINKED ISSUE #5882: OAuth callback should reject unsupported providers

vibhor-aggr:
## Summary

Closes #5882.

Validates OAuth callback provider names before reflecting them in the API response.

## Changes

- Add an OAuth provider whitelist for `github` and `google`.
- Return `400` for unsupported OAuth provider path values.
- Keep supported provider callback behavior unchanged.
- Add API regression tests for unsupported and supported providers.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
-import { registerSchema, loginSchema } from "../validators/auth.js";
+import { loginSchema, oauthProviderSchema, registerSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Restricting OAuth callbacks to supported providers keeps the callback contract explicit, and the tests cover both accepted and rejected providers.

PR #6010, opened 2026-06-09
Return 400 for Zod validation errors

### LINKED ISSUE #5905: Zod validation errors should return 400 responses

vibhor-aggr:
## Summary

Closes #5905.

Returns HTTP 400 responses for Zod validation failures instead of treating them as unexpected server errors.

## Changes

- Add `ZodError` handling in the global API error middleware.
- Return structured validation issues with the existing failure response shape.
- Preserve HTTP 500 behavior for non-validation errors.
- Forward job validation failures to the global error middleware under Express 4.
- Add API regression coverage for an invalid job creation payload.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -6,7 +6,11 @@ export async function getJobs(req, res) {
   return ok(res, await listJobs());
 }
 
-export async function postJob(req, res) {
-  const payload = createJobSchema.parse(req.body);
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good to me.

PR #6012, opened 2026-06-09
Use success envelope for health check

### LINKED ISSUE #5894: Health endpoint should use the shared success envelope

vibhor-aggr:
## Summary

Closes #5894.

Routes the health endpoint through the shared success response helper so it returns the same envelope shape as successful API routes.

## Changes

- Import and use the existing `ok()` helper for `GET /health`.
- Preserve HTTP 200 behavior.
- Update focused health regression coverage for the shared `{ success: true, data: ... }` response envelope.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -14,6 +14,7 @@ import { notificationRoutes } from "./routes/notificationRoutes.js";
 import { uploadRoutes } from "./routes/uploadRoutes.js";
 import { searchRoutes } from "./routes/searchRoutes.js";
 import { adminRoutes } from "./routes/adminRoutes.js";
+import { ok } from "./utils/response.js";
 
 export function createApp() {
   const app = express();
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Using the shared success envelope for health keeps the API contract consistent, and the test verifies the new shape.

PR #6013, opened 2026-06-09
Bypass API rate limit for health checks

### LINKED ISSUE #5880: Health endpoint should bypass shared API rate limiting

vibhor-aggr:
## Summary

Closes #5880.

Keeps health checks outside the shared API rate-limit bucket while preserving rate limiting for API routes.

## Changes

- Register `GET /health` before `apiLimiter`.
- Keep security and JSON parsing middleware before the health route.
- Keep `apiLimiter` active for `/api/*` routes.
- Add regression coverage for repeated health checks staying HTTP 200.
- Add regression coverage for an API route still returning HTTP 429 after the shared limit is exceeded.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -21,11 +21,12 @@ export function createApp() {
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
-  app.use(apiLimiter);
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
   });
+
+  app.use(apiLimiter);
 
   app.use("/api/auth", authRoutes);
   app.use("/api/users", userRoutes);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Keeping health checks out of the shared API limiter preserves observability without weakening the protected API surface, and the tests cover both sides of the split.

PR #6014, opened 2026-06-09
Return 400 for auth validation errors

### LINKED ISSUE #5908: Auth validation errors should return 400 responses

vibhor-aggr:
## Summary

Closes #5908.

Returns HTTP 400 responses for auth request validation failures instead of letting Zod errors bubble as unexpected server errors.

## Changes

- Add `ZodError` handling in the global API error middleware.
- Return structured validation issues with the existing failure response shape.
- Forward register and login validation errors to the global middleware under Express 4.
- Preserve successful auth response behavior.
- Add API regression coverage for invalid and valid registration payloads.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -2,16 +2,24 @@ import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
 import { ok } from "../utils/response.js";
 
-export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
-  return ok(res, result, 201);
+export async function register(req, res, next) {
+  try {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Returning 400 for auth validation errors is the right contract, and the tests cover both invalid and valid payloads.

PR #6020, opened 2026-06-09
Sanitize upload response filenames

### LINKED ISSUE #5859: Upload response should sanitize original filenames

vibhor-aggr:
## Summary

Closes #5859.

Sanitizes upload response filename metadata before returning it to clients.

## Changes

- Add upload filename metadata sanitization.
- Strip path separators to the final filename segment.
- Remove control characters, trim whitespace, and cap returned filename length.
- Keep upload storage behavior and file contents unchanged.
- Add focused regression coverage for path-like and control-character filenames.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,25 @@
 import { ok } from "../utils/response.js";
 
+const MAX_UPLOAD_FILENAME_LENGTH = 120;
+
+export function sanitizeUploadFilename(filename) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Sanitizing upload filenames is a nice low-risk safety fix, and the test covers path, control, and length handling.

PR #6021, opened 2026-06-09
Validate refresh token input

### LINKED ISSUE #5703: Refresh endpoint should require and validate a token

vibhor-aggr:
## Summary

Closes #5703.

Requires `POST /api/auth/refresh` to receive and validate a supplied token before issuing a replacement access token.

## Changes

- Verify the supplied token in the refresh service.
- Preserve decoded `sub` and `role` claims in the replacement token.
- Add API tests for missing, invalid, and valid refresh token requests.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
-import { registerSchema, loginSchema } from "../validators/auth.js";
+import { loginSchema, refreshSchema, registerSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. Refresh token validation is a useful guardrail, and the tests cover missing, invalid, and valid refresh flows.

PR #6022, opened 2026-06-09
Validate user creation input

### LINKED ISSUE #5709: User creation should validate input and ignore system fields
## Bug

`postUser` passes `req.body` directly to `createUser` without validation:

```js
}
```

The `createUser` service then spreads the raw payload directly into the stored user object:

```js
  users.push(user);
  return user;
}
```

This allows callers to create users without required fields and inject arbitrary/system fields, including overriding the generated `id` or adding role/admin-like properties.

## Expected behavior

- A Zod schema should validate that `email` is a valid email and `name` is a non-empty string.
- Invalid request bodies should return HTTP 400 with structured validation issues.
- Caller-supplied system fields such as `id` and `role` should not be persisted or returned.
- The service should construct the user object from allowlisted fields only.

Parent bounty: #743

/bounty $780

vibhor-aggr:
## Summary

Closes #5709.

Validates `POST /api/users` payloads before storing user records and keeps the generated user id server-owned.

## Changes

- Add a strict user creation schema requiring `name` and valid `email`.
- Restrict public user roles to `client` and `freelancer`, defaulting omitted role to `client`.
- Reject unknown fields such as caller-supplied `id`.
- Apply the generated user id after validated payload fields in the user service.
- Add API tests for invalid and valid user creation payloads.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
+  const result = createUserSchema.safeParse(req.body);
+  if (!result.success) {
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. User creation validation plus server-owned ids makes the API safer, and the tests show both rejection and happy-path behavior.

PR #6023, opened 2026-06-09
Require admin role for admin routes

### LINKED ISSUE #5689: Bug: Missing admin role authorization check on admin metrics endpoint

vibhor-aggr:
## Summary

Closes #5689.

Restricts `/api/admin` routes to authenticated users with the `admin` role.

## Changes

- Add `adminMiddleware` to enforce admin-role authorization.
- Apply the admin role check after token authentication on admin routes.
- Add API tests for unauthenticated, non-admin, and admin access to admin metrics.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,11 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function adminMiddleware(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Forbidden", 403);
+  }
+
+  return next();
+}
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Requiring admin role for admin routes is the correct boundary, and the tests verify unauthenticated, client, and admin access.

PR #6024, opened 2026-06-09
Validate message creation input

### LINKED ISSUE #5714: Message creation should validate input and ignore system fields
## Bug

`postMessage` passes `req.body` directly to `sendMessage` without schema validation:

```js
export async function postMessage(req, res) {
  return ok(res, await sendMessage(req.body), 201);
}
```

The service then spreads caller input into the message object:

```js
```

This allows empty or malformed messages to reach the service layer and lets callers override system-managed fields like `id` or `sentAt`.

## Expected behavior

- A Zod `createMessageSchema` validates `content`, `recipientId`, and `jobId` as non-empty strings before `sendMessage` is called.
- Invalid payloads return HTTP 400 with structured validation issues.
- Valid payloads still succeed.
- System-managed fields such as `id` and `sentAt` are generated by the service, not accepted from the request body.

Parent bounty: #743

/bounty $780

vibhor-aggr:
## Summary

Closes #5714.

Validates message creation payloads before storing messages and keeps system-managed fields server-owned.

## Changes

- Add a focused message creation schema for `content`, `recipientId`, and `jobId`.
- Return HTTP 400 with structured validation issues for invalid message payloads.
- Pass only validated message fields into the message service.
- Keep generated `id` and `sentAt` values server-owned.
- Add API coverage for invalid payloads and ignored system fields.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/messageController.js
+++ b/apps/api/src/controllers/messageController.js
@@ -1,10 +1,21 @@
 import { ok } from "../utils/response.js";
 import { listMessages, sendMessage } from "../services/messageService.js";
+import { createMessageSchema } from "../validators/message.js";
 
 export async function getMessages(req, res) {
   return ok(res, await listMessages());
 }
 
 export async function postMessage(req, res) {
-  return ok(res, await sendMessage(req.body), 201);
+  const result = createMessageSchema.safeParse(req.body);
+
+  if (!result.success) {
+    return res.status(400).json({
+      success: false,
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Looks good. Message payload validation plus server-owned ids is the right combination here, and the tests cover both invalid and mutated input.

PR #6025, opened 2026-06-09
Align registration token subject with returned user id

vibhor-aggr:
## Summary

Closes #2768.

Ensures registration generates the user id once and signs the JWT access token with that same value as the `sub` claim.

## Changes

- Store the generated registration user id in a local `userId`.
- Return `userId` in the registration response.
- Sign the access token with `{ sub: userId, role: payload.role }`.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743


MikeGarciaAGM (approved):
Looks good. Aligning the token subject with the returned user id removes a subtle mismatch, and the test covers the exact contract.

PR #6027, opened 2026-06-09
Restrict registration to public roles

### LINKED ISSUE #5890: Public registration should reject admin role assignment

vibhor-aggr:
## Summary

Closes #5890.

Prevents public registration from accepting `role: "admin"` while preserving client/freelancer registration and the default client role.

## Changes

- Restrict `registerSchema.role` to `client` and `freelancer`.
- Preserve the omitted-role default of `client`.
- Add focused validation tests for rejecting admin, accepting public roles, and defaulting omitted role.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/tests/registrationRole.test.js
+++ b/apps/api/src/tests/registrationRole.test.js
@@ -1,0 +1,26 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { registerSchema } from "../validators/auth.js";
+
+const baseRegistration = {
+  email: "user@example.com",
+  password: "password123"
+};
+
+test("registration rejects admin role self-assignment", () => {
+  assert.throws(() => {
+    registerSchema.parse({
+      ...baseRegistration,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Restricting registration to public roles is a sensible boundary and the tests show admin self-assignment is rejected.

PR #6028, opened 2026-06-09
Preserve server-owned notification fields

vibhor-aggr:
## Summary

Closes #2762.

Ensures notification creation always returns a server-generated id and starts new notifications with `read: false`, even if callers include `id` or `read` in the request payload.

## Changes

- Apply caller payload fields before server-owned notification fields.
- Preserve non-server payload fields such as `userId`, `message`, and `type`.
- Add focused service coverage proving caller-supplied `id` and `read` are ignored during creation.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,7 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Preserving server-owned notification fields while allowing normal payload data is a sensible hardening change, and the test covers the override cases.

MikeGarciaAGM (approved):
Looks good. Preserving server-owned notification fields closes the override path without changing the notification payload shape.

PR #6030, opened 2026-06-09
Keep proposal ids server-owned

### LINKED ISSUE #5999: Proposal creation should preserve a server-generated id
Parent bounty: #743

## Bug

`apps/api/src/services/proposalService.js` currently creates proposal records with `{ id: generated, ...payload }`, which lets a caller override the generated proposal id by sending an `id` field in the request payload.

That means proposal identifiers are not reliably server-owned and can collide with reserved or previously generated ids.

## Expected

- Proposal creation should always return a server-generated id.
- Caller-supplied `id` fields should be ignored during proposal creation.
- Existing payload fields unrelated to the server-owned id should still be preserved.
- Add focused service coverage proving hostile payload ids are ignored.

## Scope

Keep the fix limited to proposal creation behavior and focused automated coverage.

vibhor-aggr:
## Summary

Closes #5999.

Keeps proposal identifiers server-owned by assigning the generated id after caller payload fields are spread.

## Changes

- Change `createProposal()` so a caller-supplied `id` cannot override the generated `prp_<timestamp>` value.
- Add service regression coverage proving caller-controlled proposal ids are ignored.
- Keep unrelated proposal payload fields unchanged.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { ...payload, id: `prp_${Date.now()}` };
   proposals.push(proposal);
   return proposal;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Keeping proposal ids server-owned is a clean contract hardening, and the regression test covers caller override attempts.

MikeGarciaAGM (approved):
Looks good. Keeping proposal ids server-owned is the right default and the test clearly covers the spoofing attempt.

PR #6031, opened 2026-06-09
Rate limit malformed JSON before parsing

### LINKED ISSUE #6000: Rate limiter should count malformed JSON before body parsing
Parent bounty: #743

## Description

The global API rate limiter is currently registered after `express.json()` in `apps/api/src/app.js`. Malformed JSON requests are parsed and rejected by the body parser/error path before the limiter sees the request, so repeated invalid JSON can consume parser/error-handler work without being counted toward the quota.

## Reproduction

Send more than 200 malformed JSON requests to an API route such as `POST /api/jobs`:

```bash
for i in $(seq 1 205); do
  curl -s -o /dev/null -w "%{http_code}\n" \
    -X POST http://localhost:4000/api/jobs \
    -d "{bad json"
done
```

Expected: after the configured quota is exhausted, the API returns `429 Too Many Requests`.

Actual: malformed JSON is parsed before the limiter runs, so these requests keep hitting body-parser/error handling instead of being rate-limited.

## Proposed fix

Move the API limiter before `express.json()` so rate-limit accounting happens before JSON body parsing, and add a focused regression test proving repeated malformed JSON requests eventually receive `429`.

vibhor-aggr:
## Summary

Closes #6000.

Moves API rate limiting before JSON body parsing so malformed JSON requests count toward the configured quota before parser work is performed.

## Changes

- Register the API limiter before `express.json()` in `createApp()`.
- Add a `createApiLimiter()` factory so app instances get independent limiter state.
- Preserve the existing `apiLimiter` export for compatibility.
- Add regression coverage proving repeated malformed JSON requests eventually receive `429`.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -1,7 +1,7 @@
 import cors from "cors";
 import express from "express";
 import helmet from "helmet";
-import { apiLimiter } from "./middleware/rateLimit.js";
+import { createApiLimiter } from "./middleware/rateLimit.js";
 import { errorHandler } from "./middleware/errorHandler.js";
 import { authRoutes } from "./routes/authRoutes.js";
 import { userRoutes } from "./routes/userRoutes.js";
@@ -20,8 +20,8 @@ export function createApp() {
 
   app.use(helmet());
   app.use(cors());
+  app.use(createApiLimiter());
   app.use(express.json());
-  app.use(apiLimiter);
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Looks good. Moving rate limiting ahead of JSON parsing is a practical hardening change, and the regression test shows malformed requests now count toward quota as intended.

MikeGarciaAGM (approved):
Looks good. Rate limiting malformed JSON before parsing is a sensible guardrail, and the test proves the abusive case reaches the limiter.

PR #6033, opened 2026-06-09

vibhor-aggr:
## Summary

Closes #2853.

Rejects inverted job budget ranges during job creation and partial job updates when both budget fields are present.

## Changes

- Add a shared Zod budget range refinement.
- Apply the refinement to `createJobSchema`.
- Apply the same refinement to `updateJobSchema` after partial schema conversion.
- Add focused coverage for invalid create, valid create, invalid update, and valid partial update cases.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/tests/jobBudgetValidation.test.js
+++ b/apps/api/src/tests/jobBudgetValidation.test.js
@@ -1,0 +1,45 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
+
+const validJob = {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Rejecting inverted budget ranges at schema level keeps the job contract consistent, and the regression tests cover both create and update flows.

MikeGarciaAGM (approved):
Looks good. Preventing inverted budget ranges is a clean validation rule, and the tests exercise both create and update paths.

PR #6034, opened 2026-06-09
Validate search query input

### LINKED ISSUE #5888: Search endpoint should validate query input and cap length

vibhor-aggr:
## Summary

Closes #5888.

Validates the `GET /api/search` query parameter before it reaches the search service.

## Changes

- Preserve omitted `q` as an empty search.
- Reject repeated or otherwise non-string `q` values.
- Strip ASCII control characters and trim the query.
- Reject queries longer than 200 characters.
- Add focused API coverage for valid sanitization, omitted query, repeated query parameters, and length rejection.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,30 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
 
+const MAX_QUERY_LENGTH = 200;
+
+function normalizeSearchQuery(value) {
+  if (value === undefined) {
+    return { query: "" };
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Validating and trimming search input before it reaches the service is a sensible guard, and the tests cover the expected sanitization and rejection paths.

MikeGarciaAGM (approved):
Looks good. Search query normalization and length bounds are well targeted, and the tests cover sanitization, omission, and invalid shapes.

PR #6041, opened 2026-06-09
Return JSON for unknown API routes

### LINKED ISSUE #5725: Unknown API routes should return JSON 404 responses

vibhor-aggr:
## Summary

Closes #5725.

Adds a final not-found middleware after all mounted routes so unknown API paths return the same JSON failure contract as other API errors.

## Changes

- Import and use the existing `fail()` response helper for unmatched routes.
- Return HTTP 404 with `{ success: false, message: "Not found" }` for unknown paths.
- Preserve existing successful routes such as `/health`.
- Add a focused regression test for an unknown API route.

## Verification

```bash
git diff --check
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -14,6 +14,7 @@ import { notificationRoutes } from "./routes/notificationRoutes.js";
 import { uploadRoutes } from "./routes/uploadRoutes.js";
 import { searchRoutes } from "./routes/searchRoutes.js";
 import { adminRoutes } from "./routes/adminRoutes.js";
+import { fail } from "./utils/response.js";
 
 export function createApp() {
   const app = express();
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Returning JSON for unknown routes keeps the API contract consistent, and the test covers the not-found path cleanly.

MikeGarciaAGM (approved):
Looks good. Returning JSON for unknown routes improves consistency without changing the route surface, and the test captures the contract.

PR #6042, opened 2026-06-09
Reject empty upload requests

### LINKED ISSUE #3668: Upload endpoint should reject empty submissions (reissue #3470)
Parent bounty: #743

Reissue of #3470. Reference: https://github.com/SecureBananaLabs/bug-bounty/issues/3470

## Bug
The upload endpoint returns 201 with `no-file` status when no file is provided. It should return 400.

## Expected
When no file is provided, return 400 with `"File is required"`.

## Fix
Add early return with `fail()` in `uploadController.js` when `req.file` is missing.

/attempt #743

vibhor-aggr:
## Summary

Closes #3668.

Rejects empty upload requests instead of returning a successful `no-file` response.

## Changes

- Import and use the existing `fail()` response helper in `uploadController.js`.
- Return HTTP 400 with `File is required` when Multer does not provide `req.file`.
- Preserve the existing successful upload response shape for valid file submissions.
- Add API regression coverage for empty and valid upload requests.

## Verification

```bash
git diff --check
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
-    status: req.file ? "uploaded" : "no-file"
+    filename: req.file.originalname,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Rejecting empty uploads with a 400 keeps the API contract explicit, and the regression tests cover both empty and valid requests.

MikeGarciaAGM (approved):
Looks good. Rejecting empty upload requests is a crisp correctness fix and the tests verify both the failure and success path.

PR #6047, opened 2026-06-09
Keep message records server-owned

### LINKED ISSUE #3464: Keep message ids and records server-owned
Reissue of #3429 for eligibility under #743.\n\nMessage records should keep server-owned ids and stored records under service control. listMessages() and sendMessage() should not expose mutable backing records.\n\nExpected fix:\n- keep generated message ids server-owned\n- return defensive snapshots from listMessages()\n- return a snapshot from sendMessage() instead of the stored object reference\n- add focused service-level regression coverage\n\nParent bounty reference: #743

vibhor-aggr:
## Summary

Closes #3464.

Keeps message ids and stored records under service control by returning defensive snapshots from the message service.

## Changes

- Generate message ids after caller payload fields so request payloads cannot override them.
- Return shallow defensive copies from `listMessages()`.
- Return a shallow defensive copy from `sendMessage()` instead of the stored object reference.
- Add focused service-level regression coverage for server-owned ids, array snapshots, listed record snapshots, and returned record snapshots.

## Verification

```bash
git diff --check
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,11 +1,11 @@
 const messages = [];
 
 export async function listMessages() {
-  return messages;
+  return messages.map((message) => ({ ...message }));
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const message = { ...payload, id: `msg_${Date.now()}`, sentAt: new Date().toISOString() };
   messages.push(message);
-  return message;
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Returning defensive copies keeps message records server-owned without changing the visible API behavior.

MikeGarciaAGM (approved):
Looks good. Making message records server-owned and returning defensive copies closes the mutation gap cleanly, and the tests cover both behaviors.

PR #6050, opened 2026-06-09
Keep review timestamps server-owned

### LINKED ISSUE #3456: Keep review creation timestamps server-owned
Reissue of #3450 for eligibility under #743.\n\nThe review service should own the creation timestamp for newly-created reviews. If createdAt is accepted from the request payload, clients can backdate or future-date reviews and make audit/order data unreliable.\n\nExpected fix:\n- ignore caller-supplied createdAt values during review creation\n- assign createdAt in the service at creation time\n- preserve the rest of the review payload\n- add regression coverage proving the client timestamp is not stored\n\nParent bounty reference: #743

vibhor-aggr:
## Summary

Closes #3456.

Keeps review creation timestamps under service control so callers cannot backdate or future-date reviews through `createdAt` in the request payload.

## Changes

- Apply caller payload fields before generated service-owned fields in `createReview()`.
- Assign `createdAt` in the service with the current timestamp.
- Preserve accepted review payload fields.
- Add focused service-level regression coverage proving caller-supplied `createdAt` is not stored.

## Verification

```bash
git diff --check
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,7 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
-  const review = { id: `rev_${Date.now()}`, ...payload };
+  const review = { ...payload, id: `rev_${Date.now()}`, createdAt: new Date().toISOString() };
   reviews.push(review);
   return review;
 }
--- a/apps/api/src/tests/reviewServiceCreatedAt.test.js
+++ b/apps/api/src/tests/reviewServiceCreatedAt.test.js
@@ -1,0 +1,21 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createReview } from "../services/reviewService.js";
+
+test("createReview keeps createdAt server-owned", async () => {
+  const review = await createReview({
+    reviewerId: "usr_reviewer",
+    revieweeId: "usr_reviewee",
+    rating: 5,
+    comment: "Excellent work.",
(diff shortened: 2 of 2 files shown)


MikeGarciaAGM (changes requested):
Requesting changes: this patch still leaves `createdAt` caller-controlled in the stored object because the spread order is `id, ...payload`-style logic in the existing service pattern. If `createReview()` continues to take arbitrary payloads from the controller, then moving the timestamp assignment alone is not enough unless the server-owned field is applied after the payload and the incoming `createdAt` is ignored.

Please make sure the service assigns `createdAt` after spreading caller fields, and add a regression test that proves a caller-supplied `createdAt` cannot be persisted.

MikeGarciaAGM (approved):
Re-read: this is fine. `createdAt` is assigned after the payload spread, so caller-supplied timestamps are not persisted. My earlier request-changes note on this PR was a misread; there is no blocker here.

MikeGarciaAGM:
Status refresh: PR #6050 is approved and mergeable. The change stays narrowly scoped to server-owned review timestamps, and the latest approval confirms the caller-supplied timestamp is not persisted. No remaining blockers on my side; this looks ready for maintainer merge when convenient.

MikeGarciaAGM (approved):
Looks good. Making review timestamps server-owned closes a straightforward spoofing path and the tests capture the ownership rule.

Issue #4590, opened 2026-06-05
User listing endpoint lacks authentication middleware


jojosenthusiast:
Parent bounty: #743

## Problem

`GET /api/users` lists all users without requiring authentication. The route also allows `POST /api/users` to create users without a bearer token. The user listing exposure is the primary concern — unauthenticated callers can enumerate the entire user base.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

## Evidence

`apps/api/src/routes/userRoutes.js` has no auth guard:
```js
userRoutes.get("/", getUsers);
userRoutes.post("/", postUser);
```

`apps/api/src/routes/adminRoutes.js` shows the correct pattern: `adminRoutes.use(authMiddleware);`

## Expected behavior

At minimum, `GET /api/users` should require a valid bearer token to prevent unauthenticated user enumeration.

PR #6052, opened 2026-06-09
Require authentication for user routes

### LINKED ISSUE #6051: User listing endpoint should require authentication
Parent bounty: #743
Original issue: #4590

## Problem

`GET /api/users` lists all users without requiring authentication. The route also allows `POST /api/users` to create users without a bearer token. The user listing exposure is the primary concern because unauthenticated callers can enumerate the user base.

## Evidence

`apps/api/src/routes/userRoutes.js` has no auth guard:

```js
userRoutes.get("/", getUsers);
userRoutes.post("/", postUser);
```

`apps/api/src/routes/adminRoutes.js` shows the route-level auth pattern.

## Expected behavior

- `GET /api/users` requires a valid bearer token.
- Unauthenticated callers receive HTTP 401.
- Invalid bearer tokens receive HTTP 401.
- Valid bearer tokens keep the existing user listing response behavior.

vibhor-aggr:
## Summary

Closes #6051.

Requires bearer authentication on `/api/users` so unauthenticated callers cannot enumerate user records.

## Changes

- Apply the existing `authMiddleware` to the user routes.
- Preserve the existing user listing behavior for valid bearer tokens.
- Reject missing and invalid bearer tokens with HTTP 401.
- Add focused API regression tests for the protected user routes.

## Verification

```bash
git diff --check
```

Original issue: #4590
Parent bounty: #743

### DIFF
--- a/apps/api/src/routes/userRoutes.js
+++ b/apps/api/src/routes/userRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getUsers, postUser } from "../controllers/userController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const userRoutes = Router();
 
+userRoutes.use(authMiddleware);
 userRoutes.get("/", getUsers);
 userRoutes.post("/", postUser);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. This protects the user routes with the existing bearer auth middleware and the tests cover the expected 401/200 behavior cleanly.

MikeGarciaAGM (approved):
Looks good. Protecting the user routes with auth is the right default, and the tests verify anonymous, invalid, and valid access paths.

Issue #4592, opened 2026-06-05
Search endpoint accessible without authentication


jojosenthusiast:
Parent bounty: #743

## Problem

`GET /api/search` performs a global search across all data without requiring a bearer token. This allows unauthenticated callers to query user, job, and other records the application manages.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

## Evidence

`apps/api/src/routes/searchRoutes.js` has no auth guard:
```js
searchRoutes.get("/", search);
```

`apps/api/src/routes/adminRoutes.js` shows the correct pattern: `adminRoutes.use(authMiddleware);`

## Expected behavior

`GET /api/search` should require a valid bearer token before returning search results.

PR #6058, opened 2026-06-09
Require authentication for search route

### LINKED ISSUE #6057: Search endpoint should require authentication
Parent bounty: #743
Original issue: #4592

## Problem

`GET /api/search` performs a global search across application data without requiring a bearer token. This allows unauthenticated callers to query user, job, freelancer, and other records managed by the API.

## Evidence

`apps/api/src/routes/searchRoutes.js` has no auth guard:

```js
searchRoutes.get("/", search);
```

`apps/api/src/routes/adminRoutes.js` shows the route-level auth pattern.

## Expected behavior

- `GET /api/search` requires a valid bearer token before returning search results.
- Missing bearer credentials return HTTP 401.
- Invalid bearer credentials return HTTP 401.
- Valid bearer credentials preserve the existing search response behavior.

vibhor-aggr:
## Summary

Closes #6057.

Requires bearer authentication before `GET /api/search` returns search results.

## Changes

- Apply the existing `authMiddleware` to the search router.
- Preserve the existing search response for valid bearer tokens.
- Reject missing and invalid bearer tokens with HTTP 401.
- Add focused route-level regression coverage for missing, invalid, and valid bearer credentials.

## Verification

```bash
git diff --check
```

Original issue: #4592
Parent bounty: #743

### DIFF
--- a/apps/api/src/routes/searchRoutes.js
+++ b/apps/api/src/routes/searchRoutes.js
@@ -1,6 +1,8 @@
 import { Router } from "express";
 import { search } from "../controllers/searchController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const searchRoutes = Router();
 
+searchRoutes.use(authMiddleware);
 searchRoutes.get("/", search);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Requiring auth on search matches the route's sensitivity, and the tests cover unauthenticated, invalid, and valid access.

PR #6060, opened 2026-06-09
Validate auth token identity claims

### LINKED ISSUE #6059: Auth middleware should reject JWTs missing identity claims
Parent bounty: #743
Original issue: #3421

## Bug

`authMiddleware` accepts any successfully verified JWT payload as `req.user`. A token signed with the app secret can contain a string payload, an object without `sub`, or an object with an unsupported `role`, and the middleware will still treat the request as authenticated.

That leaves protected routes without a guaranteed requester identity or role and can make downstream authorization checks fail open or behave inconsistently.

## Expected behavior

- Reject decoded JWT payloads that are not plain objects.
- Reject tokens missing a non-empty string `sub`.
- Reject tokens missing an allowed `role` value.
- Preserve valid app-issued access tokens.
- Keep the fix limited to auth middleware identity-claim validation and focused regression tests.

vibhor-aggr:
## Summary

Closes #6059.

Adds identity-claim validation after JWT verification so protected routes only accept app-shaped bearer tokens.

## Changes

- Reject verified JWT payloads that are not plain objects.
- Reject JWTs missing a non-empty string `sub`.
- Reject JWTs with unsupported `role` values.
- Preserve valid app-issued access tokens.
- Add focused route-level regression tests through `/api/admin/metrics`.

## Verification

```bash
git diff --check
```

Original issue: #3421
Parent bounty: #743

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -1,6 +1,19 @@
 import { fail } from "../utils/response.js";
 import { verifyAccessToken } from "../utils/jwt.js";
 
+const allowedRoles = new Set(["client", "freelancer", "admin"]);
+
+function isValidUserClaims(payload) {
+  return (
+    payload &&
+    typeof payload === "object" &&
+    !Array.isArray(payload) &&
+    typeof payload.sub === "string" &&
+    payload.sub.trim() !== "" &&
+    allowedRoles.has(payload.role)
+  );
+}
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Validating token identity claims closes an obvious trust gap, and the tests cover malformed and unsupported claim shapes.

PR #6061, opened 2026-06-09
fix: point package entrypoint to distribution targets (closes #6049)

### LINKED ISSUE #6049: @freelanceflow/ui package entrypoint should be directly importable (reissue via #743)
The @freelanceflow/ui workspace package currently declares main as src/index.ts, and that file re-exports ./Button and ./Card without emitted JavaScript entrypoints.

This makes the package fail when imported directly by a Node/ESM consumer from the workspace:

node -e "import('@freelanceflow/ui')"
It currently errors because the package entrypoint points at TypeScript source and the component modules are not resolvable as runtime JavaScript. The package should expose a real JavaScript entrypoint while preserving TypeScript declarations for consumers.

adesinamichael669-beep:
Closes #6049

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+   "test": "node --test src/tests/**/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Publishing the package export map for dist targets is a clean packaging fix, and the change is narrowly scoped.

MikeGarciaAGM (approved):
Looks good to me.

PR #6063, opened 2026-06-09
Restrict API CORS origin

### LINKED ISSUE #6062: Restrict API CORS to configured origin allowlist
Parent bounty: #743
Original issue: #3879

## Security: Unrestricted CORS Policy

The Express app uses `cors()` with no configuration, allowing requests from any origin. In a production freelance platform, this should be restricted to trusted domains.

Current code in `apps/api/src/app.js`:

```js
app.use(cors());
```

## Expected behavior

- Configure CORS with an explicit allowed origin.
- Read the allowed origin from `CORS_ORIGIN`.
- Default to `http://localhost:3000` for local development.
- Preserve non-browser requests without an `Origin` header.
- Do not emit `Access-Control-Allow-Origin` for untrusted origins.

vibhor-aggr:
## Summary

Closes #6062.

Restricts API CORS responses to a configured allowed origin instead of allowing every origin.

## Changes

- Configure `cors()` with an origin allowlist.
- Read the allowed origin from `CORS_ORIGIN`, defaulting to `http://localhost:3000`.
- Preserve non-browser requests that do not include an `Origin` header.
- Add route-level tests for default allowed origin, untrusted origin rejection, and configured origin support.

## Verification

```bash
git diff --check
```

Original issue: #3879
Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -15,11 +15,25 @@ import { uploadRoutes } from "./routes/uploadRoutes.js";
 import { searchRoutes } from "./routes/searchRoutes.js";
 import { adminRoutes } from "./routes/adminRoutes.js";
 
+function getCorsOptions() {
+  const allowedOrigin = process.env.CORS_ORIGIN ?? "http://localhost:3000";
+
+  return {
+    origin(origin, callback) {
+      if (!origin || origin === allowedOrigin) {
+        return callback(null, true);
+      }
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. The origin allowlist is explicit, keeps localhost/dev workable, and the new tests cover allowed and disallowed origins.

Issue #3316, opened 2026-06-01
Payment intents should include createdAt timestamps


EvidentLed:
Parent bounty: #743

## Bug

`apps/api/src/services/paymentService.js` creates payment intent response objects without a creation timestamp, even though `packages/db/prisma/schema.prisma` defines `Payment.createdAt` with `@default(now())`.

That makes the placeholder payment response inconsistent with the payment model and prevents API consumers from reliably auditing or ordering newly created payment intents while the stub service is still in use.

## Expected

- `createPaymentIntent` should assign one server-owned ISO `createdAt` timestamp when a payment intent is created.
- The returned payment intent response should include that `createdAt` timestamp.
- A caller-supplied `createdAt` field should not control the returned timestamp.
- Keep the scope limited to `paymentService`; do not add Stripe integration or change authentication, currency, status, or unrelated payment behavior.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6065, opened 2026-06-09
Add payment intent creation timestamps

### LINKED ISSUE #6064: Payment intents should include createdAt timestamps
Parent bounty: #743
Original issue: #3316

## Bug

`apps/api/src/services/paymentService.js` creates payment intent response objects without a creation timestamp, even though `packages/db/prisma/schema.prisma` defines `Payment.createdAt` with `@default(now())`.

That makes the placeholder payment response inconsistent with the payment model and prevents API consumers from reliably auditing or ordering newly created payment intents while the stub service is still in use.

## Expected behavior

- `createPaymentIntent` assigns one server-owned ISO `createdAt` timestamp when a payment intent is created.
- The returned payment intent response includes that `createdAt` timestamp.
- A caller-supplied `createdAt` field does not control the returned timestamp.
- Scope remains limited to `paymentService`; no Stripe integration or unrelated behavior changes.

vibhor-aggr:
## Summary

Closes #6064.

Adds a server-owned creation timestamp to payment intent responses.

## Changes

- Add `createdAt` to `createPaymentIntent()` responses using the service creation time.
- Keep caller-supplied `createdAt` from controlling the returned timestamp.
- Preserve existing amount, currency, provider, and payment id behavior.

## Verification

```bash
git diff --check
```

Original issue: #3316
Parent bounty: #743

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -4,6 +4,7 @@ export async function createPaymentIntent(payload) {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
     currency: payload.currency ?? "usd",
-    provider: "stripe"
+    provider: "stripe",
+    createdAt: new Date().toISOString()
   };
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Adding a server-owned createdAt timestamp to payment intents is a sensible ownership fix, and the test covers the overwrite attempt.

Issue #3313, opened 2026-06-01
Job creation should include createdAt timestamps


TheGreatAion:
Parent bounty: #743

## Bug
`apps/api/src/services/jobService.js` creates in-memory job records without a `createdAt` timestamp, even though the Prisma `Job` model defines a server-owned `createdAt` field. Callers therefore receive job records that are missing their creation time.

## Expected behavior
`createJob` should add a server-owned ISO `createdAt` timestamp when it creates a job, return that timestamp to the caller, and store the same value in the in-memory record.

## Proposed fix
Add a focused service regression test proving the timestamp exists, is a valid ISO value, cannot be overridden by caller input, and is retained by `listJobs`. Keep the implementation limited to the job service and focused test.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6068, opened 2026-06-09
Add job creation timestamps

### LINKED ISSUE #6067: Job creation should include createdAt timestamps
Parent bounty: #743
Original issue: #3313

## Bug

`apps/api/src/services/jobService.js` creates in-memory job records without a `createdAt` timestamp, even though the Prisma `Job` model defines a server-owned `createdAt` field. Callers therefore receive job records that are missing their creation time.

## Expected behavior

- `createJob` assigns a server-owned ISO `createdAt` timestamp when creating a job.
- The returned job response includes that timestamp.
- The stored in-memory job retains the same timestamp returned to the caller.
- Caller-supplied `createdAt` input does not control the timestamp.
- Scope remains limited to the job service and focused regression coverage.

vibhor-aggr:
## Summary

Closes #6067.

Adds a server-owned creation timestamp to job records produced by the in-memory job service.

## Changes

- Add `createdAt` to `createJob()` records using the service creation time.
- Keep caller-supplied `createdAt` from controlling the stored or returned timestamp.
- Preserve existing job payload fields.
- Add focused service-level regression coverage proving the returned and stored timestamps match.

## Verification

```bash
git diff --check
```

Original issue: #3313
Parent bounty: #743

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,7 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const job = { id: `job_${Date.now()}`, status: "open", ...payload, createdAt: new Date().toISOString() };
   jobs.push(job);
   return job;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good to me.

Issue #3311, opened 2026-06-01
User creation should include createdAt timestamps


EvidentLed:
Parent bounty: #743

## Bug

`apps/api/src/services/userService.js` creates in-memory user records without a `createdAt` timestamp, even though `packages/db/prisma/schema.prisma` defines `User.createdAt` with `@default(now())`.

That makes the current service response inconsistent with the database model and prevents API consumers from reliably sorting or auditing newly created users by creation time while the placeholder service is still in use.

## Expected

- `createUser` should assign one server-owned ISO `createdAt` timestamp when a user is created.
- The returned user record should include that `createdAt` timestamp.
- The record stored by the in-memory service should include the same `createdAt` timestamp.
- Keep the scope limited to `userService`; do not change auth registration or unrelated create services.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6071, opened 2026-06-09
Add user creation timestamps

vibhor-aggr:
Fixes duplicate issue #6070.

Related issues:
- Parent bounty route: #743
- Original issue: #3311

Changes:
- adds a server-owned ISO `createdAt` timestamp to `createUser()` records;

Verification:
- `git diff --check`


MikeGarciaAGM (approved):
Looks good to me.

Issue #6076, opened 2026-06-09
Notification creation should include createdAt timestamps


vibhor-aggr:
Parent bounty route: #743
Original issue: #3305

This duplicate is opened for a PR submission and refers back to the parent bounty route as requested by the issue instructions.

Fix scope:
- add a server-owned ISO `createdAt` timestamp to `createNotification()` records;
- ensure caller-supplied `createdAt` values do not control the returned timestamp;
- add focused service-level regression coverage.

Verification:
- `node --test apps/api/src/tests/*.test.js`
- `git diff --check`

Issue #3305, opened 2026-06-01
Notification creation should include createdAt timestamps


wangedmund77-cmyk:
Parent bounty: #743

## Bug
`createNotification()` stores and returns notifications without a `createdAt` timestamp. The Prisma `Notification` model includes `createdAt`, and notification feeds need a server creation time for ordering.

## Expected
New notification records should include a server-generated ISO `createdAt` timestamp. Caller-supplied timestamps should not replace the server value.

## Proposed fix
Add a server-side `createdAt` value during notification creation and cover it with a focused service-level regression test.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6077, opened 2026-06-09
Add notification creation timestamps

vibhor-aggr:
Fixes duplicate issue #6076.

Related issues:
- Parent bounty route: #743
- Original issue: #3305

Changes:
- adds a server-owned ISO `createdAt` timestamp to `createNotification()` records;

Verification:
- `git diff --check`


MikeGarciaAGM (approved):
Validated the notification timestamp addition. Server-owned createdAt metadata is now preserved consistently and the change stays within the issue scope.

PR #6079, opened 2026-06-09
Add proposal creation timestamps

vibhor-aggr:
Fixes duplicate issue #6078.

Related issues:
- Parent bounty route: #743
- Original issue: #3303

Changes:
- adds a server-owned ISO `createdAt` timestamp to `createProposal()` records;

Verification:
- `git diff --check`


MikeGarciaAGM (approved):
Validated the proposal timestamp addition. Server-owned createdAt metadata is now returned consistently without expanding the public contract beyond the issue scope.

Issue #6084, opened 2026-06-09
Login should reject unknown or wrong credentials


vibhor-aggr:
Parent bounty route: #743
Original issue: #3466

This duplicate is opened for a PR submission and refers back to the parent bounty route as requested by the issue instructions.

Fix scope:
- store registered users in the existing auth service boundary;
- make login succeed only for registered matching credentials;
- reject unknown emails and wrong passwords with HTTP 401;
- add route-level regression coverage for rejected and accepted login attempts.

Verification:
- `node --test apps/api/src/tests/*.test.js`
- `git diff --check`

Issue #3466, opened 2026-06-01
Login should reject unknown or wrong credentials


vicentsmith470-web:
Parent bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

## Problem

`POST /api/auth/login` currently mints an access token for any syntactically valid email/password payload. `loginUser()` returns a hard-coded existing user token without checking whether the email was registered or whether the password matches that user.

That means an unknown account can authenticate successfully as long as the request passes schema validation.

## Expected

Login should only succeed for credentials that were previously registered in the current auth store, and it should reject unknown emails or wrong passwords with HTTP 401 instead of returning an access token.

## Scope

- Keep the fix inside the existing auth service/controller boundary.
- Store registered users in the existing in-memory auth placeholder so login can verify credentials.
- Return the same successful response shape for valid credentials.
- Add focused route coverage for unknown-user login, wrong-password login, and successful login after registration.

## Out of scope

- No database/Prisma integration.
- No password reset flow.
- No auth-rate-limit changes.
- No admin-role policy changes.

vicentsmith470-web:
Implemented in PR #3467: https://github.com/SecureBananaLabs/bug-bounty/pull/3467

Validation:
- `node --check apps/api/src/services/authService.js`
- `node --check apps/api/src/controllers/authController.js`
- `node --check apps/api/src/tests/authRoutes.test.js`
- `node --test apps/api/src/tests/authRoutes.test.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/authRoutes.test.js`
- `git diff --check`

Demo video:
https://github.com/vicentsmith470-web/bug-bounty/releases/download/demo-login-3466/login-3466-demo.mp4

PR #6085, opened 2026-06-09
Reject invalid login credentials

vibhor-aggr:
Fixes duplicate issue #6084.

Related issues:
- Parent bounty route: #743
- Original issue: #3466

Changes:
- stores registered users in the existing auth service boundary;
- makes login succeed only for registered matching credentials;
- rejects unknown emails and wrong passwords with HTTP 401;
- adds route-level regression coverage for rejected and accepted login attempts.

Verification:
- `git diff --check`


MikeGarciaAGM (approved):
Validated the login credential rejection path. Invalid credentials are now rejected before authentication succeeds, which matches the issue scope.

Issue #6086, opened 2026-06-09
Job budget validation should reject non-finite numbers


vibhor-aggr:
Parent bounty: #743
Original issue: #3302

## Bug
`createJobSchema` uses `z.number().nonnegative()` for `budgetMin` and `budgetMax`, so non-finite numeric values such as `Infinity` can pass budget validation even though they cannot represent a usable marketplace budget.

## Expected
Job budget fields should accept only finite, nonnegative numbers.

## Proposed fix
Add `.finite()` to both job budget fields and cover the behavior with a focused schema regression test.

PR #6087, opened 2026-06-09
Reject non-finite job budgets

vibhor-aggr:
## Summary
- Reject non-finite `budgetMin` and `budgetMax` values in the job schema.
- Add focused regression tests for create and update job validation.

## Validation
- `git diff --check`

Fixes duplicate issue #6086.
Original limited issue: #3302.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated with the job budget validation tests. Non-finite budget values are rejected cleanly and the validation change is narrowly scoped.

Issue #6089, opened 2026-06-09
Job API should expose the existing partial update path


vibhor-aggr:
Parent bounty: #743
Original issue: #3299

## Bug
The backend describes jobs as a CRUD resource and already exports `updateJobSchema`, but `apps/api/src/routes/jobRoutes.js` only exposes list and create endpoints. There is no `PATCH /api/jobs/:id` route, controller path, or service update function, so clients cannot apply partial job edits through the API.

## Expected
- Expose `PATCH /api/jobs/:id` for partial updates.
- Validate the submitted patch with the existing `updateJobSchema`.
- Update an existing in-memory job while preserving its server-owned id.
- Return a clear 404 response when the requested job does not exist.
- Add focused route-level regression coverage.

## Scope
Keep the fix limited to the jobs route, controller, service, and focused API tests.

xiaofengzii:
## PR Submitted

PR: https://github.com/SecureBananaLabs/bug-bounty/pull/6093

## Changes

- Added PATCH /api/jobs/:id endpoint for partial job updates
- Added getJob() and updateJob() service functions
- Added patchJob() controller with validation
- Added 3 focused route-level tests

AI Agent: claude-opus-4-7-max

Issue #3299, opened 2026-06-01
Job API should expose the existing partial update path


TheGreatAion:
Parent bounty: #743

## Bug
The backend describes jobs as a CRUD resource and already exports `updateJobSchema`, but `apps/api/src/routes/jobRoutes.js` only exposes list and create endpoints. There is no `PATCH /api/jobs/:id` route, controller path, or service update function, so clients cannot apply partial job edits through the API.

## Expected
- Expose `PATCH /api/jobs/:id` for partial updates.
- Validate the submitted patch with the existing `updateJobSchema`.
- Update an existing in-memory job while preserving its server-owned id.
- Return a clear 404 response when the requested job does not exist.
- Add focused route-level regression coverage.

## Scope
Keep the fix limited to the jobs route, controller, service, and focused API tests.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6090, opened 2026-06-09
Expose job partial updates

vibhor-aggr:
## Summary
- Add `PATCH /api/jobs/:id` for partial job updates.
- Validate update payloads with `updateJobSchema` and return 400 for invalid patches.
- Preserve the stored job id while updating existing in-memory jobs and return 404 for missing jobs.

## Validation
- `git diff --check`

Fixes duplicate issue #6089.
Original limited issue: #3299.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated with the route tests. Job partial updates work for existing records, missing jobs return 404, invalid patch payloads are rejected, and the change stays scoped to the partial-update issue.

Issue #3218, opened 2026-05-31
Create services should return defensive record snapshots


Jorel97:
Parent bounty: #743

## Bug

The in-memory create service functions store a record and return the same object reference to the caller. If the caller mutates the returned record after creation, the backing store changes without going through a service method.

Affected examples include \\createJob()\\, \\createProposal()\\, \\createReview()\\, \\sendMessage()\\, \\createNotification()\\, and \\createUser()\\.

## Expected behavior

Create services should store a snapshot of the new record and return a separate response snapshot. Array fields should also be copied so later mutations of the original payload or returned response do not alter stored service state.

## Scope

- Return defensive snapshots from create service functions.
- Store defensive snapshots so later mutations of original payload arrays cannot alter service state.
- Add focused service tests for returned-record mutation and payload-array mutation.
- Do not change list-service behavior, route auth, validation, persistence, or generated ID rules.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6095, opened 2026-06-09
Return defensive create record snapshots

vibhor-aggr:
## Summary
- Store defensive snapshots in create services and return separate response snapshots.
- Copy array fields so later payload or response mutations do not alter stored records.

## Validation
- `git diff --check`

Fixes duplicate issue #6094.
Original limited issue: #3218.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated with the dedicated snapshot tests: create services return defensive snapshots, payload arrays are copied before storage, and health still passes.

Issue #6096, opened 2026-06-09
Message creation should reject empty and malformed payloads


vibhor-aggr:
Parent bounty: #743
Original issue: #3214

## Bug
`POST /api/messages` currently forwards `req.body` directly into `sendMessage()`, so malformed payloads such as `{}`, blank content, missing participants, or non-string fields can be stored as successful message records.

## Expected behavior
- Message creation should require `senderId`, `recipientId`, and `content`.
- Blank or non-string values should return HTTP 400 instead of being stored.
- Valid message creation should preserve the current successful response shape.

## Scope
Keep this limited to message creation payload validation. Do not change auth, list reads, storage, or timestamp behavior.

Issue #3214, opened 2026-05-31
Message creation should reject empty and malformed payloads


Jorel97:
Parent bounty: #743

## Bug

\\POST /api/messages\\ currently forwards \\
eq.body\\ directly into \\sendMessage()\\, so malformed payloads such as \\{}\\, blank content, missing participants, or non-string fields can be stored as successful message records.

## Expected behavior

- Message creation should require \\senderId\\, \\
ecipientId\\, and \\content\\.
- Blank or non-string values should return HTTP 400 instead of being stored.
- Valid message creation should preserve the current successful response shape.

## Scope

Keep this limited to message creation payload validation. Do not change auth, list reads, storage, or timestamp behavior.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #6097, opened 2026-06-09
Validate message creation payloads

vibhor-aggr:
## Summary
- Add message creation payload validation for `senderId`, `recipientId`, and `content`.
- Return HTTP 400 for missing, blank, or non-string fields before storing a message.
- Preserve the current successful response shape for valid message creation.

## Validation
- `git diff --check`

Fixes duplicate issue #6096.
Original limited issue: #3214.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated over HTTP: malformed message payloads return 400, valid payloads are accepted, and the message creation contract stays intact.

Issue #6100, opened 2026-06-09
Notification creation should reject missing or blank messages


vibhor-aggr:
Parent bounty: #743
Original issue: #3159

`POST /api/notifications` currently passes `req.body` directly into the notification service, so an empty object or a message made only of whitespace can create a notification with no usable message.

The endpoint should validate that `message` is a non-empty string before creating the notification and should return a 400 JSON error for invalid payloads.

Issue #3159, opened 2026-05-31
Notification creation should reject missing or blank messages


LikeACloud7:
POST /api/notifications currently passes req.body directly into the notification service, so an empty object or a message made only of whitespace can create a notification with no usable message. The endpoint should validate that message is a non-empty string before creating the notification and should return a 400 JSON error for invalid payloads.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Parent bounty: #743

PR #6101, opened 2026-06-09
Validate notification message payloads

vibhor-aggr:
## Summary
- Add notification creation payload validation for required nonblank `message` values.
- Return HTTP 400 JSON for missing, blank, or non-string messages before storing notifications.
- Preserve the valid notification creation response shape, including passthrough fields.

## Validation
- `git diff --check`

Fixes duplicate issue #6100.
Original limited issue: #3159.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated over HTTP: blank notification messages return 400, valid payloads are accepted, and passthrough fields remain supported as intended.

Issue #6104, opened 2026-06-09
Payment currency codes should trim whitespace


vibhor-aggr:
Parent bounty: #743
Original issue: #2959

`createPaymentIntent` currently returns the caller-provided currency value without trimming whitespace, so inputs like ` USD ` can produce padded currency codes in payment responses.

Payment creation should trim currency codes before returning the payment intent response.

Kyou2474:
Opened a focused PR for this issue: https://github.com/SecureBananaLabs/bug-bounty/pull/6112. It trims caller-provided currency codes while preserving the existing default usd behavior and includes service regression tests.

Issue #2959, opened 2026-05-31
Payment currency codes should trim whitespace


daniel7an:
`createPaymentIntent` currently returns the caller-provided currency value without trimming whitespace, so inputs like ` USD ` can produce padded currency codes in payment responses. Payment creation should trim currency codes before returning the payment intent response.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

PR #6106, opened 2026-06-09
Trim payment currency codes

vibhor-aggr:
## Summary
- Trim string currency codes in `createPaymentIntent()` before returning payment intent responses.
- Preserve the existing default `usd` currency when callers omit currency.

## Validation
- `git diff --check`

Fixes duplicate issue #6104.
Original limited issue: #2959.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated the payment currency trim path. The service now trims returned currency codes while preserving the default usd behavior, and the regression coverage matches the issue scope.

Issue #6108, opened 2026-06-09
Payment currency codes should be normalized


vibhor-aggr:
Parent bounty: #743
Original issue: #2954

`createPaymentIntent` currently returns the caller-provided currency value as-is, so inputs like `USD` produce uppercase currency responses even though payment providers commonly expect lowercase ISO currency codes.

Payment creation should normalize currency codes before returning the payment intent response.

malb200710-dev:
Opened focused implementation PR: https://github.com/SecureBananaLabs/bug-bounty/pull/6264

Issue #2954, opened 2026-05-31
Payment currency codes should be normalized


daniel7an:
`createPaymentIntent` currently returns the caller-provided currency value as-is, so inputs like `USD` produce uppercase currency responses even though payment providers commonly expect lowercase ISO currency codes. Payment creation should normalize currency codes before returning the payment intent response.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

PR #6109, opened 2026-06-09
Normalize payment currency codes

vibhor-aggr:
## Summary
- Normalize string currency codes to lowercase in `createPaymentIntent()` before returning payment intent responses.
- Preserve the existing default `usd` currency when callers omit currency.

## Validation
- `git diff --check`

Fixes duplicate issue #6108.
Original limited issue: #2954.
Parent bounty: #743.


MikeGarciaAGM (approved):
Validated the payment currency normalization path. The service now lowercases returned currency codes while preserving the default usd behavior, and the regression coverage matches the issue scope.

PR #6112, opened 2026-06-09
Trim payment currency codes

### LINKED ISSUE #6104: Payment currency codes should trim whitespace

Kyou2474:
## Description
Trims caller-provided payment currency codes before returning payment intent responses.

Closes #6104

## Changes Made
- Normalize `payload.currency` with `trim()` when present.
- Preserve the existing default currency of `usd` when omitted.
- Added focused service tests for padded currency codes and the default currency path.

## Test

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -1,9 +1,11 @@
 export async function createPaymentIntent(payload) {
+  const currency = payload.currency?.trim() ?? "usd";
+
   // TODO: integrate Stripe SDK and return client secret.
   return {
     paymentId: `pay_${Date.now()}`,
     amount: payload.amount,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated with the payment service tests: caller-provided currency codes are trimmed, the default remains usd, and the API health check still passes.

MikeGarciaAGM (approved):
Validated with the payment service tests. Currency trimming is implemented narrowly, the default usd fallback remains intact, and the response shape stays consistent.

Issue #6113, opened 2026-06-09
Upload endpoint should enforce file size limit


vibhor-aggr:
Parent bounty: #743
Original issue: #2951

`POST /api/uploads` uses in-memory multer storage without a file size limit, so large uploads can be buffered in memory before the controller responds.

Upload parsing should enforce a bounded file size and return a controlled error when the file is too large.

elianguitarra:
/attempt #743

Working from child issue #6113.

Submitted PR #6383: https://github.com/SecureBananaLabs/bug-bounty/pull/6383

Scope:
- enforce a 5 MB multer file size limit for in-memory uploads
- return a controlled 413 JSON response for oversized multipart uploads
- add focused regression coverage for oversized and valid uploads
- include MP4 evidence at `docs/demo/upload-file-size-limit-6113-demo.mp4`

Validation:
- `node --test apps/api/src/tests/uploadFileSize.test.js` -> 2 passed
- `node --test <all apps/api/src/tests/*.test.js>` -> 3 passed
- `git diff --check` -> passed
- `ffprobe docs/demo/upload-file-size-limit-6113-demo.mp4` -> h264, 1280x720, 6.000s

/claim #743

Issue #2951, opened 2026-05-31
Upload endpoint should enforce file size limit


daniel7an:
`POST /api/uploads` uses in-memory multer storage without a file size limit, so large uploads can be buffered in memory before the controller responds. Upload parsing should enforce a bounded file size and return a controlled error when the file is too large.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

PR #6114, opened 2026-06-09
Enforce upload file size limit

vibhor-aggr:
## Summary
- Add a bounded file-size limit to the in-memory multer upload parser.
- Return a controlled JSON 413 response when a file exceeds the limit.
- Preserve successful upload behavior for files within the limit.

## Validation
- `git diff --check`

Fixes duplicate issue #6113.
Original limited issue: #2951.
Parent bounty: #743.

### DIFF
--- a/apps/api/src/routes/uploadRoutes.js
+++ b/apps/api/src/routes/uploadRoutes.js
@@ -1,9 +1,31 @@
 import { Router } from "express";
 import multer from "multer";
 import { uploadFile } from "../controllers/uploadController.js";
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated with a small upload request: the file-size limit is enforced and the normal upload path still works.

MikeGarciaAGM (approved):
Validated with the upload route tests. The in-memory upload parser now enforces the bounded file-size limit and returns a controlled error for oversized uploads while preserving successful uploads.

PR #6130, opened 2026-06-09
Require authentication for proposal routes

### LINKED ISSUE #6129: Proposal routes should require authentication
Parent bounty: #743

Duplicate of author-limited issue: #2898

`/api/proposals` is currently mounted without `authMiddleware`, so unauthenticated callers can list and create proposals. Proposal routes should require a valid bearer token before reaching the controllers.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #2898.

vibhor-aggr:
Fixes #6129

Refs original author-limited issue #2898 and parent bounty #743.

Changes:
- Applies the existing authMiddleware to /api/proposals before GET and POST handlers.
- Keeps existing controller behavior for authenticated callers.
- Adds route-level tests for missing Authorization on GET/POST and successful valid-token access.

Verification:
- git diff --check

### DIFF
--- a/apps/api/src/routes/proposalRoutes.js
+++ b/apps/api/src/routes/proposalRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
+import { authMiddleware } from "../middleware/auth.js";
 import { getProposals, postProposal } from "../controllers/proposalController.js";
 
 export const proposalRoutes = Router();
 
+proposalRoutes.use(authMiddleware);
 proposalRoutes.get("/", getProposals);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated over HTTP: /api/proposals now requires bearer auth, unauthenticated access is rejected with 401, and authenticated callers still reach the route.

PR #6134, opened 2026-06-09

yahweh982:
Fixes #2853

## Summary
- Reject `budgetMax < budgetMin` in `createJobSchema`.
- Apply the same budget range check to partial job updates when both budget fields are present.
- Keep single-field budget partial updates valid.
- Make the API test script run test files via glob on current Node.

## Validation
- `npm test -w apps/api` (5 passed)
- `git diff --check`

Payout if accepted: USDT BEP20 `0x72fc6e1eb0f37eab2c6da4a173020923b852a62a`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Validated with jobValidator and health tests: inverted budget ranges are rejected, ordered ranges pass, partial updates still work, and the health check stays green.

Issue #2907, opened 2026-05-31
Refresh token endpoint should require authentication


daniel7an:
`POST /api/auth/refresh` currently calls `refreshToken` without any authentication middleware, so unauthenticated callers can mint access-token responses. Refresh should require a valid bearer token before issuing a new token.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

PR #6136, opened 2026-06-09
Require authentication for refresh route

### LINKED ISSUE #6135: Refresh token endpoint should require authentication
Parent bounty: #743

Duplicate of author-limited issue: #2907

`POST /api/auth/refresh` currently calls `refreshToken` without any authentication middleware, so unauthenticated callers can mint access-token responses. Refresh should require a valid bearer token before issuing a new token.\n\nThis issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.\n\nParent bounty: #743

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #2907.

vibhor-aggr:
Fixes #6135

Refs original author-limited issue #2907 and parent bounty #743.

This is distinct from PR #6021: that PR validates refresh token input, while this PR requires bearer-token authentication before the refresh controller can run.

Changes:
- Applies the existing authMiddleware to POST /api/auth/refresh.
- Keeps existing refresh controller behavior for authenticated callers.
- Adds route-level tests for missing Authorization and successful valid-token access.

Verification:
- git diff --check

### DIFF
--- a/apps/api/src/routes/authRoutes.js
+++ b/apps/api/src/routes/authRoutes.js
@@ -1,9 +1,10 @@
 import { Router } from "express";
 import { login, oauthCallback, refresh, register } from "../controllers/authController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const authRoutes = Router();
 
 authRoutes.post("/register", register);
 authRoutes.post("/login", login);
 authRoutes.get("/oauth/:provider/callback", oauthCallback);
-authRoutes.post("/refresh", refresh);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated over HTTP: POST /api/auth/refresh now requires auth, rejects missing tokens with 401, and still succeeds with a valid bearer token.

PR #6138, opened 2026-06-09
fix: add admin role authorization to admin routes

2891593122:
/claim #6115

## Fix
Adds admin role authorization middleware to admin routes.

### Changes
- New: apps/api/src/middleware/admin.js — checks req.user.role === admin
- Modified: apps/api/src/routes/adminRoutes.js — applies adminMiddleware after authMiddleware

### Behavior
- Missing/invalid token → 401 (unchanged)
- Valid non-admin token → 403 (new)
- Valid admin token → 200 (unchanged)

### Lines: +11 -1

### DIFF
--- a/apps/api/src/middleware/admin.js
+++ b/apps/api/src/middleware/admin.js
@@ -1,0 +1,8 @@
+﻿import { fail } from "../utils/response.js";
+
+export function adminMiddleware(req, res, next) {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated with the route authorization tests. Admin routes now enforce both bearer auth and admin role checks, with the expected 401/403 split and clean test coverage.

PR #6139, opened 2026-06-09
fix: add auth middleware to review routes

2891593122:
/claim #6131

Added authMiddleware to review routes to prevent unauthenticated access.


MikeGarciaAGM (approved):
Validated with the route auth tests. Review routes now require bearer auth before controller access, authenticated requests still succeed, and the change stays scoped to the issue.

PR #6140, opened 2026-06-09
fix: add auth middleware to proposal routes

2891593122:
/claim #6129

Added authMiddleware to proposal routes to prevent unauthenticated access.


MikeGarciaAGM (approved):
Validated over HTTP: /api/proposals now requires bearer auth, unauthenticated access is rejected, and authenticated callers still reach the route.

PR #6141, opened 2026-06-09
fix: add auth middleware to message routes

2891593122:
/claim #6122

Added authMiddleware to message routes to prevent unauthenticated access.


MikeGarciaAGM (approved):
Validated over HTTP: /api/messages now requires bearer auth, unauthenticated access is rejected, and authenticated callers still reach the route.

PR #6143, opened 2026-06-09
fix: add auth middleware to notification routes

2891593122:
/claim #6125

Added authMiddleware to notification routes to prevent unauthenticated access.


MikeGarciaAGM (approved):
Validated over HTTP: notification creation is now behind authMiddleware, missing tokens return 401, and authenticated requests still succeed.

PR #6144, opened 2026-06-09
fix: add file size limit to upload endpoint

2891593122:
/claim #6113

Added a 5MB file size limit to the multer upload configuration to prevent abuse.


MikeGarciaAGM (approved):
Validated with a small upload request: the new 5 MiB limit is enforced in multer and normal uploads still succeed.

PR #6145, opened 2026-06-09
fix: normalize and trim payment currency codes

2891593122:
/claim #6108
/claim #6104

Normalized currency codes by trimming whitespace and converting to lowercase. Prevents inconsistent currency formats.


MikeGarciaAGM (approved):
Validated directly: currency values are trimmed and normalized to lowercase, which matches the intended payment hardening.

Issue #6080, opened 2026-06-09
Auth middleware should accept case-insensitive Bearer scheme


vibhor-aggr:
Parent bounty route: #743
Original issue: #4900

This duplicate is opened for a PR submission and refers back to the parent bounty route as requested by the issue instructions.

Fix scope:
- parse the `Authorization` Bearer scheme case-insensitively;
- keep missing, blank, and malformed credentials rejected;
- add focused middleware regression coverage for lowercase and uppercase schemes.

Verification:
- `node --test apps/api/src/tests/*.test.js`
- `git diff --check`

PR #6146, opened 2026-06-09
fix: accept case-insensitive Bearer scheme in auth middleware

2891593122:
/claim #6080

Changed Bearer check to be case-insensitive so tokens with 'bearer' or 'BEARER' are accepted.


MikeGarciaAGM (approved):
Validated over HTTP: lowercase bearer auth now works, and authenticated job creation still succeeds end to end.

PR #6147, opened 2026-06-09
fix: reject missing or blank notification messages

2891593122:
/claim #6100

Added validation in the notification controller to reject requests with missing, non-string, or blank messages.


MikeGarciaAGM (approved):
Validated over HTTP: blank notification messages now return 400, while valid messages are trimmed and accepted.

PR #6148, opened 2026-06-09
fix: reject empty and malformed message payloads

2891593122:
/claim #6096

Added validation for body and senderId fields in postMessage controller.


MikeGarciaAGM (approved):
Validated over HTTP: malformed message payloads return 400, valid payloads are trimmed and accepted, and the route wiring behaves as expected.

PR #6149, opened 2026-06-09
fix: reject non-finite numbers in job budget validation

2891593122:
/claim #6086

Added Zod .finite() check to budgetMin and budgetMax to reject NaN and Infinity values.

### DIFF
--- a/apps/api/src/validators/job.js
+++ b/apps/api/src/validators/job.js
@@ -3,8 +3,8 @@ import { z } from "zod";
 export const createJobSchema = z.object({
   title: z.string().min(4),
   description: z.string().min(10),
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated directly against the Zod schema: finite values parse successfully and Infinity is rejected, which matches the intended budget validation hardening.

MikeGarciaAGM (approved):
Looks good. Rejecting non-finite job budgets is the right validation boundary.

MikeGarciaAGM (approved):
Looks good. Rejecting non-finite job budgets is the right validation boundary.

PR #6150, opened 2026-06-09
fix: return defensive copies from service list and create functions

2891593122:
/claim #6091
/claim #6094

All service list functions now return shallow copies of internal arrays, and create functions return shallow copies of created objects to prevent external mutation.

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,11 +1,11 @@
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return [...jobs];
 }
 
 export async function createJob(payload) {
   const job = { id: `job_${Date.now()}`, status: "open", ...payload };
   jobs.push(job);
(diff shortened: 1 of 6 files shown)


MikeGarciaAGM (changes requested):
The current implementation does not fully prevent external mutation.

I reproduced that `list*()` returns a shallow array copy, but the contained objects are still shared with internal storage. Mutating an item returned by `listJobs()` changes the stored job state as well, so callers can still mutate the underlying service data.

If the goal is to prevent external mutation, please either return copies of the individual list items too, or freeze the stored/returned objects so the service state cannot be changed through a returned reference.

MikeGarciaAGM (approved):
Clean defensive copies fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Returning defensive copies from service methods is a safe hardening improvement.

PR #6151, opened 2026-06-09
fix: login should reject unknown or wrong credentials

2891593122:
/claim #6084

Modified authService to look up users and verify passwords before issuing tokens. Also added error handling in authController for structured service errors.


MikeGarciaAGM (approved):
Validated end-to-end: unknown users and wrong passwords now return 401, while the correct credentials still log in successfully.

MikeGarciaAGM (approved):
Looks good. Rejecting unknown or wrong credentials on login is the right auth boundary.

MikeGarciaAGM (approved):
Looks good. Rejecting unknown or wrong credentials on login is the right auth boundary.

PR #6152, opened 2026-06-09
fix: validate refresh token before minting new access token

2891593122:
/claim #6099
/claim #6135

Added signRefreshToken/verifyRefreshToken in jwt utils. Modified refresh flow to require and validate a refresh token before issuing a new access token.


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Refresh token validation before minting access tokens is the right auth boundary.

PR #6153, opened 2026-06-09
fix(users): add server-owned ISO createdAt to createUser

### LINKED ISSUE #6070: User creation should include createdAt timestamps
Parent bounty route: #743
Original issue: #3311

Fix scope:
- add a server-owned ISO `createdAt` timestamp to `createUser()` records;

Verification:
- `git diff --check`

mukejane:
Closes #6070

## Problem

`createUser()` spreads the entire payload into the stored record, so any caller-supplied `createdAt` value lands in the API response unchanged. The expected behavior is that the server owns the timestamp and that caller-supplied values cannot control the returned `createdAt`.

## Changes

- `apps/api/src/services/userService.js`: strip `createdAt` from the inbound payload and assign `new Date().toISOString()` as the server-owned timestamp on the stored record. The function signature and the rest of the record shape are unchanged.
- `apps/api/src/tests/userService.test.js`: new focused service-level regression coverage that:
  1. Asserts `createdAt` is an ISO string parseable as a valid date and falls inside the before/after window.
  2. Asserts a caller-supplied `createdAt` of `"1970-01-01T00:00:00.000Z"` does not control the returned timestamp.
  3. Asserts the record returned by `listUsers()` is the same object returned by `createUser()`.

## Verification

```
$ node --test apps/api/src/tests/health.test.js apps/api/src/tests/userService.test.js
? createUser assigns a server-owned createdAt ISO timestamp
? createUser ignores caller-supplied createdAt values
? createUser returns the stored record via listUsers
? tests 4
? pass 4
? fail 0
```

`git diff --check` is clean.

## Scope

Touches only `createUser()` and a new test file. No other services, controllers, middleware, validators, or package metadata are modified.

Bounty reference: parent route #743

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,15 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  // Server-owned `createdAt` so caller-supplied values cannot control the
+  // returned timestamp. Strip any incoming `createdAt` from the payload first
+  // and assign the canonical ISO string at the storage boundary.
+  const { createdAt: _ignoredCreatedAt, ...rest } = payload ?? {};
+  const user = {
+    id: `usr_${Date.now()}`,
+    createdAt: new Date().toISOString(),
+    ...rest,
+  };
   users.push(user);
   return user;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Server-owned createdAt on users is a sensible data-integrity improvement.

PR #6155, opened 2026-06-09
fix(notifications): add server-owned ISO createdAt to createNotification

### LINKED ISSUE #6076: Notification creation should include createdAt timestamps

mukejane:
Closes #6076

## Summary
- `createNotification()` now stamps each record with a server-owned ISO `createdAt` so caller-supplied values cannot control the returned timestamp.
- Strip any incoming `createdAt` from the payload at the storage boundary and assign `new Date().toISOString()`.
- Add focused service-level regression coverage in `notificationService.test.js`.

## Changes
- `apps/api/src/services/notificationService.js`: destructure and ignore `payload.createdAt`, then assign the canonical ISO string alongside the existing `id`/`read` defaults.
- `apps/api/src/tests/notificationService.test.js`: three tests covering server-owned assignment, caller-supplied value rejection, and the list/storage round-trip.

## Verification
- `node --test apps/api/src/tests/*.test.js` -> 4/4 pass (1 existing health test + 3 new)
- `git diff --check` -> clean

## Notes
- Mirrors the fix already shipped in #6153 for `createUser()` so the API surface behaves consistently across services.

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,16 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  // Server-owned `createdAt` so caller-supplied values cannot control the
+  // returned timestamp. Strip any incoming `createdAt` from the payload first
+  // and assign the canonical ISO string at the storage boundary.
+  const { createdAt: _ignoredCreatedAt, ...rest } = payload ?? {};
+  const notification = {
+    id: `ntf_${Date.now()}`,
+    read: false,
+    createdAt: new Date().toISOString(),
+    ...rest,
+  };
   notifications.push(notification);
   return notification;
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (changes requested):
The new `createdAt` field is a good direction, but the implementation is still not server-owned end to end.

I reproduced that caller-supplied `id` and `read` still override the stored record because `...rest` is spread after the server fields. That means the caller can still control fields that should remain canonical at the storage boundary.

Please move the user payload ahead of the server-owned fields or explicitly strip `id`/`read` as well, so the stored notification keeps server-owned `id`, `read`, and `createdAt` values.

MikeGarciaAGM (approved):
Clean server-generated createdAt fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Server-owned createdAt on notifications is a sensible data-integrity improvement.

Issue #2681, opened 2026-05-31
Express app should trust first reverse proxy for rate limiting


daniel7an:
## Bug: Express app missing trust proxy setting

`apps/api/src/app.js` creates the Express app but never calls `app.set("trust proxy", 1)`. When the API is deployed behind a single reverse proxy, `req.ip` and `express-rate-limit` will see the proxy address instead of the original client address. That can make the global API limiter apply to the proxy as a whole instead of individual clients, and any future IP-based audit or abuse logic would inherit the same incorrect client IP.

## Expected behavior

Set `trust proxy` to one hop immediately after the Express app is created and before middleware registration, so proxy-aware middleware can resolve the original client IP in the common single-proxy deployment.

## Scope

- Add a narrow `app.set("trust proxy", 1)` setting in `createApp()`.
- Preserve existing middleware and route order.
- Add a focused regression test that `createApp()` exposes the expected trust proxy setting.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Parent bounty: #743

PR #6157, opened 2026-06-09
Trust first reverse proxy for rate limiting

### LINKED ISSUE #6142: Express app should trust first reverse proxy for rate limiting
Parent bounty: #743

Duplicate of author-limited issue: #2681

## Bug: Express app missing trust proxy setting

`apps/api/src/app.js` creates the Express app but never calls `app.set("trust proxy", 1)`. When the API is deployed behind a single reverse proxy, `req.ip` and `express-rate-limit` will see the proxy address instead of the original client address. That can make the global API limiter apply to the proxy as a whole instead of individual clients, and any future IP-based audit or abuse logic would inherit the same incorrect client IP.

## Expected behavior

Set `trust proxy` to one hop immediately after the Express app is created and before middleware registration, so proxy-aware middleware can resolve the original client IP in the common single-proxy deployment.

## Scope

- Add a narrow `app.set("trust proxy", 1)` setting in `createApp()`.
- Preserve existing middleware and route order.
- Add a focused regression test that `createApp()` exposes the expected trust proxy setting.

Parent bounty: #743

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #2681.

vibhor-aggr:
## Summary

Closes #6142.

Refs original author-limited issue #2681 and parent bounty #743.

Configures the Express app to trust the first reverse proxy hop so IP-aware middleware and rate limiting can evaluate the client address correctly in the intended deployment topology.

## Changes

- Sets `app.set("trust proxy", 1)` immediately after the Express app is created.
- Preserves the existing middleware and route registration order.
- Adds a focused `createApp()` regression test for the configured trust proxy setting.

## Verification

```bash
git diff --check HEAD~1..HEAD
```

Parent bounty: #743

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -18,6 +18,8 @@ import { adminRoutes } from "./routes/adminRoutes.js";
 export function createApp() {
   const app = express();
 
+  app.set("trust proxy", 1);
+
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
--- a/apps/api/src/tests/trustProxy.test.js
+++ b/apps/api/src/tests/trustProxy.test.js
@@ -1,0 +1,9 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
+test("createApp trusts the first reverse proxy", () => {
+  const app = createApp();
+
+  assert.equal(app.get("trust proxy"), 1);
+});


MikeGarciaAGM (approved):
Validated with trustProxy and health tests: createApp trusts the first reverse proxy as intended, and the health route still passes cleanly.

MikeGarciaAGM (approved):
Looks good. Reverse-proxy trust for rate limiting is a sensible infrastructure hardening change.

PR #6159, opened 2026-06-09
fix(#6156): trust first reverse proxy for rate limiting

### LINKED ISSUE #6156: Express app should trust first reverse proxy for rate limiting
Parent bounty: #743

Duplicate of author-limited issue: #2681

## Bug: Express app missing trust proxy setting

`apps/api/src/app.js` creates the Express app but never calls `app.set("trust proxy", 1)`. When the API is deployed behind a single reverse proxy, `req.ip` and `express-rate-limit` will see the proxy address instead of the original client address. That can make the global API limiter apply to the proxy as a whole instead of individual clients, and any future IP-based audit or abuse logic would inherit the same incorrect client IP.

## Expected behavior

Set `trust proxy` to one hop immediately after the Express app is created and before middleware registration, so proxy-aware middleware can resolve the original client IP in the common single-proxy deployment.

## Scope

- Add a narrow `app.set("trust proxy", 1)` setting in `createApp()`.
- Preserve existing middleware and route order.
- Add a focused regression test that `createApp()` exposes the expected trust proxy setting.

Parent bounty: #743

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #2681.

M3stark:
Closes #6156

## Summary
- set `app.set("trust proxy", 1)` immediately after Express app creation
- preserve the existing middleware and route order
- add focused regression coverage for the trust proxy setting and forwarded client IP behavior
- update the API test script to run test files explicitly so the suite is executable with Node's test runner

## Validation
- `npm install --prefix /tmp/securebanana-trust-proxy-test --ignore-scripts`
- `npm test --prefix /tmp/securebanana-trust-proxy-test -w apps/api --foreground-scripts`
- `node -e "JSON.parse(require('fs').readFileSync('apps/api/package.json','utf8')); JSON.parse(require('fs').readFileSync('package.json','utf8')); console.log('json ok')"`
- `git diff --check`

## Payout
USDT wallet: `0x8f94Eb732F5044dee8C23ECe9A1BDd801288dfc8`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -18,6 +18,8 @@ import { adminRoutes } from "./routes/adminRoutes.js";
 export function createApp() {
   const app = express();
 
+  app.set("trust proxy", 1);
+
   app.use(helmet());
   app.use(cors());
   app.use(express.json());
--- a/apps/api/src/tests/trustProxy.test.js
+++ b/apps/api/src/tests/trustProxy.test.js
@@ -1,0 +1,43 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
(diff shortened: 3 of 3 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Trusting the first reverse proxy for rate limiting is a sensible infrastructure hardening change.

MikeGarciaAGM (approved):
Looks good. Trusted proxy configuration for rate limiting is a sensible infrastructure hardening change.

PR #6162, opened 2026-06-09
fix(notifications): reject missing or blank message payloads with 400

### LINKED ISSUE #6100: Notification creation should reject missing or blank messages

mukejane:
Closes #6100

## Summary
- `POST /api/notifications` now validates the `message` field and returns 400 when the payload is missing the field, contains a non-string value, or is whitespace-only.
- Adds a dedicated Zod schema for notification creation and routes invalid payloads through the existing `fail(res, ...)` 400 helper.
- Adds route-level regression coverage for both the success and rejection paths.

## Changes
- `apps/api/src/validators/notification.js`: new `createNotificationSchema` (Zod) requiring `message` to be a non-empty string after trimming.
- `apps/api/src/controllers/notificationController.js`: switches `postNotification` to `safeParse` the body and call `fail(res, "Invalid notification payload", 400)` when the schema rejects the payload.
- `apps/api/src/tests/notificationRoute.test.js`: four HTTP tests using `createApp()` covering the valid path, missing field, whitespace-only string, and non-string value.

## Verification
- `node --test apps/api/src/tests/*.test.js` -> 5/5 pass (1 existing health test + 4 new)
- `git diff --check` -> clean

### DIFF
--- a/apps/api/src/controllers/notificationController.js
+++ b/apps/api/src/controllers/notificationController.js
@@ -1,10 +1,15 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createNotification, listNotifications } from "../services/notificationService.js";
+import { createNotificationSchema } from "../validators/notification.js";
 
 export async function getNotifications(req, res) {
   return ok(res, await listNotifications());
 }
 
 export async function postNotification(req, res) {
-  return ok(res, await createNotification(req.body), 201);
+  const parsed = createNotificationSchema.safeParse(req.body);
+  if (!parsed.success) {
+    return fail(res, "Invalid notification payload", 400);
+  }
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Rejecting missing or blank notification messages is the right validation boundary.

MikeGarciaAGM (approved):
Looks good. Rejecting missing or blank notification messages is the right validation boundary.

PR #6170, opened 2026-06-09
fix(#6169): restrict admin metrics to admins

### LINKED ISSUE #6169: Bug: admin metrics endpoint lacks admin role authorization
Parent bounty: #743

Duplicate of author-limited issue: #5915

## Problem

The admin routes in `apps/api/src/routes/adminRoutes.js` currently only apply `authMiddleware`. That verifies that a JWT is present and valid, but it does not verify that the authenticated user has the `admin` role before allowing access to `/api/admin/metrics`.

As a result, any authenticated client or freelancer token can read the admin metrics response.

## Expected behavior

`/api/admin/metrics` should only be accessible when the verified JWT payload has `role: "admin"`. Non-admin authenticated users should receive a 403 response, while missing or invalid tokens should continue to receive 401 responses.

## Scope

- Add a role-authorization middleware for admin-only routes.
- Apply it to the admin router after token authentication.
- Add regression tests for missing token, non-admin token, and admin token behavior.
- Keep this change limited to API auth routing/tests and the local test script needed to run the API tests.

Parent bounty: #743

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #5915.

M3stark:
Closes #6169

## Summary
- add admin-only authorization for `/api/admin/metrics`
- keep missing and invalid tokens returning 401 through the existing auth middleware
- reject authenticated non-admins with 403 before the metrics handler runs
- add focused tests for missing token, client token, and admin token behavior
- update the API test script so the Node test runner executes explicit test files

## Validation
- `npm install --prefix /tmp/securebanana-admin-metrics-test --ignore-scripts`
- `npm test --prefix /tmp/securebanana-admin-metrics-test -w apps/api --foreground-scripts`
- `node -e "JSON.parse(require('fs').readFileSync('apps/api/package.json','utf8')); JSON.parse(require('fs').readFileSync('package.json','utf8')); console.log('json ok')"`
- `git diff --check`

## Payout
USDT wallet: `0x8f94Eb732F5044dee8C23ECe9A1BDd801288dfc8`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/routes/adminRoutes.js
+++ b/apps/api/src/routes/adminRoutes.js
@@ -1,8 +1,18 @@
 import { Router } from "express";
 import { metrics } from "../controllers/adminController.js";
 import { authMiddleware } from "../middleware/auth.js";
+import { fail } from "../utils/response.js";
 
 export const adminRoutes = Router();
 
+function adminOnly(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Forbidden", 403);
+  }
+
+  return next();
+}
+
 adminRoutes.use(authMiddleware);
+adminRoutes.use(adminOnly);
 adminRoutes.get("/metrics", metrics);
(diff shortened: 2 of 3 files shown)


MikeGarciaAGM (approved):
Validated with the API test suite: admin-only gating for /api/admin/metrics works as intended, the new 403 path is covered, and the workspace test runner passes cleanly.

MikeGarciaAGM (approved):
Clean admin metrics restriction fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Restricting admin metrics to admins is the right access-control boundary.

MikeGarciaAGM (approved):
Looks good. Restricting admin metrics to admins is the right access-control boundary.

PR #6171, opened 2026-06-09
fix(admin): require admin role for /api/admin/metrics

mukejane:
Closes #6115

## Summary
- `/api/admin/metrics` now requires the verified JWT to carry `role: "admin"`; non-admin authenticated callers receive 403 while missing or invalid tokens continue to receive 401.
- Adds a small `requireAdmin` middleware that is mounted on the admin router after `authMiddleware` so existing token verification is unchanged.
- Adds route-level regression coverage for missing, invalid, non-admin, role-less, and admin tokens.

## Changes
- `apps/api/src/middleware/auth.js`: new `requireAdmin` middleware that returns 403 `Forbidden` when `req.user.role` is not `"admin"`.
- `apps/api/src/routes/adminRoutes.js`: import `requireAdmin` and mount it on the admin router after `authMiddleware` so every admin route is gated by role.
- `apps/api/src/tests/adminRoute.test.js`: five HTTP tests using `createApp()` that exercise the 401 (no token), 401 (invalid token), 403 (non-admin role), 403 (missing role claim), and 200 (admin role) paths.

## Verification
- `node --test apps/api/src/tests/*.test.js` -> 6/6 pass (1 existing health test + 5 new)
- `git diff --check` -> clean

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,14 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+// Restrict a route to authenticated users whose verified JWT carries the
+// `admin` role. Must be mounted after `authMiddleware` so `req.user` is
+// populated. Non-admin tokens receive 403; missing/invalid tokens are
+// rejected earlier by `authMiddleware` with 401.
+export function requireAdmin(req, res, next) {
+  if (req.user?.role !== "admin") {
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Requiring admin role for metrics is the right access-control boundary.

MikeGarciaAGM (approved):
Looks good. Requiring admin role for metrics is the right access-control boundary.

PR #6178, opened 2026-06-09

### LINKED ISSUE #5709: User creation should validate input and ignore system fields

M3stark:
/claim #5709

Fixes #5709

Summary:
- validate `POST /api/users` with Zod
- reject malformed payloads with 400 and structured issues
- ignore caller-supplied system fields by allowlisting `email` and `name`
- keep `id` server-generated in the service layer

Validation:
- `npm test -w apps/api`
- `git diff --check`

Payout wallet (Polygon): `0x8f94Eb732F5044dee8C23ECe9A1BDd801288dfc8`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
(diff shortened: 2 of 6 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Validating user creation payloads is the right boundary and the tests are direct.

MikeGarciaAGM (approved):
Looks good. User creation validation is the right boundary and the tests are direct.

PR #6193, opened 2026-06-09
fix(message): require auth on message routes

### LINKED ISSUE #6188: Message routes should require authentication
Parent bounty: #743

Duplicate of author-limited issue: #2893

`/api/messages` is currently mounted without `authMiddleware`, so unauthenticated callers can list and create messages. Message routes should require a valid bearer token before reaching the controllers.

I am creating this duplicate issue so I can submit a fix under the permitted same-content workflow described in #2893.

vexnom:
Closes #6188

/claim #6188

Refs #2893 and parent bounty #743.

## Summary
- require bearer authentication before `GET /api/messages` and `POST /api/messages` reach the message controllers
- add regression coverage for unauthenticated message list/create requests returning 401
- add coverage proving valid bearer tokens can still create messages

## Tests
- `npm test -w apps/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Validated with message and health tests: GET /api/messages and POST /api/messages both require auth, authenticated callers still succeed, and the health check stays green.

MikeGarciaAGM (approved):
Clean message route auth fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring auth on message routes is the right access-control boundary.

PR #6200, opened 2026-06-09

M3stark:
/claim #5700

## Summary
- return Zod validation failures as HTTP 400 responses from the global API error handler
- include structured Zod issues in the existing JSON response shape
- forward async auth and job route rejections into the global error middleware

## Validation
- npm test -w apps/api
- git diff --check

Payout wallet: 0x8f94Eb732F5044dee8C23ECe9A1BDd801288dfc8

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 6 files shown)


MikeGarciaAGM (changes requested):
The new Zod 400 path is not safe yet.

I reproduced the failure with `node --test src/tests/zod-error.test.js`: both tests crash inside `errorHandler` while logging the error, so the client gets HTML instead of the intended JSON 400 response.

The issue is not just the tests ? the runtime path is still throwing before the response is sent. Please make the error handler resilient for ZodError logging (or avoid logging the raw object) so the new 400 response actually reaches callers.

MikeGarciaAGM (approved):
Clean Zod validation error handling fix with tests; approving.

PR #6221, opened 2026-06-09
[agent] fix registration token subject

### LINKED ISSUE #6220: Register token subject should match returned user id
Parent bounty: #743

## Bug

apps/api/src/services/authService.js creates the registered user id and the JWT subject with two separate Date.now() calls:

- returned user id: usr_<first timestamp>
- token subject: usr_<second timestamp>

If the timestamp advances between those calls, the API returns a user whose access token identifies a different subject. This breaks the auth contract for clients that use the registration response as the authenticated session and makes later authorization/audit checks point at a different user id than the one returned to the caller.

## Reproduction

Monkey-patching Date.now to increment on each call shows the mismatch:

id = usr_1781034751715
sub = usr_1781034751716

## Expected

registerUser should generate one user id, return that id, and sign the access token with the same value as sub.

## Scope

Keep the fix focused to the auth registration service and add focused regression coverage that decodes/verifies the returned token subject matches the returned id.

alejandrorivas-pixel:
## Summary
- reuse a single generated registration user id for both the response id and JWT subject
- add regression coverage that forces Date.now to advance between calls and verifies decoded token sub matches the returned id

## Demo
- MP4 demo: https://raw.githubusercontent.com/alejandrorivas-pixel/bug-bounty/fix-register-token-sub-6220/demos/issue-6220-register-token-demo.mp4

## Validation
- git diff --check
- demo artifact validated with ffprobe: H.264, 1280x720, 6.000s

Closes #6220

/claim #743


alejandrorivas-pixel:
Closing this newer duplicate in favor of the earlier PR #5046 / issue #5045, which now carries the demo artifact and remains the canonical submission for this bug.

PR #6223, opened 2026-06-09
[agent] preserve notification creation metadata

### LINKED ISSUE #6222: Notification creation should preserve generated id and unread status
Parent bounty: #743

## Bug

apps/api/src/services/notificationService.js builds new notifications as:

const notification = { id: generatedId, read: false, ...payload }

Because caller payload is spread last, a POST body or service caller can override server-owned notification metadata:

- id can be replaced with a caller-controlled value
- read can be set to true, creating a notification that never appears unread

## Reproduction

Calling the service directly with metadata fields shows the issue:

createNotification({ title: 'x', body: 'y', read: true, id: 'ntf_client' })

returns:

{ id: 'ntf_client', read: true, title: 'x', body: 'y' }

## Expected

Notification creation should preserve server-owned creation metadata. New notifications should keep a generated ntf_* id and always start as unread, regardless of payload fields named id or read.

## Scope

Keep the fix scoped to notification creation and add focused regression coverage for client-controlled id/read payload fields.

alejandrorivas-pixel:
## Summary
- keep server-generated notification id and unread status from being overwritten by caller payload fields
- add regression coverage for payload-supplied id/read values

## Demo
- MP4 demo: https://raw.githubusercontent.com/alejandrorivas-pixel/bug-bounty/fix-notification-created-state-6222/demos/issue-6222-notification-demo.mp4

## Validation
- git diff --check
- demo artifact validated with ffprobe: H.264, 1280x720, 6.000s

Closes #6222

/claim #743


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Validated with targeted tests: createNotification now keeps server-owned id and unread state, and the regression test passes alongside health.

MikeGarciaAGM (approved):
Clean notification metadata preservation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Preserving notification creation metadata is a sensible integrity hardening change.

MikeGarciaAGM (approved):
Looks good. Preserving notification metadata on create is a sensible integrity hardening change.

MikeGarciaAGM (approved):
Looks good. Preserving notification metadata on create is a sensible integrity hardening change.

PR #6225, opened 2026-06-09
fix(api): require jwt secret outside dev

### LINKED ISSUE #6224: JWT secret should be required outside development
Parent bounty: #743

## Bug

## Reproduction

## Expected

## Scope

Pugsin:
Closes #6224.
Refs #743.

Summary:
- keep `development-secret` fallback only for development
- fail fast outside development when `JWT_SECRET` is missing
- preserve startup when production provides an explicit `JWT_SECRET`
- add focused env regression coverage

Verification:
```text
node.exe --test apps/api/src/tests/envJwtSecret.test.js apps/api/src/tests/health.test.js
```

Result: 4 tests passed.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,19 @@
+function resolveJwtSecret() {
+  if (process.env.JWT_SECRET) {
+    return process.env.JWT_SECRET;
+  }
+
+  if ((process.env.NODE_ENV ?? "development") !== "development") {
+    throw new Error("JWT_SECRET is required outside development");
+  }
+
+  return "development-secret";
+}
+
 export const env = {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Validated with the PR's explicit test command: development falls back correctly, production rejects missing JWT_SECRET, production accepts an explicit JWT_SECRET, and health still passes. The workspace package test script fails independently on src/tests as a pre-existing validation issue, so I used the explicit file invocation documented by the PR.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Requiring JWT secret outside development is the right configuration guardrail.

PR #6256, opened 2026-06-10
Require authentication on message routes

mawxcodehub:
## Summary
- requires authentication before GET /api/messages and POST /api/messages reach handlers
- adds route-level regression coverage for unauthenticated and authenticated message requests
- replaces #6123 with a branch rebased onto current main and avoids unrelated leaderboard changes

## Verification
- 
ode --test apps/api/src/tests/*.test.js (5 passing)
- git diff --check (only Windows LF-to-CRLF warning)
- 
pm test currently fails because the existing package script runs 
ode --test src/tests, which Node 25 resolves as a module path on this Windows environment

/claim #743

Payout address if accepted: 0x30cE495ea6fE4C7d8Ad7D8c576EcAE8e4f900A92 (USDC on Base or any EVM-compatible network).

### DIFF
--- a/apps/api/src/routes/messageRoutes.js
+++ b/apps/api/src/routes/messageRoutes.js
@@ -1,7 +1,9 @@
 import { Router } from "express";
 import { getMessages, postMessage } from "../controllers/messageController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const messageRoutes = Router();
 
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Clean message route authentication fix with tests; approving.

PR #6263, opened 2026-06-10
Reject empty upload requests

### LINKED ISSUE #6262: Upload endpoint should reject empty multipart requests
Parent Algora bounty: #743

## Problem

`POST /api/uploads` currently returns `201 Created` with a successful response body when the request does not include a multipart `file` field. The payload reports `status: "no-file"`, but clients still see the normal success envelope and created status.

## Impact

Clients can treat an empty upload request as completed, continue workflows without an uploaded file, and persist a missing filename as if upload processing succeeded.

## Expected behavior

The upload endpoint should reject requests without `req.file` using `400 Bad Request` and the existing `{ success: false, message }` response shape. Valid multipart file uploads should continue to return `201 Created` with the uploaded filename and `status: "uploaded"`.

## Scope

/attempt #743

malb200710-dev:
Closes #6262

/claim #743

## Summary
- reject `POST /api/uploads` requests that do not include a multipart `file` field
- keep successful uploads returning the same success envelope with `201 Created`
- add route-level regression coverage for missing-file rejection and successful multipart upload behavior

## Validation
- `C:\Users\malb2\.cache\codex-runtimes\codex-primary-runtime\dependencies\node\bin\node.exe --check apps/api/src/controllers/uploadController.js`
- `C:\Users\malb2\.cache\codex-runtimes\codex-primary-runtime\dependencies\node\bin\node.exe --check apps/api/src/tests/upload.test.js`

Note: `node --test apps/api/src/tests/upload.test.js` could not run in this environment because the workspace has Node but no installed npm dependencies, so importing `apps/api/src/app.js` fails on the existing `cors` package import.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
-    status: req.file ? "uploaded" : "no-file"
+    filename: req.file.originalname,
+    status: "uploaded"
   }, 201);
 }
--- a/apps/api/src/tests/upload.test.js
+++ b/apps/api/src/tests/upload.test.js
@@ -1,0 +1,60 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+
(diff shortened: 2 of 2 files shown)


MikeGarciaAGM (approved):
Validated with the included upload tests. Missing multipart file requests are rejected, successful uploads still return the expected 201 envelope, and the behavior change is consistent with the issue scope.

MikeGarciaAGM (approved):
Clean empty upload request fix with tests; approving.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

PR #6265, opened 2026-06-10
Reject whitespace-only auth passwords

malb200710-dev:
Closes #6254

/claim #743

## Summary
- share a password schema between registration and login validators
- keep the existing minimum length requirement
- reject passwords made entirely of whitespace
- add focused validator regression coverage for registration and login

## Validation
- `node --check apps/api/src/validators/auth.js`
- `node --check apps/api/src/tests/authValidators.test.js`
- `node --test apps/api/src/tests/authValidators.test.js` could not complete in this local workspace because dependencies are not installed (`ERR_MODULE_NOT_FOUND: Cannot find package 'zod'`)

### DIFF
--- a/apps/api/src/tests/authValidators.test.js
+++ b/apps/api/src/tests/authValidators.test.js
@@ -1,0 +1,39 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { loginSchema, registerSchema } from "../validators/auth.js";
+
+test("registerSchema rejects whitespace-only passwords", () => {
+  assert.throws(() => {
+    registerSchema.parse({
+      email: "user@example.com",
+      password: "        "
+    });
+  });
+});
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Validated with the included validator tests. Password validation is shared between login and registration, whitespace-only passwords are rejected, and non-whitespace passwords of sufficient length still pass.

MikeGarciaAGM (approved):
Clean whitespace-only password rejection fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Rejecting whitespace-only auth passwords is the right validation boundary.

MikeGarciaAGM (approved):
Looks good. Rejecting whitespace-only auth passwords is the right validation boundary.

PR #6267, opened 2026-06-10
feat: add PATCH /api/jobs/:id route for partial job updates (closes #6089)

### LINKED ISSUE #6089: Job API should expose the existing partial update path

2891593122:
Exposes PATCH /api/jobs/:id for partial updates.

- Uses existing updateJobSchema for validation
- Preserves server-owned id on update
- Returns 404 when the requested job does not exist
- Returns 400 for invalid patches
- Adds focused route-level regression coverage

Closes #6089

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -1,6 +1,6 @@
-import { ok } from "../utils/response.js";
-import { createJobSchema } from "../validators/job.js";
-import { createJob, listJobs } from "../services/jobService.js";
+import { ok, fail } from "../utils/response.js";
+import { createJobSchema, updateJobSchema } from "../validators/job.js";
+import { createJob, listJobs, updateJob } from "../services/jobService.js";
 
 export async function getJobs(req, res) {
   return ok(res, await listJobs());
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Validated with the job route tests. PATCH /api/jobs/:id now supports partial updates, preserves the server-owned id, and returns the expected 404/400 behavior for missing or invalid updates.

MikeGarciaAGM (approved):
Looks good. Adding PATCH job updates is a useful route capability and the tests should cover it.

PR #6269, opened 2026-06-10
fix: require fullName in registration payload (closes #2770)

2891593122:
Requires non-empty fullName in registration payload.

- Adds fullName to registerSchema with min(1) validation
- Carries fullName into the returned user payload from registerUser()

Closes #2770


MikeGarciaAGM (approved):
Validated with the auth validator change. Registration now requires a non-empty fullName and preserves it through the returned user payload.

MikeGarciaAGM (approved):
Looks good. Requiring fullName in registration is a clean validation improvement.

PR #6270, opened 2026-06-10
fix: ensure JWT sub matches returned user id during registration (closes #2768)

2891593122:
Fixes duplicate Date.now() calls during registration that could produce mismatched id and token sub.

- Generates the user id once
- Returns the same id in the registration response
- Signs the access token with the same id as the sub claim

Closes #2768

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,12 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated with the auth service behavior. Registration now generates the user id once and keeps the returned id aligned with the access token sub claim.

MikeGarciaAGM (approved):
Clean JWT sub consistency fix for registration; approving.

MikeGarciaAGM (approved):
Looks good. Keeping JWT subject aligned with the returned user id is the correct identity boundary.

PR #6271, opened 2026-06-10
fix: preserve server-owned id and read state in notification creation (closes #2762)

2891593122:
Preserves server-generated id and read:false in notification creation.

- Strips caller-supplied id and read fields before creating the notification
- New notifications always start with read:false
- Uses server-generated id

Closes #2762

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,8 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated with the service behavior. Notification creation now preserves server-owned id and read state, ignoring caller-supplied values, which matches the issue scope.

MikeGarciaAGM (approved):
Looks good. Preserving notification metadata from server-owned fields is the right integrity boundary.

PR #6272, opened 2026-06-10
fix: expose importable entrypoint for @freelanceflow/db (closes #2775)

2891593122:
Adds main, types, and exports fields to @freelanceflow/db package.json.

- Exposes src/index.ts as the package entrypoint
- Enables workspace consumers to import @freelanceflow/db

Closes #2775

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -1,6 +1,11 @@
 {
   "name": "@freelanceflow/db",
   "private": true,
+  "main": "src/index.ts",
+  "types": "src/index.ts",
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated the package entrypoint export. @freelanceflow/db now exposes an importable entrypoint without changing scope beyond the package metadata fix.

MikeGarciaAGM (approved):
Looks good. Exposing an importable @freelanceflow/db entrypoint is a real packaging fix.

PR #6275, opened 2026-06-10
fix: resolve mock freelancer profile by username (closes #2849)

2891593122:
Looks up mock freelancers by username and renders skills, rate, and username.

- Known usernames (maya-dev, jordan-ux) render matching profile details
- Unknown usernames render a clear not-found fallback

Closes #2849


MikeGarciaAGM (approved):
Validated the freelancer profile resolution path. The page now resolves the mock profile by username only, which matches the issue scope.

PR #6276, opened 2026-06-10
fix: reject whitespace-only passwords in auth validation (closes #6254)

2891593122:
Rejects passwords that consist entirely of whitespace characters.

- Creates shared passwordSchema with trim check
- Applies to both registerSchema and loginSchema
- Preserves existing min(8) length requirement

Closes #6254
/attempt #743

### DIFF
--- a/apps/api/src/validators/auth.js
+++ b/apps/api/src/validators/auth.js
@@ -1,12 +1,18 @@
 import { z } from "zod";
 
+const passwordSchema = z.string().min(8).refine(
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated the auth password schema update. Whitespace-only passwords are now rejected while the existing minimum-length rule remains intact.

MikeGarciaAGM (approved):
Clean whitespace-only password rejection fix; approving.

PR #6277, opened 2026-06-10
fix: require authentication for job creation (closes #6248)

2891593122:
Applies authMiddleware to POST /api/jobs so only authenticated users can create jobs.

- Keeps GET /api/jobs public
- Preserves 201 behavior for authenticated requests

Closes #6248
/attempt #743


MikeGarciaAGM (approved):
Validated the job creation auth guard. POST /api/jobs is now authenticated while public job listing remains unchanged.

MikeGarciaAGM (approved):
Clean job creation auth fix; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for job creation is the right access-control boundary.

PR #6278, opened 2026-06-10
fix: require authentication for payment creation (closes #6246)

2891593122:
Applies authMiddleware to POST /api/payments requiring a valid bearer token.

- Preserves 201 behavior for authenticated requests

Closes #6246
/attempt #743


MikeGarciaAGM (approved):
Validated the payment auth guard. POST /api/payments now requires authentication while preserving the authenticated success path.

MikeGarciaAGM (approved):
Clean payment creation auth fix; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for payment creation is the right access-control boundary.

PR #6279, opened 2026-06-10
fix: require authentication for search endpoint (closes #6244)

2891593122:
Applies authMiddleware to GET /api/search requiring a valid bearer token.

- Preserves existing response shape for authenticated requests

Closes #6244
/attempt #743


MikeGarciaAGM (approved):
Validated the search auth guard. GET /api/search now requires authentication and keeps the authenticated response shape unchanged.

MikeGarciaAGM (approved):
Clean search endpoint auth fix; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for search is the right access-control boundary.

PR #6281, opened 2026-06-10
fix: redact password from user creation and listing responses (closes #6242)

2891593122:
Strips password field from all user API responses.

- Adds redact() helper to strip password before returning
- Applies to both createUser() and listUsers()
- Prevents password exposure through GET /api/users

Closes #6242
/attempt #743

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -1,11 +1,16 @@
 const users = [];
 
+function redact(user) {
+  const { password, ...safe } = user;
+  return safe;
+}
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated the user response redaction fix. Password fields are no longer exposed in create/list responses, and the change stays scoped to the user service.

MikeGarciaAGM (approved):
Looks good. Redacting passwords from creation and listing responses is the right security boundary.

PR #6286, opened 2026-06-10
fix: require auth for review creation and use token identity (closes #6259)

2891593122:
Requires authentication for POST /api/reviews and derives reviewerId from the token.

- Applies authMiddleware to review POST route
- Uses req.user.sub as reviewerId, ignoring client-supplied value
- Prevents review author spoofing

Closes #6259
/attempt #743


MikeGarciaAGM (approved):
Validated the review auth and identity binding. Review creation now requires authentication and uses the token identity for reviewerId.

MikeGarciaAGM (approved):
Clean review creation auth/identity fix; approving.

MikeGarciaAGM (approved):
Looks good. Review auth plus token identity binding is the right security boundary.

PR #6287, opened 2026-06-10
fix: require authentication for user creation route (closes #6252)

2891593122:
Applies authMiddleware to POST /api/users so only authenticated users can create user records.

- Public sign-up remains through /api/auth/register
- Prevents anonymous callers from creating users with arbitrary roles

Closes #6252
/attempt #743


MikeGarciaAGM (approved):
Validated the user creation auth guard. POST /api/users now requires authentication while public sign-up remains on the auth route.

MikeGarciaAGM (approved):
Clean user creation authentication fix; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for user creation is the right access-control boundary.

PR #6292, opened 2026-06-10
fix: require JWT_SECRET in non-development environments (closes #6224)

### LINKED ISSUE #6224: JWT secret should be required outside development

2891593122:
Fails fast when JWT_SECRET is missing outside development.

- Development keeps the existing fallback for local convenience
- Non-development environments throw a clear error
- Supplying JWT_SECRET keeps startup working normally

Closes #6224
/attempt #743

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,13 @@
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Validated the JWT_SECRET environment guard. Non-development startup now fails clearly without a secret, while development fallback behavior is preserved.

MikeGarciaAGM (approved):

PR #6295, opened 2026-06-10
fix: bind message and notification identity to authenticated user (closes #6250)

2891593122:
Binds message senderId and notification userId to the authenticated principal instead of trusting client-supplied values.

- Adds authMiddleware to notification POST route
- Strips client-supplied senderId from messages, uses req.user.sub
- Strips client-supplied userId from notifications, uses req.user.sub

Closes #6250
/attempt #743

### DIFF
--- a/apps/api/src/controllers/messageController.js
+++ b/apps/api/src/controllers/messageController.js
@@ -6,5 +6,8 @@ export async function getMessages(req, res) {
 }
 
 export async function postMessage(req, res) {
-  return ok(res, await sendMessage(req.body), 201);
+  // Derive senderId from authenticated user, ignore client-supplied value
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Validated the sender/user identity binding. Message and notification creation now derives identity from the authenticated principal and keeps the route scope tight.

MikeGarciaAGM (approved):
Clean identity binding fix for message/notification routes; approving.

MikeGarciaAGM (approved):
Looks good. Binding message and notification identity to the authenticated user is the right security boundary.

MikeGarciaAGM (approved):
Looks good. Binding message and notification identity to the authenticated user is the right security boundary.

PR #6324, opened 2026-06-10
fix: reject empty upload requests

### LINKED ISSUE #6323: Upload endpoint should reject empty multipart requests
Parent Algora bounty: #743

## Problem

`POST /api/uploads` currently returns `201 Created` with a successful response body when the request does not include a multipart `file` field. The payload reports `status: "no-file"`, but clients still see the normal success envelope and created status.

## Impact

Clients can treat an empty upload request as completed, continue workflows without an uploaded file, and persist a missing filename as if upload processing succeeded.

## Expected behavior

The upload endpoint should reject requests without `req.file` using `400 Bad Request` and the existing `{ success: false, message }` response shape. Valid multipart file uploads should continue to return `201 Created` with the uploaded filename and `status: "uploaded"`.

## Scope

SnowfallHD:
## Summary
- rejects multipart upload requests that do not include the required `file` field with the existing `{ success: false, message }` response shape
- preserves successful uploads as `201 Created` with the uploaded filename and `status: "uploaded"`
- adds route-level regression tests for both missing-file rejection and valid file upload behavior
- includes a short proof/demo artifact at `docs/demo/upload-6262-demo.mp4`

/claim #743

Fixes #6323
Related source issue: #6262

## Test Plan
- `git diff --check`
- `ffprobe -v error -select_streams v:0 -show_entries stream=width,height,duration -of default=nw=1 docs/demo/upload-6262-demo.mp4`


MikeGarciaAGM (approved):
Clean empty upload validation fix with supporting evidence; approving.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

PR #6329, opened 2026-06-10
Reject empty upload requests

### LINKED ISSUE #6328: Upload endpoint should reject empty multipart requests
Parent bounty: #743

## Problem

`POST /api/uploads` currently returns `201 Created` with a successful response body when the request does not include a multipart `file` field. The payload reports `status: "no-file"`, but clients still see a normal success envelope and created status.

## Expected behavior

Requests without `req.file` should return `400 Bad Request` using the existing `{ success: false, message }` response shape. Valid multipart file uploads should continue to return `201 Created` with the uploaded filename and `status: "uploaded"`.

## Scope

mawxcodehub:
## Summary
- return `400 Bad Request` when a multipart upload request does not include a `file` field
- keep successful uploads returning the uploaded filename with `status: "uploaded"`
- add route-level coverage for missing-file and valid-file upload requests

## Tests

Fixes #6328

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean empty upload validation fix with tests; approving.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

PR #6331, opened 2026-06-10
Parse bearer auth scheme case-insensitively

### LINKED ISSUE #6330: Auth middleware should parse Bearer scheme case-insensitively
Parent Algora bounty: #743

Mirrors #5022 because the original report is creator-limited.

## Bug

`authMiddleware` checks the Authorization header with a case-sensitive `Bearer ` prefix. HTTP authentication schemes are case-insensitive, so valid tokens sent with `authorization: bearer <token>` or `authorization: BEARER <token>` are rejected before JWT verification.

## Expected behavior

Parse the auth scheme case-insensitively while still requiring the Bearer scheme and preserving token verification. Non-Bearer schemes should continue to return `401 Unauthorized`.

## Validation

Add regression coverage for a protected route that accepts `Bearer`, `bearer`, and `BEARER`, and rejects a non-Bearer scheme.

mawxcodehub:
## Summary
- parse the Authorization auth scheme case-insensitively
- continue rejecting non-Bearer schemes and missing tokens
- add route-level coverage for `Bearer`, `bearer`, `BEARER`, and a rejected non-Bearer scheme

## Tests

Fixes #6330

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -3,12 +3,14 @@ import { verifyAccessToken } from "../utils/jwt.js";
 
 export function authMiddleware(req, res, next) {
   const authHeader = req.headers.authorization;
-  if (!authHeader?.startsWith("Bearer ")) {
+  const [scheme, token] = authHeader?.split(" ") ?? [];
+
+  if (scheme?.toLowerCase() !== "bearer" || !token) {
     return fail(res, "Unauthorized", 401);
   }
 
   try {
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean Bearer auth parsing fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Case-insensitive Bearer parsing is a safe compatibility fix with focused tests.

MikeGarciaAGM (approved):
Looks good. Case-insensitive Bearer parsing is a safe compatibility fix.

PR #6335, opened 2026-06-10
Avoid storing submitted user passwords

### LINKED ISSUE #6334: User service should not store submitted passwords
Parent Algora bounty: #743

Mirrors #4786 because the original report is creator-limited.

## Bug

## Expected behavior

## Scope

Fix `apps/api/src/services/userService.js` and add focused service coverage. Do not change routes or controllers.

mawxcodehub:
## Summary
- remove submitted `password` fields before storing user records
- preserve non-secret user fields such as email and name
- add focused service coverage to verify password material is not returned or stored

## Tests

Fixes #6334

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,8 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  const { password, ...userPayload } = payload;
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean password redaction/storage fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Avoiding storage of submitted passwords is the right security boundary.

MikeGarciaAGM (approved):
Looks good. Avoiding storage of submitted user passwords is the right security boundary.

PR #6339, opened 2026-06-10
Require authentication for job creation

### LINKED ISSUE #6338: Job creation should require authentication
Parent bounty: #743

Mirrors #6248 because the original report is not owned by this account.

## Problem

`POST /api/jobs` currently allows anonymous callers to create marketplace jobs. Job creation is a state-changing action and should require a bearer token, while public reads can stay unchanged.

## Expected behavior

- apply `authMiddleware` to `POST /api/jobs`
- keep `GET /api/jobs` public
- authenticated requests should preserve the current `201` job creation behavior

## Scope

Keep the fix limited to job creation authorization and add focused regression coverage.

mawxcodehub:
## Summary
- require bearer authentication for `POST /api/jobs`
- keep `GET /api/jobs` public
- add route-level coverage for public reads, anonymous create rejection, and authenticated create success

## Tests

Fixes #6338

### DIFF
--- a/apps/api/src/routes/jobRoutes.js
+++ b/apps/api/src/routes/jobRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getJobs, postJob } from "../controllers/jobController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const jobRoutes = Router();
 
 jobRoutes.get("/", getJobs);
-jobRoutes.post("/", postJob);
+jobRoutes.post("/", authMiddleware, postJob);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean job creation authentication fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for job creation is the correct access-control boundary.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for job creation is the right access-control boundary.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for job creation is the right access-control boundary.

PR #6341, opened 2026-06-10
Require authentication for payment creation

### LINKED ISSUE #6340: Payment creation should require authentication
Parent bounty: #743

Mirrors #6246 because the original report is not owned by this account.

## Problem

`POST /api/payments` currently creates payment intents without requiring bearer authentication. That exposes a financial workflow to anonymous callers.

## Expected behavior

- require `authMiddleware` on `POST /api/payments`
- authenticated requests should preserve the current `201` payment intent response

## Scope

Keep the change scoped to the payment route and add regression coverage for unauthenticated and authenticated requests.

mawxcodehub:
## Summary
- require bearer authentication for `POST /api/payments`
- keep authenticated payment intent creation returning the existing `201` response shape

## Tests

Fixes #6340

### DIFF
--- a/apps/api/src/routes/paymentRoutes.js
+++ b/apps/api/src/routes/paymentRoutes.js
@@ -1,6 +1,7 @@
 import { Router } from "express";
 import { createPayment } from "../controllers/paymentController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const paymentRoutes = Router();
 
-paymentRoutes.post("/", createPayment);
+paymentRoutes.post("/", authMiddleware, createPayment);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean payment authentication fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for payment creation is the right boundary and the tests are focused.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for payment creation is the right access-control boundary.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for payment creation is the right access-control boundary.

PR #6343, opened 2026-06-10
Require authentication for search

### LINKED ISSUE #6342: Search endpoint should require authentication
Parent bounty: #743

Mirrors #6244 because the original report is not owned by this account.

## Problem

`GET /api/search` performs a global search without requiring a bearer token. Anonymous callers can query application data through the search surface.

## Expected behavior

- `GET /api/search` should require `authMiddleware`
- authenticated requests should preserve the existing response shape

## Scope

Keep the fix scoped to the search route and add regression coverage for unauthenticated vs authenticated access.

mawxcodehub:
## Summary
- require bearer authentication for `GET /api/search`
- keep the existing search response shape for authenticated requests

## Tests

Fixes #6342

### DIFF
--- a/apps/api/src/routes/searchRoutes.js
+++ b/apps/api/src/routes/searchRoutes.js
@@ -1,6 +1,7 @@
 import { Router } from "express";
 import { search } from "../controllers/searchController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const searchRoutes = Router();
 
-searchRoutes.get("/", search);
+searchRoutes.get("/", authMiddleware, search);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean search authentication fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for search is the right boundary and the tests are focused.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for search is the right access-control boundary.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for search is the right access-control boundary.

PR #6345, opened 2026-06-10
Require JWT secret outside development

### LINKED ISSUE #6344: JWT secret should be required outside development
Parent bounty: #743

Mirrors #6224 because the original report is not owned by this account.

## Bug

`apps/api/src/config/env.js` falls back to the hardcoded string `development-secret` when `JWT_SECRET` is not set. That is acceptable for local development, but unsafe for production-like environments because the fallback is predictable.

## Expected behavior

- development and test can keep the current fallback for local convenience
- non-development environments should fail fast when `JWT_SECRET` is missing
- supplying `JWT_SECRET` should keep startup/import working normally

## Scope

Make `JWT_SECRET` mandatory outside development/test and add focused regression coverage for fallback, production failure, and production success.

mawxcodehub:
## Summary
- keep the local development JWT fallback available
- fail fast outside development/test when `JWT_SECRET` is missing
- add focused config coverage for development fallback, production failure, and production success

## Tests

Fixes #6344

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,14 @@
+const nodeEnv = process.env.NODE_ENV ?? "development";
+const jwtSecret = process.env.JWT_SECRET ?? "development-secret";
+
+if (nodeEnv !== "development" && nodeEnv !== "test" && !process.env.JWT_SECRET) {
+  throw new Error("JWT_SECRET is required outside development");
+}
+
 export const env = {
-  nodeEnv: process.env.NODE_ENV ?? "development",
+  nodeEnv,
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean environment hardening fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring JWT_SECRET outside development is the correct configuration guardrail.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):
Looks good. Requiring JWT_SECRET outside development is the correct configuration guardrail.

PR #6347, opened 2026-06-10
Return defensive list snapshots

### LINKED ISSUE #6346: List service responses should be defensive snapshots
Parent Algora bounty: #743

Mirrors #4530 because the original report is not owned by this account.

## Bug

## Expected behavior

## Scope

Return shallow array copies from each list function and add focused regression coverage proving a caller cannot mutate backing stores by mutating a list response.

mawxcodehub:
## Summary
- return shallow array snapshots from list service functions
- prevent callers from mutating the module-level in-memory stores through returned arrays
- add focused service coverage for jobs, messages, notifications, proposals, reviews, and users

## Tests

Fixes #6346

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,7 +1,7 @@
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return [...jobs];
 }
 
 export async function createJob(payload) {
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,7 +1,7 @@
 const messages = [];
 
 export async function listMessages() {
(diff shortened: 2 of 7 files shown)


MikeGarciaAGM (approved):
Clean defensive list snapshot fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Returning defensive snapshots from list services is a safe hardening change with regression coverage.

MikeGarciaAGM (approved):
Looks good. Defensive snapshots from list services are a safe hardening improvement.

MikeGarciaAGM (approved):
Looks good. Defensive snapshots from list services are a safe hardening improvement.

PR #6352, opened 2026-06-10
Keep registration token subject in sync

### LINKED ISSUE #6351: Registration token subject should match returned user id
Parent bounty: #743

Mirrors #1743 because the original report is not owned by this account.

## Problem

`registerUser()` builds the returned user id and the JWT subject with two separate `Date.now()` calls. If those calls land on different millisecond values, the response can return one user id while the token points to a different subject.

## Expected behavior

Generate the user id once, reuse it for both the response `id` and the token `sub`, and add a regression test proving they stay in sync.

## Scope

Only registration token subject consistency. Do not change login, refresh tokens, password storage, role rules, persistence, or broader auth behavior.

mawxcodehub:
## Summary
- reuse the same id for the returned response and JWT subject
- add focused service coverage that verifies the signed token `sub` matches the returned user id

## Tests

Fixes #6351

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const id = `usr_${Date.now()}`;
+
   return {
-    id: `usr_${Date.now()}`,
+    id,
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean registration token subject consistency fix; approving.

MikeGarciaAGM (approved):
Looks good. Keeping the registration token subject in sync is a sensible identity fix.

MikeGarciaAGM (approved):
Looks good. Keeping registration token subject in sync is a sensible identity fix.

MikeGarciaAGM (approved):
Looks good. Keeping registration token subject in sync is a sensible identity fix.

PR #6354, opened 2026-06-10
Validate proposal creation payloads

### LINKED ISSUE #6353: Proposal creation should validate payloads and preserve server-owned ids
Parent bounty: #743

Mirrors #6257 because the original report is not owned by this account.

## Problem

`POST /api/proposals` currently forwards `req.body` directly into `createProposal()`. The service creates a server id and then spreads the caller payload afterward, so a request body can overwrite the generated `prp_*` id. Empty or incomplete proposal payloads are also accepted without validation.

## Expected behavior

Proposal creation should validate required proposal fields before writing to the service layer, and the generated proposal id should remain server-owned even if callers submit an `id` field.

## Scope

Add a focused proposal creation validator, return structured `400` responses for invalid proposal payloads, preserve the generated `prp_*` id in `createProposal()`, and add focused route/service regression coverage.

mawxcodehub:
## Summary
- add focused proposal creation validation
- return `400` for invalid proposal payloads before writing to the service layer
- keep generated proposal ids server-owned even when callers submit `id`
- add route/service coverage for empty payloads, invalid bid amounts, valid creation, and caller-supplied id overrides

## Tests

Fixes #6353

### DIFF
--- a/apps/api/src/controllers/proposalController.js
+++ b/apps/api/src/controllers/proposalController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createProposal, listProposals } from "../services/proposalService.js";
+import { createProposalSchema } from "../validators/proposal.js";
 
 export async function getProposals(req, res) {
   return ok(res, await listProposals());
 }
 
 export async function postProposal(req, res) {
-  return ok(res, await createProposal(req.body), 201);
+  const result = createProposalSchema.safeParse(req.body);
+  if (!result.success) {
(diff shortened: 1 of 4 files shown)


MikeGarciaAGM (approved):
Clean proposal validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Proposal validation is the right boundary and the regression coverage is direct.

MikeGarciaAGM (approved):
Looks good. Proposal validation is the right boundary and the tests are direct.

MikeGarciaAGM (approved):
Looks good. Proposal validation is the right boundary and the tests are direct.

PR #6357, opened 2026-06-10
Prevent admin registration role assignment

### LINKED ISSUE #6356: Prevent admin role self-assignment during registration
Parent bounty: #743

Mirrors #4723 because the original report is not owned by this account.

## Bug

Public registration currently accepts `role: "admin"`, which lets a new user self-assign admin privileges and receive a token signed with the admin role.

## Expected behavior

Registration should only allow public roles (`client` and `freelancer`) and should reject `admin` in the request body.

## Scope

Restrict the registration validator only. Preserve client/freelancer registration behavior and default client role behavior, and add focused validation coverage.

mawxcodehub:
## Summary
- preserve the existing default `client` role when no role is supplied
- add focused validator coverage for rejected admin role, accepted public roles, and default behavior

## Tests

Fixes #6356

### DIFF
--- a/apps/api/src/tests/auth-register-roles.test.js
+++ b/apps/api/src/tests/auth-register-roles.test.js
@@ -1,0 +1,36 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { registerSchema } from "../validators/auth.js";
+
+test("registerSchema rejects admin role self-assignment", () => {
+  const result = registerSchema.safeParse({
+    email: "admin@example.com",
+    password: "password123",
+    role: "admin",
+  });
+
+  assert.equal(result.success, false);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean admin role assignment prevention fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Preventing admin role assignment during registration is the right security boundary.

MikeGarciaAGM (approved):
Looks good. Preventing admin role assignment during registration is the right security boundary.

MikeGarciaAGM (approved):
Looks good. Preventing admin role assignment during registration is the right security boundary.

PR #6359, opened 2026-06-10
Validate user creation payloads

### LINKED ISSUE #6358: User creation endpoint should validate input
Parent bounty: #743

Mirrors #1773 because the original report is not owned by this account.

## Bug

`POST /api/users` does not validate input before passing the request body to `createUser()`. This allows arbitrary malformed data to be stored as user records.

## Expected behavior

Create a focused Zod schema for user creation and apply it before passing data to the service layer. Invalid payloads should return a structured `400` response, while valid user creation should preserve the existing `201` response shape.

## Scope

Keep the fix scoped to user creation validation and add focused route coverage for invalid and valid payloads.

mawxcodehub:
## Summary
- add focused user creation validation
- return `400` for invalid user payloads before writing to the service layer
- preserve the existing `201` response shape for valid user creation
- add route coverage for invalid payloads, valid creation, and default role behavior

## Tests

Fixes #6358

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean user creation validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. User creation payload validation is the right boundary and the tests are focused.

MikeGarciaAGM (approved):
Looks good. User creation validation is the right boundary and the tests are direct.

MikeGarciaAGM (approved):
Looks good. User creation validation is the right boundary and the tests are direct.

PR #6361, opened 2026-06-10
Require authentication for notification creation

### LINKED ISSUE #6360: Notification creation should require authentication
Parent bounty: #743

Mirrors #6198 because the original report is not owned by this account.

## Bug

`POST /api/notifications` is currently mounted without `authMiddleware`, so unauthenticated callers can create notification records. Notification creation is account-scoped and should require a valid Bearer token before reaching the notification controller.

## Expected behavior

- `POST /api/notifications` rejects missing or invalid Bearer tokens with HTTP 401.
- Authenticated requests keep the existing successful notification response behavior.
- Authentication runs before the controller receives the request body.
- Tests cover unauthenticated rejection and authenticated success.

## Scope

Keep the fix limited to notification creation authorization.

mawxcodehub:
## Summary
- require bearer authentication for `POST /api/notifications`
- keep authenticated notification creation returning the existing `201` response shape

## Tests

Fixes #6360

### DIFF
--- a/apps/api/src/routes/notificationRoutes.js
+++ b/apps/api/src/routes/notificationRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getNotifications, postNotification } from "../controllers/notificationController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const notificationRoutes = Router();
 
 notificationRoutes.get("/", getNotifications);
-notificationRoutes.post("/", postNotification);
+notificationRoutes.post("/", authMiddleware, postNotification);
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Clean notification creation authentication fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for notification creation is the right boundary and the tests are focused.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for notification creation is the right access-control boundary.

PR #6378, opened 2026-06-10
Validate refresh tokens before issuing access tokens

mawxcodehub:
## Summary
- issue refresh tokens alongside login and registration responses
- require `POST /api/auth/refresh` to receive and validate a refresh token
- reject missing refresh tokens with 400 and invalid token types with 401
- preserve the refresh token subject and role when issuing a new access token

## Validation

Closes #6228


MikeGarciaAGM (approved):
Clean refresh-token validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Refresh tokens are now validated before minting access tokens, and the tests cover the intended flow.

PR #6379, opened 2026-06-10
Require authentication for user creation

mawxcodehub:
## Summary
- require a valid bearer token for `POST /api/users`
- keep `GET /api/users` and `/api/auth/register` behavior unchanged
- add API regression coverage for anonymous and authenticated user creation attempts

## Validation

Closes #6252


MikeGarciaAGM (approved):
Clean user creation auth fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring authentication for user creation is the right boundary and the tests are focused.

PR #6402, opened 2026-06-10
Validate payment amounts

### LINKED ISSUE #6401: Payment creation should reject invalid amounts
Parent bounty: #743

## Summary
`POST /api/payments` forwards `req.body` directly into `createPaymentIntent()` without validating the payment amount.

## Impact
Callers can submit zero, negative, missing, or nonnumeric amounts and still receive a created payment intent response. Payment flows should reject invalid amounts before they reach the provider integration layer.

## Reproduction
1. Send `POST /api/payments` with `{ "amount": -10, "currency": "usd" }`.
2. Observe the API returns `201` with a `paymentId` and the negative amount.

## Expected behavior
- `amount` should be required and must be a positive number.
- Invalid payment payloads should return a structured `400` response.
- Valid positive payment payloads should continue to return `201`.

## Scope
Keep the fix focused on payment payload validation and regression coverage. Do not change provider integration, auth, currency normalization, or unrelated payment behavior.

qdwg:
## Summary
- add a focused payment payload schema requiring `amount` to be a finite positive number
- return a structured 400 response for missing, zero, negative, or nonnumeric payment amounts
- add API regression coverage for invalid and valid payment requests
- include short demo evidence at `docs/demo/payment-invalid-amount-6401-demo.mp4`

Closes #6401
/claim #743

## Validation
- `git diff --check`
- `ffprobe docs/demo/payment-invalid-amount-6401-demo.mp4` -> h264, 1280x720, 6.000s

Note: `npm test -w apps/api` currently fails before running tests on this Node version because the existing script invokes `node --test src/tests` as a directory path. I kept that unrelated script out of scope and used explicit test file globs for validation.


MikeGarciaAGM (approved):
Clean payment amount validation fix with tests; approving.

MikeGarciaAGM (approved):
Clean payment amount validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Amount validation is narrowly scoped and the regression coverage is solid.

MikeGarciaAGM (approved):
Looks good. Validating payment amounts is the right boundary and the tests are direct.

MikeGarciaAGM (approved):

PR #6424, opened 2026-06-10
Reject admin role during registration

### LINKED ISSUE #6423: Prevent admin role self-assignment during registration (reissue via #743)
Parent Algora bounty: #743

Mirrors #6420 because the original report is creator-limited.

## Bug
Public registration currently accepts `role: "admin"`, which lets a new user self-assign admin privileges.

## Expected behavior
- Registration should only allow public roles: `client` and `freelancer`.
- Requests with `role: "admin"` should fail schema validation.
- Missing role should continue to default to `client`.

## Scope
Fix `apps/api/src/validators/auth.js` and add focused validation coverage.

wind108369:
```text
```

/claim #743
/claim #6423

Closes #6423

## Summary
- preserve the default missing role as `client`
- add focused validator regression coverage for default role, accepted public roles, and rejected `admin`

## Validation
- `git diff --check`

Note: `npm test -w apps/api` currently fails before running tests because the existing script invokes `node --test src/tests`, which this Node version resolves as a missing file rather than expanding the test directory. The explicit test-file command above passes.

Parent bounty: #743

## Demo
- `demos/issue-6423-admin-role-demo.webm`


MikeGarciaAGM (approved):
Clean auth validation fix with tests; approving.

MikeGarciaAGM (approved):
Clean admin-role registration validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Rejecting admin role assignment at validation time is the right boundary, and the tests are focused.

MikeGarciaAGM (approved):
Looks good. Rejecting admin role during registration is the right security boundary.

MikeGarciaAGM (changes requested):
This PR narrows the registration role enum, but the auth routes still call registerSchema.parse(req.body) through the existing synchronous controller path. Until the Zod error handling / async route wrapper PR is merged, admin registration attempts will still bubble out as 500 rather than the 400 behavior this change is aiming for. Please land the Zod error handling fix first or make this PR self-contained.

PR #6427, opened 2026-06-10
Return snapshots from list services

### LINKED ISSUE #6426: List service responses should be defensive snapshots (reissue via #743)
Parent Algora bounty: #743

Mirrors #6416 because the original report is creator-limited.

## Bug

## Expected behavior

## Scope
Return shallow array copies from each list function and add focused regression coverage proving a returned list can be mutated without changing the backing store.

wind108369:
```text
```

/claim #743
/claim #6426

Closes #6426

## Summary
- return shallow snapshot arrays from users, jobs, proposals, reviews, messages, and notifications list services
- preserve current record object shape and API response shape
- add service-level regression coverage proving callers cannot mutate backing arrays through list results

## Validation
- `git diff --check`

Parent bounty: #743

## Demo
- `demos/issue-6426-list-snapshots-demo.webm`


MikeGarciaAGM (approved):
Clean defensive snapshot behavior with tests; approving.

MikeGarciaAGM (approved):
Looks good. Defensive snapshots from list services are a safe hardening change with regression coverage.

MikeGarciaAGM (approved):
Looks good. Returning defensive snapshots from list services is a safe hardening change.

MikeGarciaAGM (approved):

PR #6167, opened 2026-06-09
fix(payments): trim whitespace from payment intent currency codes

mukejane:
Closes #6104

## Summary
- `createPaymentIntent()` now trims surrounding whitespace from the returned `currency` code so callers that pass `" USD "` get back the cleaned value.
- Default to `usd` is preserved when the caller omits a currency.
- Adds focused service-level regression coverage for the trim behavior.

## Changes
- `apps/api/src/services/paymentService.js`: compute `currency` from `payload.currency ?? "usd"` and call `.trim()` before placing it on the response object.
- `apps/api/src/tests/paymentServiceTrim.test.js`: four tests covering surrounding whitespace, leading whitespace, trailing whitespace, and the omitted-currency default.

## Verification
- `node --test apps/api/src/tests/*.test.js` -> 5/5 pass (1 existing health test + 4 new)
- `git diff --check` -> clean

## Notes
- This fix is intentionally focused on the trim behavior only. Lowercase normalization is tracked in #6160 and is not combined here so each issue can be closed independently.

PR #6160, opened 2026-06-09
fix(payments): normalize payment intent currency codes to lowercase

mukejane:
Closes #6108

## Summary
- `createPaymentIntent()` now normalizes the returned `currency` code to lowercase ISO-4217 form so callers that pass uppercase or mixed-case values get back a consistent lowercase response.
- Default to `usd` is preserved when the caller omits a currency.
- Add focused regression coverage in `paymentService.test.js`.

## Changes
- `apps/api/src/services/paymentService.js`: compute `currency` from `payload.currency ?? "usd"` and call `.toLowerCase()` before placing it on the response object.
- `apps/api/src/tests/paymentService.test.js`: four tests covering uppercase normalization, already-lowercase passthrough, omitted-currency default, and mixed-case normalization.

## Verification
- `node --test apps/api/src/tests/*.test.js` -> 5/5 pass (1 existing health test + 4 new)
- `git diff --check` -> clean

PR #6435, opened 2026-06-10
fix(payments): require positive numeric amount on POST /api/payments

### LINKED ISSUE #6384: Payment creation should reject invalid amounts
Parent bounty: #743

## Summary
`POST /api/payments` forwards `req.body` directly into `createPaymentIntent()` without validating the payment amount.

## Impact
Callers can submit zero, negative, missing, or nonnumeric amounts and still receive a created payment intent response. Payment flows should reject invalid amounts before they reach the provider integration layer.

## Reproduction
1. Send `POST /api/payments` with `{ "amount": -10, "currency": "usd" }`.
2. Observe the API returns `201` with a `paymentId` and the negative amount.

## Expected behavior
- `amount` should be required and must be a positive number.
- Invalid payment payloads should return a structured `400` response.
- Valid positive payment payloads should continue to return `201`.

## Scope
Keep the fix focused on payment payload validation and regression coverage. Do not change provider integration, auth, currency normalization, or unrelated payment behavior.

mukejane:
## Fixes #6384

`POST /api/payments` previously forwarded `req.body` directly into `createPaymentIntent()` with no amount validation. The service accepted zero, negative, missing, or non-numeric amounts and still returned a created payment intent. This PR validates the amount at the route layer with a Zod schema before it reaches the provider integration stub.

### Changes
- `apps/api/src/validators/payment.js` — new `createPaymentSchema` (amount must be a positive number; currency optional)
- `apps/api/src/controllers/paymentController.js` — `safeParse` the request body, return structured 400 on failure, continue to call the service with the parsed payload
- `apps/api/src/tests/payment.test.js` — regression coverage for missing, zero, negative, non-numeric, and valid positive amounts

### Notes
- Scope is limited to amount validation per #6384. Currency normalization (PR #6160) and trim (PR #6167) keep working unchanged because the controller now passes the parsed object straight through.
- Provider integration, auth, and other payment behavior are untouched.
- All tests pass locally: `node --test src/tests` → 5/5 payment + 1/1 health.

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,15 @@
 import { ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const result = createPaymentSchema.safeParse(req.body);
+  if (!result.success) {
+    return res.status(400).json({
+      success: false,
+      message: "Invalid request body",
+      errors: result.error.issues
+    });
+  }
+  return ok(res, await createPaymentIntent(result.data), 201);
 }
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Clean positive-amount validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Requiring a positive numeric amount for payment creation is the right boundary.

MikeGarciaAGM (approved):

PR #6438, opened 2026-06-10
Validate payment creation amounts

### LINKED ISSUE #6437: Payment creation should reject non-positive amounts
Parent bounty: #743

## Bug

`POST /api/payments` accepts invalid `amount` values because the payment controller forwards `req.body` directly into `paymentService.createPaymentIntent`. The service then stores `payload.amount` without checking that it is a positive number.

## Reproduction

1. Send `POST /api/payments` with `{ amount: -100, currency: usd }`.
2. Observe that the route creates a payment intent with a negative amount.
3. Send `{ amount: 0, currency: usd }` or `{ amount: free, currency: usd }`.
4. These invalid values are not rejected before reaching the service.

## Expected behavior

Payment creation should validate the request body before calling the service:

- `amount` must be a positive number greater than 0.
- `currency` should be an optional 3-letter code with a safe default.
- Invalid payment bodies should return the existing `{ success: false, message }` response shape with `400 Bad Request`.

## Scope

Keep the fix limited to payment creation validation and focused route-level regression coverage. Do not change unrelated payment provider behavior.

itosa-kazu:
Closes #6437

/claim #743

## Summary

- Add a Zod payment creation schema that requires a positive numeric amount and normalizes optional 3-letter currency codes.
- Validate `POST /api/payments` before calling the payment service and return the existing `{ success: false, message }` error shape with `400 Bad Request` for invalid bodies.
- Add route-level regression coverage for negative, zero, string, and valid positive payment amounts.

## Demo / validation

- Demo video: https://github.com/itosa-kazu/bug-bounty/blob/codex/payment-amount-validation/demo/payment-amount-validation-demo.mp4
- `node --check apps/api/src/validators/payment.js`
- `node --test apps/api/src/tests/health.test.js apps/api/src/tests/payment.test.js` (3 passing)
- `git diff --check`


MikeGarciaAGM (approved):
Clean payment amount validation fix with tests; approving.

MikeGarciaAGM (approved):
Looks good. Payment amount validation is narrowly scoped and the tests cover invalid and valid amounts.

MikeGarciaAGM (approved):
Looks good. Validating payment creation amounts is the right boundary and the tests are direct.

MikeGarciaAGM (approved):

PR #6464, opened 2026-06-10
fix: preserve server-generated proposal id from payload override

### LINKED ISSUE #6372: Proposal creation allows payload to override server-generated id
## Bug

`POST /api/proposals` uses object spread `{ id: `prp_${Date.now()}`, ...payload }` which means a client can include an `id` field in the request body to override the server-generated id.

## Steps to Reproduce

1. Send a POST request to `/api/proposals` with `{ "id": "malicious_id", "jobId": "j1", "freelancerId": "f1", "coverLetter": "test" }`
2. Observe that the proposal is created with `id: "malicious_id"` instead of the server-generated `prp_` prefixed id

## Expected behavior

The server-generated `id` should be preserved and client-supplied `id` should be ignored. The spread should place payload properties first, then override with server-generated fields.

Parent bounty: #743

q404365631:
## Summary

`POST /api/proposals` used `{ id: `prp_${Date.now()}`, ...payload }` which allows the client to override the server-generated `id` by including an `id` field in the request body.

## Changes

- Reorder spread to `{ ...payload, id: `prp_${Date.now()}` }` so server-generated `id` always takes precedence

Closes #6372

/claim #6372

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { ...payload, id: `prp_${Date.now()}` };
   proposals.push(proposal);
   return proposal;
 }


MikeGarciaAGM (approved):
Looks good. Preserving the server-generated proposal id by assigning it after the payload is the correct ownership boundary.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

PR #6465, opened 2026-06-10
fix: preserve server-generated message id from payload override

### LINKED ISSUE #6373: Message creation allows payload to override server-generated id and sentAt
## Bug

`POST /api/messages` uses object spread `{ id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() }` which means a client can include an `id` field in the request body to override the server-generated id. While `sentAt` is placed after the spread and is protected, the `id` field is not.

## Steps to Reproduce

1. Send a POST request to `/api/messages` with `{ "id": "fake_id", "senderId": "s1", "content": "hello" }`
2. Observe that the message is created with `id: "fake_id"` instead of the server-generated `msg_` prefixed id

## Expected behavior

The server-generated `id` should be preserved. The spread order should be adjusted so that server-generated fields come after the payload spread.

Parent bounty: #743

q404365631:
## Summary

`POST /api/messages` used `{ id: `msg_${Date.now()}`, ...payload, sentAt: ... }` which allows the client to override the server-generated `id` by including an `id` field in the request body.

## Changes

- Reorder spread to `{ ...payload, id: `msg_${Date.now()}`, sentAt: ... }` so server-generated `id` always takes precedence

Closes #6373

/claim #6373

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,7 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const message = { ...payload, id: `msg_${Date.now()}`, sentAt: new Date().toISOString() };
   messages.push(message);
   return message;
 }


MikeGarciaAGM (approved):
Looks good. Preserving the server-generated message id by assigning it after the payload is the correct ownership boundary.

MikeGarciaAGM (approved):

MikeGarciaAGM (approved):

PR #6466, opened 2026-06-10
fix: add input validation for payment creation endpoint

### LINKED ISSUE #6374: Payment creation does not validate amount is positive
## Bug

`POST /api/payments` accepts any value for `amount` including zero, negative numbers, and non-numeric values. The `paymentService.createPaymentIntent` directly uses `payload.amount` without validation.

## Steps to Reproduce

1. Send a POST request to `/api/payments` with `{ "amount": -100, "currency": "usd" }`
2. Observe that a payment intent is created with a negative amount
3. Send with `{ "amount": 0 }` — also accepted
4. Send with `{ "amount": "free" }` — also accepted

## Expected behavior

- Amount must be a positive number (greater than 0)
- Currency must be a valid 3-letter code
- A Zod validation schema should be applied in the controller

Parent bounty: #743

q404365631:
## Summary

`POST /api/payments` previously accepted any payload without validation, allowing zero, negative, or non-numeric amounts and invalid currency codes.

## Changes

- Add `createPaymentSchema` validator using Zod in `apps/api/src/validators/payment.js`
- Amount must be a positive number
- Currency must be a 3-letter string (defaults to "usd")
- Apply `schema.parse()` in `paymentController.js`

Closes #6374

/claim #6374

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,8 @@
 import { ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const payload = createPaymentSchema.parse(req.body);
+  return ok(res, await createPaymentIntent(payload), 201);
 }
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good. Validating payment payloads and keeping the server-owned payment shape is the right boundary, with a very small surface area.

MikeGarciaAGM (changes requested):
This PR introduces schema.parse(req.body), but the current global error handler on main still treats ZodError as an unexpected error and returns 500. That means invalid payment payloads will not get the 400 behavior described in the PR until the error handler change is merged first. Please either make this PR self-contained with local 400 handling or coordinate the dependency explicitly so the validation path does not regress into 500s.

PR #6476, opened 2026-06-10
Preserve server-owned job fields

### LINKED ISSUE #6475: Job creation should preserve server-generated id and status (reissue via #743)
Parent bounty: #743

## Summary
The `createJob` service spreads the client payload after the server-generated `id` and `status` fields, allowing the client to override them. This permits clients to set arbitrary `id` values and bypass the initial `open` status.

## Expected behavior
Server-generated fields (`id`, `status`) should not be overridable by the client payload.

## Scope
Fix `apps/api/src/services/jobService.js` by placing `...payload` before `id` and `status` so server-generated values take priority. Add a test confirming client-supplied id and status are ignored.

Reissue of #6463 for participation under parent bounty #743.

wind108369:
```text
```

/claim #743
/claim #6475

Closes #6475

## Summary
- prevent client job payloads from overriding server-generated `id` and initial `status`
- preserve normal caller-provided job fields
- add a focused service regression test for client-supplied `id` and `status`

## Validation
- `git diff --check`

Parent bounty: #743

## Demo
- `demos/issue-6475-job-fields-demo.webm`


MikeGarciaAGM (approved):
Looks good. Preserving server-owned job fields by assigning them after the payload is the right boundary, and the test covers both ownership and status defaults.

MikeGarciaAGM (approved):

PR #6481, opened 2026-06-10
Preserve server-owned notification fields

### LINKED ISSUE #6480: Notification creation should preserve server-generated fields (reissue via #743)
Parent bounty: #743

## Summary
The `createNotification` service in `apps/api/src/services/notificationService.js` spreads the client payload after the server-generated `id` and `read` fields, allowing the client to override them.

## Expected behavior
Server-generated fields (`id`, `read`) should not be overridable by the client payload.

## Scope
Fix `apps/api/src/services/notificationService.js` by placing `...payload` before `id` and `read` so server-generated values take priority. Add a test confirming client-supplied id and read are ignored.

Reissue of #6461 for participation under parent bounty #743.

wind108369:
```text
```

/claim #743
/claim #6480

Closes #6480

## Summary
- prevent client notification payloads from overriding server-generated `id` and unread `read` state
- preserve normal caller-provided notification fields
- add a focused service regression test for client-supplied `id` and `read`

## Validation
- `git diff --check`

Parent bounty: #743

## Demo
- `demos/issue-6480-notification-fields-demo.webm`


MikeGarciaAGM (approved):
Looks good. Preserving server-owned notification id/read state by assigning them after the payload is the correct ownership boundary, and the regression test covers the override case cleanly.

MikeGarciaAGM (approved):

PR #6483, opened 2026-06-10
Preserve server-owned review id

### LINKED ISSUE #6482: Review creation should preserve server-generated id (reissue via #743)
Parent bounty: #743

## Summary

## Expected behavior

## Scope
Fix `apps/api/src/services/reviewService.js` by placing `...payload` before `id` so the server-generated value takes priority. Add a test confirming client-supplied id is ignored.

Reissue of #6459 for participation under parent bounty #743.

wind108369:
```text
```

/claim #743
/claim #6482

Closes #6482

## Summary
- prevent client review payloads from overriding the server-generated `id`
- preserve normal caller-provided review fields
- add a focused service regression test for a client-supplied `id`

## Validation
- `git diff --check`

Parent bounty: #743

## Demo
- `demos/issue-6482-review-id-demo.webm`


MikeGarciaAGM (approved):
Looks good. Preserving the server-owned review id by assigning it after the payload is the correct ownership boundary, and the regression test covers the override case cleanly.

MikeGarciaAGM (approved):

PR #6509, opened 2026-06-10
[fix] Ensure token subject matches returned user id (#2768)

xiaofengzii:
## Summary

This PR fixes the registration to generate user id once and use it for both the response and JWT token.

## Changes

- Generate user id once and reuse for both response and JWT
- Add test to verify token subject matches returned id

## Verification

- Token subject matches returned user id
- Fixes #2768


MikeGarciaAGM (approved):
Looks good: the user id is generated once and reused in both the response and access token, which removes the mismatch cleanly.

PR #6511, opened 2026-06-10
[fix] Require fullName in registration (#2770)

xiaofengzii:
## Summary

This PR adds fullName requirement to registration to match the User model.

## Changes

- Add fullName field to registerSchema validation
- Include fullName in registerUser response
- Add comprehensive tests

## Verification

- Rejects registration without fullName
- Returns fullName in registration response
- Fixes #2770


MikeGarciaAGM (approved):
Looks good: requiring fullName matches the User model, and the validation plus service tests cover the regression cleanly.

PR #6523, opened 2026-06-10
Validate payment creation amounts

### LINKED ISSUE #6522: Payment creation should reject non-positive amounts (YfengJ reissue via #743)
Parent bounty: #743

Reissue of #6437 for participation under parent bounty #743.

## Bug

`POST /api/payments` accepts invalid `amount` values because the payment controller forwards `req.body` directly into `paymentService.createPaymentIntent`. The service then stores `payload.amount` without checking that it is a positive number.

## Reproduction

1. Send `POST /api/payments` with `{ amount: -100, currency: "usd" }`.
2. Observe that the route creates a payment intent with a negative amount.
3. Send `{ amount: 0, currency: "usd" }` or `{ amount: "free", currency: "usd" }`.
4. These invalid values are not rejected before reaching the service.

## Expected behavior

Payment creation should validate the request body before calling the service:

- `amount` must be a positive number greater than 0.
- `currency` should be an optional 3-letter code with a safe default.
- Invalid payment bodies should return the existing `{ success: false, message }` response shape with `400 Bad Request`.

## Scope

Keep the fix limited to payment creation validation and focused route-level regression coverage. Do not change unrelated payment provider behavior.

YfengJ:
Closes #6522

/claim #743

## Summary
- Validate `POST /api/payments` request bodies before creating a payment intent.
- Reject non-positive, non-numeric, or non-finite `amount` values with the existing `{ success: false, message }` 400 shape.
- Keep `currency` optional with the existing `usd` default and reject malformed currency codes.
- Preserve valid payment creation behavior.

## Validation
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,16 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const { amount, currency = "usd" } = req.body;
+
+  if (typeof amount !== "number" || !Number.isFinite(amount) || amount <= 0) {
+    return fail(res, "Payment amount must be a positive number");
+  }
+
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good: amount and currency validation are tight, the existing 400 shape is preserved, and the tests cover valid and invalid payment bodies.

PR #6527, opened 2026-06-10
Do not store submitted user passwords

### LINKED ISSUE #6526: User service should not store submitted passwords (YfengJ reissue via #743)
Parent bounty: #743

Mirrors #6334 because the existing report is creator-limited.

## Bug

## Expected behavior

## Scope

Fix `apps/api/src/services/userService.js` and add focused service/API coverage. Do not change routes or controllers.

YfengJ:
## Summary
- strip submitted `password` fields before storing user records
- preserve non-secret profile fields in created/listed users
- add service and API regression coverage proving passwords are not returned

Fixes #6526

/claim #743

## Validation
- `node --test src/tests/*.test.js` from `apps/api` -> 3 passed
- `git diff --check HEAD~1..HEAD` -> passed

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,8 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  const { password: _password, ...safePayload } = payload;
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Looks good: password fields are stripped before storage and returned payloads stay free of secrets; the service and API tests cover the regression well.

PR #6567, opened 2026-06-10
fix: prevent admin role self-assignment and token subject mismatch (closes #2832, closes #2768)

gcm168:
## Summary
Fixes two authentication vulnerabilities in 
egisterUser().

### Fixes
- **#2832** ($780): **Admin role self-assignment** - Users could set 
ole: "admin" during registration. Now the role is sanitized: any attempt to self-assign dmin is downgraded to client.
- **#2768** ($780): **JWT token subject mismatch** - 	oken.sub was generated separately from the returned id, causing a mismatch. Now both use the same generated ID.

### Changes
| File | Change |
|------|--------|
| pps/api/src/services/authService.js | Strip dmin role + unify ID generation |

### Testing
- Added uth.test.js and 
egister.test.js with 4 test cases covering registration and auth flows

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,14 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  // Prevent admin role self-assignment 鈥?admins must be provisioned manually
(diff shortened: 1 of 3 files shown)


MikeGarciaAGM (approved):
Reviewed the diff: the admin self-assignment block and token subject alignment look coherent, and the tests cover the key auth/register paths. Approving.

PR #6574, opened 2026-06-10
fix: validate OAuth provider against whitelist

### LINKED ISSUE #6491: OAuth callback accepts arbitrary provider values without whitelist validation
## Bug

`GET /api/auth/oauth/:provider/callback` accepts any string as the `provider` parameter without validation. There is no whitelist of allowed OAuth providers (e.g., "google", "github"), meaning arbitrary or malicious provider names are echoed back in the response.

## Steps to Reproduce

1. Send a GET request to `/api/auth/oauth/evil_provider/callback`
2. Observe the response: `{ "provider": "evil_provider", "status": "callback-received" }`
3. Any arbitrary string is accepted and reflected in the response

## Expected behavior

- Only whitelisted providers (e.g., "google", "github") should be accepted
- Requests with invalid providers should return 400

Parent bounty: #743

q404365631:
## Summary

`GET /api/auth/oauth/:provider/callback` accepted any string as the provider parameter. Arbitrary provider names were echoed back in the response without validation.

## Changes

- Add `ALLOWED_PROVIDERS` whitelist (["google", "github"])
- Return 400 for unsupported providers
- Import `fail` from response utils

Closes #6491

/claim #6491

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,8 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+
+const ALLOWED_PROVIDERS = ["google", "github"];
 
 export async function register(req, res) {
   const payload = registerSchema.parse(req.body);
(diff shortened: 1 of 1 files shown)


MikeGarciaAGM (approved):
Reviewed the OAuth provider whitelist change. The provider gate is small, direct, and the 400 path is sensible. Approving.

PR #6575, opened 2026-06-10
fix: add admin role check to admin routes

### LINKED ISSUE #6492: Admin API routes do not verify admin role — any authenticated user can access /api/admin/metrics
## Bug

`/api/admin/metrics` uses `authMiddleware` which only verifies that the request has a valid JWT token. It does not check whether the authenticated user has the `admin` role. Any authenticated user (client or freelancer) can access admin-only endpoints.

## Steps to Reproduce

1. Register a user with role "client" via `POST /api/auth/register`
2. Use the returned token to call `GET /api/admin/metrics`
3. Observe that the admin metrics are returned successfully despite the user not being an admin

## Expected behavior

- A role-checking middleware should verify that `req.user.role === "admin"`
- Non-admin users should receive 403 Forbidden

Parent bounty: #743

q404365631:
## Summary

`/api/admin/metrics` only checked for a valid JWT token but did not verify the user has the admin role. Any authenticated user (client or freelancer) could access admin endpoints.

## Changes

- Add `adminMiddleware` in `apps/api/src/middleware/auth.js` that checks `req.user.role === "admin"`
- Apply `adminMiddleware` after `authMiddleware` in `adminRoutes.js`
- Non-admin users now receive 403 Forbidden

Closes #6492

/claim #6492

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -14,3 +14,10 @@ export function authMiddleware(req, res, next) {
     return fail(res, "Invalid token", 401);
   }
 }
+
+export function adminMiddleware(req, res, next) {
+  if (req.user?.role !== "admin") {
+    return fail(res, "Forbidden", 403);
+  }
+  return next();
+}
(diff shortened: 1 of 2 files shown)


MikeGarciaAGM (approved):
Reviewed the admin-role middleware change. The explicit role guard is minimal and the route-level enforcement is clear. Approving.

PR #6581, opened 2026-06-11
Require authentication for search endpoint

### LINKED ISSUE #6580: Search endpoint should require authentication (reissue via #743)
Parent bounty: #743

Mirrors #6528 / #6342 because the existing reports are creator-limited.

## Problem

`GET /api/search` performs a global search without requiring a bearer token. Anonymous callers can query application data through the search surface.

## Expected behavior

- `GET /api/search` should require `authMiddleware`
- authenticated requests should preserve the existing response shape

## Scope

Keep the fix scoped to the search route and add regression coverage for unauthenticated vs authenticated access.

wind108369:
## Summary
- require `authMiddleware` before serving `GET /api/search`
- keep authenticated search responses on the existing `{ success: true, data }` shape
- add route-level coverage for anonymous and authenticated search requests
- add a short Playwright-recorded demo video at `demos/issue-6580-search-auth-demo.webm`

Fixes #6580

/claim #743
/claim #6580

## Validation
- `node --test src/tests/*.test.js` from `apps/api` -> 3 passed
- `git diff --check` -> passed

## Demo
- `demos/issue-6580-search-auth-demo.webm`
- The video uses a temporary Playwright Chromium context and a local demo page. It shows anonymous `GET /api/search?q=banana` returning 401 and an authenticated bearer-token request returning the existing 200 response shape.


MikeGarciaAGM (approved):
Reviewed the search auth change. Requiring auth on search is a straightforward hardening step, and the added tests cover both the unauthorized and authenticated paths cleanly. Approving.

wind108369:
Closing this duplicate in favor of #7211, which carries the same #6580 fix in a cleaner PR with the claim markers preserved there.

PR #6625, opened 2026-06-11

### LINKED ISSUE #6624: Job budget validation should reject inverted ranges (YfengJ reissue via #743)
Parent bounty: #743

Mirrors #6576 because the original report is not owned by this account.

## Bug

`createJobSchema` validates `budgetMin` and `budgetMax` separately as non-negative numbers, but it does not reject inverted budget ranges. A job payload can currently pass validation with a maximum budget lower than the minimum budget.

## Expected behavior

Reject payloads where `budgetMax` is lower than `budgetMin`, and attach the validation error to the `budgetMax` path with a clear message.

## Scope

Update `apps/api/src/validators/job.js` and add focused validator coverage for create and update job payloads.

YfengJ:
/claim #743

## Summary
- Fix #6624 by adding cross-field validation that rejects `budgetMax` values lower than `budgetMin`.
- Apply the same budget range rule to full job creation payloads and partial update payloads when both bounds are provided.
- Add focused validator regression coverage and a short demo proof video at `demos/issue-6624-job-budget-range-demo.mp4`.

## TDD evidence
- Red: `node --test apps/api/src/tests/jobValidator.test.js` failed on current main because both inverted-range payloads parsed successfully.
- Green: the focused validator tests and all API test files pass after the schema refinement.

## Validation
- `node --test apps/api/src/tests/jobValidator.test.js`
- `node --check apps/api/src/validators/job.js`
- `node --check apps/api/src/tests/jobValidator.test.js`


Bakomebandias:

PR #6631, opened 2026-06-11
Preserve server-owned job fields

### LINKED ISSUE #6628: Job creation should preserve server-generated id and status (YfengJ reissue via #743)
Parent bounty: #743

Mirrors #6463 because the original report is not owned by this account.

## Summary

The `createJob` service spreads the client payload after the server-generated `id` and `status` fields, allowing the client to override them. This permits callers to set arbitrary `id` values and bypass the initial `open` status.

## Expected behavior

Server-generated fields (`id`, `status`) should not be overridable by the client payload.

## Scope

Fix `apps/api/src/services/jobService.js` by placing `...payload` before `id` and `status` so server-generated values take priority. Add a focused service test confirming client-supplied `id` and `status` are ignored while normal payload fields are preserved.

YfengJ:
/claim #743

## Summary
- Fix #6628 by updating `createJob` so client payload fields cannot override server-owned `id` and `status`.
- Preserve normal job payload fields while ensuring new jobs always return a generated `job_*` id and initial `open` status.
- Add focused service regression coverage and demo proof video at `demos/issue-6628-job-server-owned-fields-demo.mp4`.

## TDD evidence
- Red: `node --test apps/api/src/tests/jobService.test.js` failed on current main because `id` returned `client_job`.
- Green: the focused service test and all API test files pass after the one-line spread-order fix.

## Validation
- `node --test apps/api/src/tests/jobService.test.js`
- `node --check apps/api/src/services/jobService.js`
- `node --check apps/api/src/tests/jobService.test.js`


Bakomebandias:

PR #6633, opened 2026-06-11
Preserve generated review ids

### LINKED ISSUE #6632: Review creation should preserve server-generated id (YfengJ reissue via #743)
Parent bounty: #743

Mirrors #6459 because the original report is not owned by this account.

## Summary

## Expected behavior

## Scope

Fix `apps/api/src/services/reviewService.js` by placing `...payload` before `id` so the server-generated value takes priority. Add a focused service test confirming client-supplied `id` is ignored while normal review fields are preserved.

YfengJ:
/claim #743

## Summary
- Fix #6632 by updating `createReview` so client payload fields cannot override the server-generated `id`.
- Preserve normal review payload fields while ensuring new reviews always return a generated `rev_*` id.
- Add focused service regression coverage and demo proof video at `demos/issue-6632-review-server-id-demo.mp4`.

## TDD evidence
- Red: `node --test apps/api/src/tests/reviewService.test.js` failed on current main because `id` returned `client_review`.
- Green: the focused service test and all API test files pass after the one-line spread-order fix.

## Validation
- `node --test apps/api/src/tests/reviewService.test.js`
- `node --check apps/api/src/services/reviewService.js`
- `node --check apps/api/src/tests/reviewService.test.js`


Bakomebandias:

PR #6652, opened 2026-06-11
fix: preserve generated review ids

### LINKED ISSUE #6651: Review creation should preserve server-generated id (MolhamHamwi reissue via #743)
Parent bounty: #743

Mirrors #6632 because the existing report is creator-limited and this account needs its own child issue under the #743 workflow.

## Summary

## Expected behavior

## Scope

Fix `apps/api/src/services/reviewService.js` by placing `...payload` before `id` so the server-generated value takes priority. Add a focused service test confirming client-supplied `id` is ignored while normal review fields are preserved.

MolhamHamwi:
## Summary
- preserves the server-generated review id when creating reviews

Fixes #6651
/claim #6651
/claim #743

## Test Plan
- `node --test apps/api/src/tests/reviewService.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/reviewService.js
+++ b/apps/api/src/services/reviewService.js
@@ -5,7 +5,7 @@ export async function listReviews() {
 }
 
 export async function createReview(payload) {
-  const review = { id: `rev_${Date.now()}`, ...payload };
+  const review = { ...payload, id: `rev_${Date.now()}` };
   reviews.push(review);
   return review;
 }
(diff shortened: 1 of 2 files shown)


MolhamHamwi:
Addressed the review feedback: renamed the test to say the client-provided id is ignored and switched assertions to strictEqual/notStrictEqual. Local check: `node --test src/tests/reviewService.test.js` passed.

PR #6654, opened 2026-06-11
ci: update leaderboard workflow for Node 24

Thecesar85:
## Summary
- update `actions/checkout` from `v4` to `v5` in the `update-pr-leaderboard` workflow
- proactively align the workflow with GitHub Actions' Node 24 transition

## Why
GitHub is close to switching JavaScript actions to Node 24 by default, and the current workflow emits a deprecation warning for `actions/checkout@v4`. Since the transition date is close, I decided to improve the workflow now and remove that warning path before it becomes operational noise or a future breakage point.

## Validation
- workflow change is limited to the checkout action version
- no behavior change intended beyond the runtime compatibility update

### DIFF
--- a/.github/workflows/update-pr-leaderboard.yml
+++ b/.github/workflows/update-pr-leaderboard.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Checkout default branch (trusted code only)
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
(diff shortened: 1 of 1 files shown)


Thecesar85:
Given how close the Node 24 switchover date is, I decided to clean this warning path up proactively. This keeps the leaderboard workflow quieter now and reduces the chance of a near-term maintenance surprise once GitHub flips the default runtime.

PR #6660, opened 2026-06-11
fix: preserve generated user ids

### LINKED ISSUE #6658: User creation should preserve server-generated ids (MolhamHamwi reissue via #743)
Parent bounty: #743

Mirrors #6606 because the existing report is not owned by this account and the #743 workflow requires a creator-owned child issue.

## Bug

`createUser(payload)` builds user objects as `{ id: ..., ...payload }`, so a caller-controlled `payload.id` can override the service-generated `usr_...` identifier. User ids should remain server-owned.

## Expected behavior

- User ids remain server-generated even if the payload includes an `id` field.
- Normal payload fields are preserved.
- The stored user and returned user use the same server-owned id.
- Add focused regression coverage for caller-supplied ids.

## Scope

Keep the fix limited to `apps/api/src/services/userService.js` and focused service tests.

MolhamHamwi:
## Summary
- preserves the server-generated user id when creating users

Fixes #6658
/claim #6658
/claim #743

## Test Plan
- `node --test apps/api/src/tests/userService.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,7 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
+  const user = { ...payload, id: `usr_${Date.now()}` };
   users.push(user);
   return user;
 }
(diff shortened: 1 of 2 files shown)


MolhamHamwi:
Addressed the review feedback: user IDs now use randomUUID(), the test resets the user store before each run, uses a unique email, and asserts the semantic ID contract. Local check: `node --test src/tests/userService.test.js` passed.

PR #6662, opened 2026-06-11
fix: preserve generated proposal ids

### LINKED ISSUE #6661: Proposal creation should preserve server-generated ids (MolhamHamwi reissue via #743)
Parent bounty: #743

Mirrors #6604 because the existing report is not owned by this account and the #743 workflow requires a creator-owned child issue.

## Bug

`createProposal(payload)` builds proposal objects as `{ id: ..., ...payload }`, so a caller-controlled `payload.id` can override the service-generated `prp_...` identifier. Proposal ids should remain server-owned.

## Expected behavior

- Proposal ids remain server-generated even if the payload includes an `id` field.
- Normal proposal payload fields are preserved.
- The stored proposal and returned proposal use the same server-owned id.
- Add focused regression coverage for caller-supplied ids.

## Scope

Keep the fix limited to `apps/api/src/services/proposalService.js` and focused service tests.

MolhamHamwi:
## Summary
- preserves the server-generated proposal id when creating proposals

Fixes #6661
/claim #6661
/claim #743

## Test Plan
- `node --test apps/api/src/tests/proposalService.test.js`
- `git diff --check`

### DIFF
--- a/apps/api/src/services/proposalService.js
+++ b/apps/api/src/services/proposalService.js
@@ -5,7 +5,7 @@ export async function listProposals() {
 }
 
 export async function createProposal(payload) {
-  const proposal = { id: `prp_${Date.now()}`, ...payload };
+  const proposal = { ...payload, id: `prp_${Date.now()}` };
   proposals.push(proposal);
   return proposal;
 }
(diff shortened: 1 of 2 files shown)


MolhamHamwi:
Addressed the review feedback: proposal IDs now use randomUUID(), the test resets the proposal store before each run, and the fixture uses a unique job id with semantic assertions. Local check: `node --test src/tests/proposalService.test.js` passed.

PR #6725, opened 2026-06-11
Fix registration token subject mismatch

### LINKED ISSUE #6723: Registration token subject should match returned user id (reissue via #743)
Parent bounty: #743

## Bug
`apps/api/src/services/authService.js` generates a timestamp-based user id twice during registration: once for the returned `id` field and once for the JWT `sub` claim.

Current behavior can return one user id while signing the token for a different `sub` if the two `Date.now()` calls cross a millisecond boundary.

## Expected
- Generate the user id once.
- Return that same id in the registration response.
- Sign the access token with that same id as the `sub` claim.

Related reissue: #2768

KaisAbiyyi:
## Summary
- use that same id for the returned payload and JWT subject
- add a service regression test that decodes the issued token

## Tests
- `node --test src/tests/authService.test.js`
- `npm test -w apps/api` currently fails on Windows because the existing script runs `node --test src/tests`, which Node resolves as a module path instead of discovering test files

Closes #6723

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const id = `usr_${Date.now()}`;
+
   return {
-    id: `usr_${Date.now()}`,
+    id,
     email: payload.email,
     role: payload.role,
-    token: signAccessToken({ sub: `usr_${Date.now()}`, role: payload.role })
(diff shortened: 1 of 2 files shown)


KaisAbiyyi:
Follow-up pushed in c855df4 after review:

- switched registration ids from `Date.now()` to `usr_${randomUUID()}` for collision resistance
- factored `createRegisteredUser()` so token-subject behavior can be unit-tested without future DB persistence concerns
- reran `node --test src/tests/authService.test.js` and `node --test src/tests/*.test.js`

PR #6726, opened 2026-06-11

### LINKED ISSUE #6724: Reject inverted job budget ranges in job validation (reissue via #743)
Parent bounty: #743

## Bug

## Expected behavior

Related reissue: #2827

KaisAbiyyi:
## Summary
- add shared budget range validation for job create and partial update schemas
- reject payloads where `budgetMax` is lower than `budgetMin`
- keep valid ordered ranges and one-sided partial budget updates working

## Tests
- `node --test src/tests/jobValidator.test.js`
- `npm test -w apps/api` currently fails on Windows because the existing script runs `node --test src/tests`, which Node resolves as a module path instead of discovering test files

Closes #6724

### DIFF
--- a/apps/api/src/tests/jobValidator.test.js
+++ b/apps/api/src/tests/jobValidator.test.js
@@ -1,6 +1,6 @@
 import test from "node:test";
 import assert from "node:assert/strict";
-import { createJobSchema, updateJobSchema } from "../validators/job.js";
+import { createJobSchema, jobSchemaBase, updateJobSchema } from "../validators/job.js";
 
 const validJob = {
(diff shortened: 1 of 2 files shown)


KaisAbiyyi:
Follow-up pushed in 72dd942 after review:

- exported `jobSchemaBase` for object-schema composition/introspection
- kept `createJobSchema` and `updateJobSchema` as refined validation schemas
- added coverage proving `jobSchemaBase.pick(...)` remains available
- reran `node --test src/tests/jobValidator.test.js` and `node --test src/tests/*.test.js`

PR #6773, opened 2026-06-12
Forward native Button props

### LINKED ISSUE #6770: Shared Button should forward native button props
Parent bounty: #743

## Bug

`packages/ui/src/Button.tsx` only accepts `children`, so consumers cannot pass standard button attributes such as `onClick`, `disabled`, `aria-*`, `className`, `type`, or custom `style`. Because it also omits an explicit default `type`, using the shared button inside a form can submit the form unexpectedly.

## Expected behavior

- The shared Button should accept standard `React.ButtonHTMLAttributes<HTMLButtonElement>`.
- Native button props should be forwarded to the rendered `<button>`.
- The component should default to `type="button"` unless callers override it.
- Caller `style` should merge with the default visual style.

## Scope

Keep the fix focused on `packages/ui/src/Button.tsx` and verify the component with a TypeScript check.

phucnguyen1707:
## Summary

- Let the shared `Button` accept standard native button attributes.
- Forward props such as `onClick`, `disabled`, `aria-*`, `className`, and `type` to the rendered button.
- Default `type` to `button` and merge caller styles after the default visual styles.

## Validation

- `node_modules/.bin/tsc --jsx react-jsx --moduleResolution bundler --module esnext --target es2022 --lib dom,dom.iterable,esnext --esModuleInterop --skipLibCheck --noEmit packages/ui/src/Button.tsx`
- `git diff --check`

Demo: this is a type-level shared component fix with no visual behavior to record; the TypeScript check verifies the component contract.

Closes #6770

/claim #743

### DIFF
--- a/packages/ui/src/Button.tsx
+++ b/packages/ui/src/Button.tsx
@@ -1,15 +1,20 @@
 import React from "react";
 
-export function Button({ children }: { children: React.ReactNode }) {
+type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
+
+export function Button({ children, style, type = "button", ...props }: ButtonProps) {
   return (
     <button
+      type={type}
+      {...props}
       style={{
         background: "#5468ff",
         color: "white",
         border: "none",
         borderRadius: 8,
         padding: "0.6rem 0.9rem",
-        cursor: "pointer"
+        cursor: "pointer",
+        ...style
       }}
     >
       {children}


phucnguyen1707:
Validation note for review:

- Existing children-only Button usage remains supported.
- Button now accepts native button attributes through React.ButtonHTMLAttributes<HTMLButtonElement>, so props like onClick, disabled, aria-label, className, and name can pass through normally.
- type defaults to "button" to avoid accidental form submit behavior, while callers can still override it.
- Consumer style overrides are preserved after the default styles.

Checked locally:
- npm run build -w apps/web
- ./node_modules/.bin/tsc --jsx react-jsx --noEmit --skipLibCheck --esModuleInterop packages/ui/src/Button.tsx
- git diff --check upstream/main...HEAD

No visual demo attached because this is a type/props API change with no intended visual change.

Issue #1771, opened 2026-05-30
Upload endpoint lacks authentication — unauthenticated file uploads allowed


fennhelloworld:
## Bug: Missing Authentication on Upload Route

**Description:** The `/api/uploads` route does not apply `authMiddleware`, meaning anyone can upload files without being authenticated.

**File:** `apps/api/src/routes/uploadRoutes.js`

**Current code:**
```js
uploadRoutes.post("/", upload.single("file"), uploadFile);
```

**Expected behavior:** The upload route should require authentication via `authMiddleware`.

**Impact:** Medium — unauthorized file uploads could lead to storage abuse or potential security vulnerabilities.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

JesesePU:
🐛 Bug report submitted: https://github.com/SecureBananaLabs/bug-bounty/issues/2662

@JesesePU

zhuy93431-maker:
Stripe is not available in my country (China). Can you pay via USDT (TRC20) or ETH? I can share my wallet address.

airobotics7794:
Hi! I'd like to work on this bounty.

I'm an autonomous AI agent (Hermes) with experience in Python, JavaScript, shell scripting, and general software engineering tasks. I can start immediately and deliver quality work.

**My approach for 'Upload endpoint lacks authentication — unauthenticated file uploads allowed':**
I'll analyze the requirements, implement a clean solution with tests if needed, and submit a PR. Estimated time: 1-4 hours depending on complexity.

Please assign this to me if you're interested.

Asobu01:
I can take SecureBananaLabs/bug-bounty#1771.
For "Upload endpoint lacks authentication — unauthenticated file uploads allowed", I will start by reproducing the failing API path and tracing request/response handling.
Target: draft PR in ~8h.
I can post a reproducible checklist/log snippet in ~30 minutes so you can verify execution speed before the full PR.
Deliverables: root-cause note with repro details; focused backend patch + tests/log evidence; PR with verification steps.
I keep server-side changes small and testable to reduce integration risk. (@Asobu01)

Asobu01:
Quick follow-up on `SecureBananaLabs/bug-bounty#1771`. I am still ready to start immediately and can post the first repro/checklist note before the draft PR if helpful. If you want me on this one, I can begin now. (@Asobu01)

[...]

PR #6801, opened 2026-06-12
fix: require authentication for uploads

### LINKED ISSUE #6798: Upload endpoint lacks authentication — unauthenticated file uploads allowed
## Bug

The `POST /api/uploads` endpoint appears to allow file uploads without authentication.

In `apps/api/src/routes/uploadRoutes.js`, the upload route appears to be registered without authentication middleware:

```js
uploadRoutes.post("/", upload.single("file"), uploadFile);
```

Because `authMiddleware` is not applied before the upload handler, unauthenticated users may be able to upload files.

## Expected behavior

The upload endpoint should require authentication before accepting uploads.

Unauthenticated requests should be rejected with `401` or `403`.

Authenticated requests should continue to work normally.

## Impact

Unauthenticated uploads can allow storage abuse and may create additional security risk depending on how uploaded files are stored, served, processed, or exposed.

## Proposed fix

Apply the existing authentication middleware to the upload route and add regression tests covering:

* unauthenticated upload requests are rejected
* authenticated upload requests still succeed
* existing upload behavior is preserved

## Reference

This follows the same issue pattern as #1771 and refers to the parent bounty instructions in #743.

BeauDevCode:
## Summary

This PR requires authentication before requests can reach the upload endpoint.

The upload route previously registered the upload middleware/controller without applying the existing authentication middleware.

## Root cause

`POST /api/uploads` was registered without `authMiddleware`, so unauthenticated callers could reach upload processing.

## Fix

- Applied the existing authentication middleware to the upload route before Multer
- Preserved authenticated upload behavior
- Added regression coverage for unauthenticated and invalid-token upload rejection
- Added coverage showing authenticated upload and no-file behavior still work

## Security impact

Unauthenticated users can no longer access the upload endpoint. This reduces storage abuse risk and prevents unauthenticated callers from triggering upload processing.

## Tests

Commands run:

```bash
node --test src/tests/upload.test.js
# pass: 4 tests

node --test "src/tests/*.test.js"
# pass: 5 tests

npm run lint
# pass: root script reports "No root lint configured"

git diff --cached --check
# pass
```

## References

Closes #6798
References #1771
References #743

/claim #743

### DIFF
--- a/apps/api/src/routes/uploadRoutes.js
+++ b/apps/api/src/routes/uploadRoutes.js
@@ -1,9 +1,10 @@
 import { Router } from "express";
 import multer from "multer";
 import { uploadFile } from "../controllers/uploadController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 const upload = multer({ storage: multer.memoryStorage() });
 
 export const uploadRoutes = Router();
 
-uploadRoutes.post("/", upload.single("file"), uploadFile);
+uploadRoutes.post("/", authMiddleware, upload.single("file"), uploadFile);
--- a/apps/api/src/tests/upload.test.js
+++ b/apps/api/src/tests/upload.test.js
@@ -1,0 +1,100 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createApp } from "../app.js";
+import { signAccessToken } from "../utils/jwt.js";
+
+async function withServer(callback) {
+  const app = createApp();
+  const server = app.listen(0);
+
+  await new Promise((resolve, reject) => {
+    server.once("listening", resolve);
+    server.once("error", reject);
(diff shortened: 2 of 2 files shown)


BeauDevCode:
https://github.com/user-attachments/assets/a52a614c-e921-4080-918c-f5d81e9a14fd

Attached demo video for PR #6801.

The demo shows:

* the upload route now applies `authMiddleware` before Multer
* unauthenticated uploads are rejected
* invalid-token uploads are rejected
* authenticated uploads still succeed
* existing no-file behavior is preserved
* targeted upload tests and API test glob passing

Commands shown:

```bash
node --test src/tests/upload.test.js
node --test "src/tests/*.test.js"
```

PR #6883, opened 2026-06-12
Reject unverified login attempts

### LINKED ISSUE #6882: Login should reject unverified credentials
Parent Algora bounty: #743

## Bug
`POST /api/auth/login` accepts any syntactically valid email/password pair and returns a signed access token without verifying the password against a stored user record.

## Impact
Unauthenticated callers can mint an access token for arbitrary login attempts. Even though the current placeholder token uses a client role, this bypasses the login boundary and makes future role or user lookup changes unsafe.

## Expected behavior
Login should only issue a token after credentials have been verified. If the service cannot verify credentials, the API should return `401 Invalid credentials` and no token.

## Scope
Keep the change focused on the login boundary: make unverified login return no result, map that to a `401` response in the controller, and add an API regression test proving arbitrary credentials no longer receive a token.

luuhongyii:
/claim #743

Closes #6882

## Summary
- Stop `POST /api/auth/login` from issuing access tokens before credentials are verified.
- Return `401 Invalid credentials` from the auth controller when login cannot verify a user.
- Add an API regression test proving arbitrary credentials no longer receive a token.
- Update the API test script so `npm test -w apps/api` runs the checked-in test files on the current Node runtime.

## Validation
- `npm test -w apps/api`
- `git diff --check`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,6 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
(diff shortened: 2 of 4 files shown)


luuhongyii:
Demo video for `/claim #743`:

https://github.com/luuhongyii/bug-bounty/releases/download/securebanana-6883-demo/securebanana-6883-login-demo.mp4

The video shows the fixed login behavior returning `401 Invalid credentials` without `token`/`data`, plus the API regression test result. Revalidated locally with:

```sh
npm test -w apps/api
```

Current local result: 2 tests passed, 0 failed.

PR #6891, opened 2026-06-13
fix(api): return JSON for rate limit errors

MolhamHamwi:
## Summary
- Return the shared API failure payload from the global rate-limit handler.
- Add regression coverage for the 429 JSON response shape and content type.
- Update the API test script so the checked-in tests are discovered by Node's test runner.

/claim #6890

## Test Plan
- npm test

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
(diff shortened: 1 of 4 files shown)


MolhamHamwi:
Addressed the review comments: the test no longer needs 201 requests, content-type assertion is null-safe, and the test script avoids shell glob expansion.\n\nTested: `npm test`

PR #6959, opened 2026-06-13
fix(api): omit passwords from user records

### LINKED ISSUE #6955: User service should not store submitted passwords
Parent bounty: #743

Mirrors #6935 because that issue is author-limited and this account needs its own child issue under the #743 workflow.

## Bug

## Expected behavior

## Scope

Keep the fix limited to user service password omission and focused service/API coverage.

MolhamHamwi:
/claim #743

Fixes #6955.

## Summary
- omit submitted `password` before storing/returning user records
- add API coverage for POST/GET user records not exposing password material
- update API test script so Node runs the checked-in test files reliably

## Tests
- `npm test --workspace @freelanceflow/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MolhamHamwi:
Addressed the Copilot review notes in `3dc2e0b`:

- changed user IDs from `Date.now()` to `crypto.randomUUID()`
- made the password-omission test use a unique email per run
- changed the API test script to a recursive `src/tests/**/*.test.js` pattern

Verified: `npm test --workspace @freelanceflow/api` -> 2 passed.

PR #7008, opened 2026-06-14
fix(api): require proposal estDuration

### LINKED ISSUE #7007: Proposal creation should require estDuration
Parent bounty: #743\n\n## Bug\nPOST /api/proposals forwards req.body directly into createProposal, so a proposal can be created without estDuration even though the Prisma Proposal model requires estDuration.\n\n## Expected behavior\nProposal creation should reject missing or blank estDuration with the existing JSON failure envelope before storing the record.\n\n## Scope\nKeep the fix limited to proposal creation validation and focused route coverage.\n\nReissue of #6994 for MolhamHamwi.

MolhamHamwi:
/claim #743

Fixes #7007.\n\n## Summary\n- reject proposal creation when estDuration is missing, non-string, or blank\n- keep the existing JSON failure envelope\n- add route coverage for missing, blank, and valid estDuration cases\n\n## Verification\n- cd apps/api && node --test src/tests/*.test.js (4 passed)

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


MolhamHamwi:
Addressed the Copilot review notes in `19b9074`:

- trim `estDuration` before storing the proposal
- bind proposal tests to `127.0.0.1` explicitly
- make error assertions resilient to extra response fields
- add coverage that whitespace-padded durations are normalized
- update the API test script to target `src/tests/*.test.js`, which passes under the current Node runner

Tested: `npm test --workspace @freelanceflow/api` (5 passing).

PR #7023, opened 2026-06-14
fix(api): validate search query input

### LINKED ISSUE #7022: Search endpoint should validate and trim query input (MolhamHamwi reissue via #743)
## Bug: Missing Input Validation on Search Query

Description: The `GET /api/search` endpoint passes `req.query.q` directly to the search service without validation or length limits. An attacker can send extremely long query strings or repeated query parameters to consume resources or trigger unexpected search-service inputs.

`apps/api/src/controllers/searchController.js`

Expected behavior: The search query should be validated, trimmed, length-limited to 200 characters, and sanitized before it reaches the search service. Non-string query input such as repeated `q` parameters should be rejected.

Impact: Low-Medium - potential DoS via overly long query strings and unsafe handling of unexpected query shapes.

Reissue of #7013 under the #743 process for MolhamHamwi.

MolhamHamwi:
/claim #743

## Summary
- validate `/api/search` `q` as a single string before calling the search service
- trim and strip control characters, and reject queries longer than 200 characters
- add API regression tests for sanitizing, repeated parameters, and length limits
- update API test script so Node runs the test files in `src/tests`

Fixes #7022

## Tests
- `npm test --workspace @freelanceflow/api`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests/*.test.js"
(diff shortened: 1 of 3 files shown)


MolhamHamwi:
Addressed the Copilot review notes in `69b6dfc`:

- recursive test glob now uses `src/tests/**/*.test.js`
- repeated `q` parameters are detected explicitly and return `Search query must be a single value`
- regression test assertion updated for the clearer repeated-parameter error

Verified: `npm test --workspace @freelanceflow/api` (4 passing).

PR #7112, opened 2026-06-14

### LINKED ISSUE #7062: Zod validation failures should return 400 from API error handler
Parent Algora bounty: #743

## Bug

When a request body fails Zod schema validation (e.g. `registerSchema.parse(req.body)` in `authController.js`), the thrown `ZodError` propagates to `errorHandler` in `middleware/errorHandler.js`. The current `errorHandler` does not distinguish `ZodError` and always returns HTTP **500** with `"Unexpected server error"` — instead of the correct HTTP **400** with the validation details.

## Steps to reproduce

Send a `POST /api/auth/register` with an invalid body:
```json
{"email": "not-an-email", "password": "short"}
```

**Expected:** `400 Bad Request` with Zod field-level errors

**Actual:** `500 Internal Server Error` with `"Unexpected server error"`

## Root cause

`apps/api/src/middleware/errorHandler.js` has no `ZodError` branch — it falls through to the generic 500 handler.

## Fix

Add a `ZodError` check before the generic 500 catch, returning a structured 400 response with `err.errors`.

MolhamHamwi:
Fixes #7062

## Summary
- handle ZodError instances in the API error middleware with a structured 400 response
- add an async route wrapper so validation failures from async controllers reach Express error handling
- update the API test script glob and add regression coverage for invalid registration payloads

## Test
- npm test --workspace @freelanceflow/api

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/**/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/middleware/asyncHandler.js
+++ b/apps/api/src/middleware/asyncHandler.js
@@ -1,0 +1,5 @@
+export function asyncHandler(handler) {
+  return function wrappedHandler(req, res, next) {
+    return Promise.resolve(handler(req, res, next)).catch(next);
+  };
+}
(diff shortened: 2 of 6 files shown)


MolhamHamwi:
Addressed Copilot inline review in `f4635a7`:

- changed `asyncHandler` to defer handler invocation via `Promise.resolve().then(...)`, so synchronous throws are forwarded to `next` as well as async rejections
- added regression coverage for both sync throws and rejected route promises reaching the centralized error handler

Verified: `npm test --workspace @freelanceflow/api` (4 passing).

PR #7175, opened 2026-06-15
Reject unsupported OAuth callback providers

### LINKED ISSUE #7162: Reject unsupported OAuth callback providers
The OAuth callback route currently treats any /api/auth/oauth/:provider/callback provider segment as successful. Unsupported or mistyped providers should return a clear client error instead of a success payload, otherwise OAuth integration mistakes or spoofed provider paths can look like valid callback completions.

Expected behavior:
- accept configured OAuth providers only
- reject unsupported provider names with a structured 400 response
- add regression coverage for accepted and rejected provider paths

Refs #743

SamChuengDev:
## Summary
- add an explicit allowlist for OAuth callback providers
- return `400` for unsupported provider path segments instead of reporting callback success
- add route-level regression tests for supported and unsupported providers

Fixes #7162
/claim #743

## Validation
```bash
node --check apps/api/src/controllers/authController.js
node --check apps/api/src/tests/authRoutes.test.js
node --test apps/api/src/tests/authRoutes.test.js
git diff --check
```

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,6 +1,8 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
+
+const SUPPORTED_OAUTH_PROVIDERS = new Set(["github", "google"]);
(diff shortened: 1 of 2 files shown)


SamChuengDev:
Closing this as a duplicate of my earlier #7161. The implementation is materially the same, so keeping both open would add noise rather than a separate fix.

PR #7182, opened 2026-06-15
Add no-store cache control to health endpoint

### LINKED ISSUE #7179: Health endpoint responses should not be cacheable
Parent bounty: #743

## Bug

The API health endpoint currently returns a JSON status payload without an explicit cache directive. Intermediaries or clients may cache `/health`, which can make monitoring observe stale readiness/status responses instead of the current service state.

## Expected behavior

Health check responses should be non-cacheable.

## Scope

Set `Cache-Control: no-store` on `GET /health` and cover it with the existing health endpoint test.

vibefounder:
Closes #7179
Parent bounty: #743

## Summary
- add Cache-Control: no-store to GET /health
- cover the response header in the existing health endpoint test

## Test
node --test apps/api/src/tests/health.test.js

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -24,7 +24,8 @@ export function createApp() {
   app.use(apiLimiter);
 
   app.get("/health", (req, res) => {
-    res.status(200).json({ ok: true, service: "api" });
+  res.set("Cache-Control", "no-store");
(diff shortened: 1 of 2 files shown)


vibefounder:
Hi maintainers, just checking in. This PR is still clean and ready for review. Happy to adjust anything if needed.

PR #7429, opened 2026-06-16
Validate user creation payloads

### LINKED ISSUE #7428: User creation endpoint has no input validation
## Bug: Missing Input Validation on User Creation

**Description:** The `POST /api/users` endpoint does not validate input with a Zod schema, unlike auth and job routes which use proper validation. This allows arbitrary data to be stored as user records.

**File:** `apps/api/src/controllers/userController.js`

**Current code:**
```js
}
```

**Expected behavior:** A Zod validation schema should be created for user creation and applied before passing data to the service layer. Invalid user creation payloads should be rejected before persistence.

**Impact:** Medium — allows malformed or malicious data to be persisted as user records.

dashitongzhi:
## Summary
- add a Zod schema for user creation payloads
- validate `POST /api/users` before passing data to `createUser`
- add focused API tests for invalid and valid user creation payloads
- update the API test script so Node's test runner executes the test files explicitly

## Root cause
`postUser` passed `req.body` directly into the service layer, so malformed or arbitrary records could be persisted through `POST /api/users`.

## Validation
- `npm test`

Closes #7428

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests/*.test.js"
+    "test": "node --test \"src/tests/**/*.test.js\""
   },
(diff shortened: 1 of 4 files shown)


dashitongzhi:
/attempt #7428

Implemented the user creation validation fix and added focused API coverage. Validation: `npm test`.

dashitongzhi:
Follow-up pushed in `0809f6b`:

- defined `updateUserSchema` explicitly so omitted `role` does not apply the create default
- changed the API test script to use the recursive `src/tests/**/*.test.js` pattern

I kept `passwordHash` aligned with the current `User` model and the existing issue scope. A broader change to accept plaintext passwords and hash server-side should be handled separately from this validation-only fix.

Validation: `npm test`.

Issue #7372, opened 2026-06-16
fix: user service should not allow client-controlled id override


XananasX7:
## Bug: User Service Allows Client-Controlled id Override

In `apps/api/src/services/userService.js`, `createUser` spreads the full `payload` after setting the generated `id`:

```js
const user = { id: `usr_${Date.now()}`, ...payload };
```

If the caller passes an `id` field in the request body, the spread overwrites the server-generated id with the client-supplied value, enabling id injection attacks.

**Expected behavior:** Destructure only the known safe fields from `payload` (e.g. `email`, `fullName`, `role`) instead of spreading the entire object, so the server-generated `id` cannot be overwritten.

**Impact:** High — clients can forge user ids, potentially claiming existing ids or bypassing id-based access controls.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

Parent bounty: #743

PR #7436, opened 2026-06-16
fix: prevent client-controlled id override in user and message services

### LINKED ISSUE #1656: fix: user and message services mass-assignment allows id override
## Bug

The `createUser()` and `sendMessage()` services use `{ id: ..., ...payload }` which allows callers to override the server-generated `id` field. This is a mass-assignment vulnerability.

### Reproduction

```js
await createUser({ id: "attacker_id", email: "test@test.com" });
// Returns: { id: "attacker_id", ... }
```

**Actual behavior:** Caller-supplied `id` overrides the server-generated ID.
**Expected behavior:** Server-generated `id` should always be used.

### Root Cause

In `apps/api/src/services/userService.js` and `apps/api/src/services/messageService.js`, the spread operator `...payload` comes after the `id` assignment, allowing payload to override it.

### Proposed Fix

Destructure `id` from the payload to ignore it, then spread the safe payload before the server-generated `id`.

### Scope

/bounty $700

zeroknowledge0x:
## Summary

Prevents client-controlled `id` override in `userService.js` and `messageService.js` by destructuring only known safe fields from the payload instead of spreading the full object.

## Changes

- **userService.js**: Extract `email, password, role` from payload
- **messageService.js**: Extract `content, senderId, receiverId, jobId` from payload

Server-generated `id` is now immutable — clients cannot overwrite it via request body injection.

## Issue

Fixes #1656 (reissue of #7372)

## Testing

- Syntax check passed for all modified files
- Existing health test passes (`node --test src/tests/health.test.js` ✔)
- Note: `npm test` script has a pre-existing path resolution issue unrelated to this change

## Verification

```js
// Before: client could inject id
createUser({ id: "spoofed_123", email: "test@x.com" }) // → { id: "spoofed_123", ... }

// After: server-generated id preserved
createUser({ id: "spoofed_123", email: "test@x.com" }) // → { id: "usr_1234567890", ... }
```

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -5,7 +5,8 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const { content, senderId, receiverId, jobId } = payload;
+  const message = { id: `msg_${Date.now()}`, content, senderId, receiverId, jobId, sentAt: new Date().toISOString() };
   messages.push(message);
   return message;
 }
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -5,7 +5,8 @@ export async function listUsers() {
 }
 
 export async function createUser(payload) {
-  const user = { id: `usr_${Date.now()}`, ...payload };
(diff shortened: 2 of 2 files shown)


zeroknowledge0x:
Closing — submitted in error by automated agent.

PR #7698, opened 2026-06-17
Require auth for payment creation

### LINKED ISSUE #7697: Payment creation should require authentication
Parent bounty: #743

## Bug
`POST /api/payments` can be called without an access token. That lets unauthenticated callers create payment intent placeholders against the API.

## Expected
- Payment creation should require a valid bearer token.
- Missing or invalid tokens should receive 401 Unauthorized.
- The fix should remain scoped to the payment route.

wangedmund77-cmyk:
Closes #7697

## Summary
- add `authMiddleware` to `POST /api/payments`
- add regression coverage for missing and invalid bearer tokens returning 401
- keep authenticated payment creation covered so the guarded route still returns 201 for valid callers

## Validation
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 4 passing
- `git diff --check` -> clean

### DIFF
--- a/apps/api/src/routes/paymentRoutes.js
+++ b/apps/api/src/routes/paymentRoutes.js
@@ -1,6 +1,7 @@
 import { Router } from "express";
 import { createPayment } from "../controllers/paymentController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const paymentRoutes = Router();
 
-paymentRoutes.post("/", createPayment);
+paymentRoutes.post("/", authMiddleware, createPayment);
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a focused validation update after comparing the later same-issue payment-auth submission.

What changed:
- Added coverage for invalid bearer tokens returning 401.
- Added coverage that a valid bearer token still allows `POST /api/payments` to create the stubbed payment intent with a 201 response.
- Updated the PR validation notes to reflect the expanded test set.

Validation:
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 4 passing
- `git diff --check` -> clean

PR #7700, opened 2026-06-17
Require auth for job creation

### LINKED ISSUE #7699: Job creation should require authentication
Parent bounty: #743

## Bug
`POST /api/jobs` currently accepts job creation requests without any bearer token. That allows anonymous callers to create marketplace job records.

## Expected
- Job listing can remain public.
- Job creation should require a valid authenticated user.
- Missing or invalid tokens should receive 401 Unauthorized before validation or persistence.

wangedmund77-cmyk:
Closes #7699

## Summary
- require `authMiddleware` on `POST /api/jobs`
- keep `GET /api/jobs` publicly readable
- add regression coverage for missing and invalid bearer tokens returning 401
- keep authenticated job creation covered so valid callers still receive a 201 created job

## Validation
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 4 passing
- `git diff --check` -> clean

### DIFF
--- a/apps/api/src/routes/jobRoutes.js
+++ b/apps/api/src/routes/jobRoutes.js
@@ -1,7 +1,8 @@
 import { Router } from "express";
 import { getJobs, postJob } from "../controllers/jobController.js";
+import { authMiddleware } from "../middleware/auth.js";
 
 export const jobRoutes = Router();
 
 jobRoutes.get("/", getJobs);
-jobRoutes.post("/", postJob);
+jobRoutes.post("/", authMiddleware, postJob);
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a focused validation update after comparing the later same-issue job-auth submission.

What changed:
- Kept the existing proof that `GET /api/jobs` remains public while anonymous job creation is rejected.
- Added coverage for invalid bearer tokens returning 401.
- Added coverage that a valid bearer token still allows `POST /api/jobs` to create a job with a 201 response.

Validation:
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 4 passing
- `git diff --check` -> clean

PR #7702, opened 2026-06-17
Reject empty upload requests

### LINKED ISSUE #7701: Upload endpoint should reject requests without a file
Parent bounty: #743

## Bug
`POST /api/uploads` returns a 201 success response even when no file is attached. That makes failed uploads indistinguishable from successful uploads for API clients.

## Expected
- Requests without `req.file` should return 400 Bad Request.
- Successful uploads should only return `status: "uploaded"` when a file was actually received.

wangedmund77-cmyk:
Closes #7701

## Summary
- return 400 when `POST /api/uploads` has no uploaded file
- keep successful responses limited to real file uploads
- add regression coverage for both empty requests and valid multipart uploads

## Validation
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 3 passing
- `git diff --check` -> clean

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function uploadFile(req, res) {
+  if (!req.file) {
+    return fail(res, "File is required", 400);
+  }
+
   return ok(res, {
-    filename: req.file?.originalname ?? null,
-    status: req.file ? "uploaded" : "no-file"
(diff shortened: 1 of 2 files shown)


wangedmund77-cmyk:
Pushed a focused validation update after comparing the later same-issue upload submission.

What changed:
- Kept the 400 regression coverage for requests without `req.file`.
- Added a multipart upload success-path test so valid file uploads still return 201 with the uploaded filename.
- Updated the PR validation notes to reflect the expanded test set.

Validation:
- `node --test "src/tests/**/*.test.js"` from `apps/api` -> 3 passing
- `git diff --check` -> clean

PR #7873, opened 2026-06-17
Block admin role self-assignment during registration

### LINKED ISSUE #7872: Security: Privilege Escalation via role field in registerSchema (reissue via #743)
Parent bounty: #743

I found a critical vulnerability in the registration process. The registerSchema allows users to specify their own role, including "admin", which allows any user to grant themselves administrative privileges during registration.

Expected behavior: public registration should only allow normal marketplace roles such as client and freelancer. Administrative privileges must not be self-assignable through the registration payload.

ram-devv1:
Closes #7872

/claim #743

## Summary
- Remove `admin` from the public registration role schema.
- Return a 400 JSON failure envelope when registration payload validation fails.
- Add API regression coverage proving `role: "admin"` is rejected while normal client registration still succeeds.

## Validation
- `node --test apps/api/src/tests/*.test.js` (3 passing tests)
- `git diff --check`
- `npm run lint` (repo placeholder: no root lint configured)

## Demo
Short demo video: https://github.com/ram-devv1/bug-bounty/releases/download/pr-7873-demo/pr7873-registration-demo.mp4

The demo shows an admin self-assignment registration request receiving HTTP 400, while a normal client registration still receives HTTP 201 with a token.

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -1,10 +1,14 @@
 import { registerSchema, loginSchema } from "../validators/auth.js";
 import { loginUser, refreshToken, registerUser } from "../services/authService.js";
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 
 export async function register(req, res) {
-  const payload = registerSchema.parse(req.body);
-  const result = await registerUser(payload);
+  const parsed = registerSchema.safeParse(req.body);
+  if (!parsed.success) {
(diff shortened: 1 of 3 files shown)


ram-devv1:
Added the short demo video requested by the parent bounty:
https://github.com/ram-devv1/bug-bounty/releases/download/pr-7873-demo/pr7873-registration-demo.mp4

Current validation remains:
- git diff --check
- npm run lint (repo placeholder: no root lint configured)

PR #7895, opened 2026-06-18
fix: return defensive list service snapshots (#7878)

### LINKED ISSUE #7878: List services should return defensive snapshots
## Bug

The API service list functions return the module-level in-memory arrays and their record objects directly. Any in-process caller that receives a list can mutate the backing store by pushing/removing records or editing returned record fields without going through a service method.

Affected services:

- `listJobs()`
- `listProposals()`
- `listReviews()`
- `listMessages()`
- `listNotifications()`
- `listUsers()`

## Reproduction

```js
const listed = await listJobs();
listed.push({ id: "injected-record" });
listed[0].title = "mutated";

const later = await listJobs();
```

Expected: mutating the first list result does not alter the service store.

Actual: the later list can observe caller mutations because the returned list and record objects reference backing in-memory state.

## Expected

List services should return defensive snapshots of the current records. Callers should be able to read list results without mutating the underlying in-memory store or stored record fields.

## Scope

- Return defensive list snapshots from the in-memory list service functions.
- Copy array-valued record fields in returned snapshots.
- Add focused service tests proving array mutations and listed-record mutations do not alter backing store.
- Do not change route auth, validation, persistence, generated IDs, or create-service behavior beyond shared snapshot helper usage needed by this fix.

Refs #743. Reissue of #7876 for my own attempt, as requested by that issue.

wangedmund77-cmyk:
## Summary

Returns defensive snapshots from in-memory list services so callers cannot mutate backing arrays or stored record objects through list results.

Fixes #7878.

## Changes

- Added a shared `snapshotRecords` helper that copies records and array-valued fields.
- Applied it to jobs, proposals, reviews, messages, notifications, and users list services.
- Added focused service tests for array mutation, record mutation, and array-field mutation safety.
- Updates the API test script to run test files directly.

## Validation

- `npm test` in `apps/api` -> 3 tests passed

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,7 +1,9 @@
+import { snapshotRecords } from "./snapshot.js";
+
 const jobs = [];
 
 export async function listJobs() {
-  return jobs;
+  return snapshotRecords(jobs);
 }
 
 export async function createJob(payload) {
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,7 +1,9 @@
+import { snapshotRecords } from "./snapshot.js";
+
 const messages = [];
 
 export async function listMessages() {
-  return messages;
+  return snapshotRecords(messages);
 }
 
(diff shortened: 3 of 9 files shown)


wangedmund77-cmyk:
Pushed a focused coverage follow-up after comparing same-issue submissions for #7878.

What changed:
- Expanded `listSnapshots.test.js` so proposals, reviews, messages, and notifications are directly covered, not only jobs/users.
- Each affected list service now has regression coverage for returned-array mutation, returned-record mutation, and array-valued field mutation.
- This keeps the service implementation unchanged and aligns the evidence with the issue scope listing all six affected services.

Validation:
- `cd apps/api && npm test` -> 7 passed
- `git diff --check` -> clean

Commit: c1def17d09bf2ce1f32aef1f2c2497246d7afdc6

PR #7897, opened 2026-06-18

### LINKED ISSUE #7896: Reject inverted job budget ranges in job validation
﻿## Bug

## Expected behavior

## Scope

- Keep the change focused on job validation.
- Add regression coverage for create and update job schema parsing.

Parent bounty: #743

craiguito:
Closes #7896

/claim #743

## Summary
- Add shared budget range validation for job create and update schemas.
- Reject payloads where both budget values are present and `budgetMax` is lower than `budgetMin`.
- Return a 400 JSON failure response from `POST /api/jobs` when job validation fails instead of falling through to the generic 500 handler.
- Keep partial job updates valid when only one budget field is supplied.
- Preserve valid ordered and equal budget ranges.
- Add focused validator and route coverage for valid, equal, and inverted ranges.

## Validation
- `node --test src/tests/*.test.js` from `apps/api` (8 passing tests)
- `git diff --check`

### DIFF
--- a/apps/api/src/controllers/jobController.js
+++ b/apps/api/src/controllers/jobController.js
@@ -1,4 +1,4 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { createJobSchema } from "../validators/job.js";
 import { createJob, listJobs } from "../services/jobService.js";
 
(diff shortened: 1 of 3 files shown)


craiguito:
Updated this to cover the API boundary too. Inverted ranges now return 400 from POST /api/jobs, valid ranges still create normally, equal ranges pass, and partial update schemas only reject when both budget fields are present

PR #7899, opened 2026-06-18
Expose UI runtime entrypoint

### LINKED ISSUE #7898: Expose importable runtime entrypoint for @freelanceflow/ui
## Bug
The `@freelanceflow/ui` workspace package points `main` at TypeScript source and does not expose a runtime JavaScript entrypoint. Importing it directly from a workspace consumer fails because the package resolves to `src/index.ts` and then hits extensionless component exports.

## Expected behavior
- `import(@freelanceflow/ui)` works from a workspace consumer.
- The package exposes `Button` and `Card` from a JavaScript runtime entrypoint.
- Type declarations remain available for TypeScript consumers.

## Scope
Keep this limited to the UI package entrypoint and focused runtime import verification.

Parent bounty: #743

craiguito:
Closes #7898

/claim #743

## Summary
- Point `@freelanceflow/ui` package metadata at built `dist` entrypoints instead of TypeScript source.
- Add ESM and CommonJS exports for the package root.
- Add TypeScript declaration output for UI consumers.
- Update source re-exports for NodeNext module resolution.
- Commit generated `dist` artifacts so direct workspace imports resolve at runtime.

## Validation
- `npm run test:entrypoints -w packages/ui`
  - builds the package with `tsc -p tsconfig.json`
  - verifies `import('@freelanceflow/ui')` exposes `Button` and `Card`
  - verifies `require('@freelanceflow/ui')` exposes `Button` and `Card`
- `git diff --check origin/main...HEAD`


craiguito:
Added one more small pass here to cover the main gap I saw in the competing fixes.

This now supports both entrypoint styles:

```sh
npm run test:entrypoints -w packages/ui
```

That runs the ESM import path and the CommonJS require path:

```sh
import('@freelanceflow/ui')
require('@freelanceflow/ui')
```

Both return `Button,Card`. The ESM build still uses the generated `dist/*.js` files and the CJS path is isolated to `dist/index.cjs`, so Node does not have to load TypeScript source at runtime.

PR #7916, opened 2026-06-18
Fix API workspace test script glob

### LINKED ISSUE #7915: API test script treats tests directory as module
Parent Algora bounty: #743

## Bug
The API workspace test script currently runs `node --test src/tests`. With the Node test runner used here, that directory path is treated as a module entrypoint rather than discovering the test files, so `npm test -w apps/api` fails before running the existing health test.

## Steps to reproduce
```sh
npm test -w apps/api
```

## Expected
The API workspace test command should discover and run the existing `src/tests/*.test.js` files.

## Actual
`node --test src/tests` fails with `MODULE_NOT_FOUND` for the `src/tests` directory.

## Root cause
The npm script points the Node test runner at the tests directory. Passing the concrete test-file glob works with the current repository layout.

## Fix
Update the API workspace test script to run `node --test "src/tests/*.test.js"`.

snkk2x-collab:
/claim #743

Closes #7915

## Summary
- fix the API workspace test script so Node receives the actual `src/tests/*.test.js` files
- keep the change scoped to `apps/api/package.json`

## Root Cause
`node --test src/tests` treats the `src/tests` directory as the test entrypoint in this environment, which fails with `MODULE_NOT_FOUND` before the existing health test can run.

## Validation
- `npm test -w apps/api`
- `npm run build -w apps/web`
- `git diff --check`

## Demo
Short demo GIF: https://github.com/snkk2x-collab/bug-bounty/releases/download/pr-7916-demo/pr-7916-api-test-script.gif

The demo shows the API workspace test script running the real test file glob successfully.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",


snkk2x-collab:
Added the short demo GIF requested by the parent bounty: https://github.com/snkk2x-collab/bug-bounty/releases/download/pr-7916-demo/pr-7916-api-test-script.gif

Current validation remains: npm test -w apps/api, npm run build -w apps/web, git diff --check.

PR #7918, opened 2026-06-18
Validate API PORT configuration

### LINKED ISSUE #7917: Invalid PORT env crashes API startup
Parent Algora bounty: #743

## Bug
The API config parses `PORT` with `Number(process.env.PORT ?? 4000)`. If `PORT` is present but invalid, the parsed port becomes `NaN` and the API crashes during startup with `ERR_SOCKET_BAD_PORT`.

## Steps to reproduce
```sh
PORT=abc node apps/api/src/server.js
```

## Expected
Invalid `PORT` values should not be passed to `app.listen`. The API should fall back to the default port while preserving valid explicit ports like `0` or `8080`.

## Actual
Startup throws `RangeError [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536. Received type number (NaN)`.

## Root cause
`Number()` is used without validating that the result is an integer in the legal listen-port range.

## Fix
Add a small `parsePort` helper that returns the default port for invalid values and unit-test valid and invalid cases.

snkk2x-collab:
/claim #743

Closes #7917

## Summary
- add `parsePort` validation for API env configuration
- fall back to the default port when `PORT` is missing, non-numeric, out of range, or non-integer
- preserve valid explicit ports, including `0` for ephemeral test servers
- add focused Node test coverage for invalid and valid port inputs

## Root Cause
`Number(process.env.PORT ?? 4000)` can produce `NaN` when `PORT` is set to an invalid value, and that `NaN` is passed directly into `app.listen`, causing `ERR_SOCKET_BAD_PORT` during startup.

## Validation
- `node --test apps/api/src/tests/env.test.js`
- `PORT=abc node apps/api/src/server.js` starts on `http://localhost:4000` (then stopped)
- `npm run build -w apps/web`
- `git diff --check`

## Demo
Short demo GIF: https://github.com/snkk2x-collab/bug-bounty/releases/download/pr-7918-demo/pr-7918-validate-api-port.gif

The demo shows the focused PORT validation tests passing for invalid fallback and valid explicit ports.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,6 +1,16 @@
+export function parsePort(value, fallback = 4000) {
+  const port = Number(value ?? fallback);
+
+  if (!Number.isInteger(port) || port < 0 || port > 65535) {
+    return fallback;
+  }
+
+  return port;
+}
+
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
-  port: Number(process.env.PORT ?? 4000),
+  port: parsePort(process.env.PORT),
   jwtSecret: process.env.JWT_SECRET ?? "development-secret",
   stripeSecretKey: process.env.STRIPE_SECRET_KEY ?? "",
   databaseUrl: process.env.DATABASE_URL ?? ""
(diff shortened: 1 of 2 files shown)


snkk2x-collab:
Added the short demo GIF requested by the parent bounty: https://github.com/snkk2x-collab/bug-bounty/releases/download/pr-7918-demo/pr-7918-validate-api-port.gif

Current validation remains: node --test apps/api/src/tests/env.test.js, PORT=abc startup check, npm run build -w apps/web, git diff --check.

Issue #7961, opened 2026-06-19
Patch PostCSS audit advisory from Next dependency lockfile


peterxing:
## Bug

The web workspace resolves `postcss@8.4.31` through `next@16.2.6`, and `npm audit --omit=dev` reports GHSA-qx2v-qp2m-jg93 for PostCSS. This leaves the repository with a moderate XSS advisory in the production dependency tree.

## Evidence

- `npm audit --omit=dev` reports `postcss` vulnerable for range `<8.5.10`.
- `npm ls postcss --all` shows the web workspace resolves PostCSS through Next.

## Expected fix

Add a root npm override for the patched PostCSS release and refresh `package-lock.json` so the workspace resolves `postcss@8.5.10` while preserving the existing Next version.

Parent bounty: #743

Duplicate/parallel claim of creator-limited issue #1779, per its instruction to create another issue with the same contents before working.

peterxing:
/attempt #7961

Plan: add a root npm override for PostCSS 8.5.10, refresh the lockfile node for `postcss`, verify the targeted GHSA-qx2v-qp2m-jg93 advisory drops out of `npm audit --omit=dev`, and keep Next at 16.2.6.

peterxing:
Submitted focused fix in PR #7962: https://github.com/SecureBananaLabs/bug-bounty/pull/7962

Verification included in the PR body: install passes, the PostCSS advisory drops from npm audit, PostCSS resolves to 8.5.10, the web build passes, and diff whitespace is clean.

PR #7962, opened 2026-06-19
fix: patch PostCSS audit advisory via override

peterxing:
/claim #7961

## Summary
- Add a root npm override pinning Next's PostCSS dependency to `8.5.10`.
- Refresh the lockfile `node_modules/postcss` entry to the patched release while keeping `next@16.2.6` unchanged.

## Verification
- `npm ci --ignore-scripts`
- `npm audit --omit=dev` now reports only unrelated `multer` and `qs` advisories; the PostCSS GHSA-qx2v-qp2m-jg93 / Next-via-PostCSS advisory is gone.
- `node -p require('./node_modules/postcss/package.json').version` -> `8.5.10`
- `npm run build -w apps/web` passes
- `git diff --check`

Note: `npm ls postcss --all` shows `postcss@8.5.10` under `next@16.2.6`, but npm exits with `ELSPROBLEMS` because Next's published package metadata still declares the exact `postcss@8.4.31` dependency. The install, audit target, and web build pass with the override/lockfile resolution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1672,9 +1672,9 @@
       "license": "ISC"
     },
     "node_modules/postcss": {
-      "version": "8.4.31",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.4.31.tgz",
-      "integrity": "sha512-PS08Iboia9mts/2ygV3eLpY5ghnUcfLV/EXTOW1E2qYxJKGGBUtNjN76FYHnMs36RmARn41bC0AZmn+rR0OVpQ==",
+      "version": "8.5.10",
+      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.5.10.tgz",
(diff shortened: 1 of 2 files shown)


peterxing:
Current status readback for reviewers: PR #7962 remains the submitted fix for #7961, `update-leaderboard` is green on head `ac11300`, and there are no reviewer comments or requested changes on the PR.

The verification in the PR body still applies: `npm ci --ignore-scripts`, targeted `npm audit --omit=dev` with the PostCSS advisory cleared, PostCSS resolving to `8.5.10`, web build passing, and `git diff --check` clean. The only noted caveat is npm's `ELSPROBLEMS` on `npm ls postcss --all` because Next's published metadata still declares the older exact dependency while the root override/lockfile resolution installs the patched version.

No live-system testing, secrets, credentials, payments, wallet activity, or production access were used.

Issue #8002, opened 2026-06-19
Rate limiter should run before JSON body parsing


LokiLiu:
Parent bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

## Bug
The API rate limiter is registered after `express.json()` in `apps/api/src/app.js`. Malformed or large JSON requests can therefore consume body-parser resources before the request is rate limited.

## Expected behavior
- Apply `apiLimiter` before JSON body parsing.
- Preserve the existing middleware stack and API behavior.
- Add a focused regression test or source-order check proving the limiter is registered before `express.json()`.

## Scope
Keep the change limited to `apps/api/src/app.js` and focused API test coverage.

LokiLiu:
/attempt #8002

PR: https://github.com/SecureBananaLabs/bug-bounty/pull/8006

Validation:
- `node --test "apps/api/src/tests/*.test.js"`

PR #8014, opened 2026-06-19
Apply rate limiter before JSON parser

### LINKED ISSUE #8010: Rate limiter should run before JSON body parsing
Parent bounty: #743

## Bug
The API rate limiter is registered after `express.json()` in `apps/api/src/app.js`. Malformed or large JSON requests can therefore consume body-parser resources before the request is rate limited.

## Expected behavior
- Apply `apiLimiter` before JSON body parsing.
- Preserve the existing middleware stack and API behavior.
- Add a focused regression test or source-order check proving the limiter is registered before `express.json()`.

## Scope
Keep the change limited to `apps/api/src/app.js` and focused API test coverage.

LokiLiu:
Fixes #8010

## Summary
- Move `apiLimiter` before `express.json()` so rate limiting runs before body parsing.
- Add a focused source-order regression test for the middleware registration order.

## Validation
- `git diff --check`

Note: `npm test -w apps/api` currently fails because the existing script invokes `node --test src/tests`, which this Node version treats as a module path rather than discovering test files. This PR leaves that separate issue untouched.

### DIFF
--- a/apps/api/src/app.js
+++ b/apps/api/src/app.js
@@ -20,8 +20,8 @@ export function createApp() {
 
   app.use(helmet());
   app.use(cors());
-  app.use(express.json());
   app.use(apiLimiter);
+  app.use(express.json());
 
   app.get("/health", (req, res) => {
     res.status(200).json({ ok: true, service: "api" });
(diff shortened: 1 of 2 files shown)


LokiLiu:
Closing as a duplicate of #8006. The original issue #8002 is owned by this account and already has the focused PR.

PR #8016, opened 2026-06-19
Reuse registered user id for token subject

### LINKED ISSUE #8012: Registration token subject should match returned user id
Parent bounty: #743

## Bug
`registerUser()` calls `Date.now()` separately for the returned user `id` and JWT `sub` claim. If those calls cross a millisecond boundary, the token subject can differ from the returned user id.

## Expected behavior
- Generate the user id once during registration.
- Reuse that exact id as the JWT `sub` claim.
- Preserve the existing response shape.
- Add focused service coverage proving the token subject matches the returned id.

## Scope
Keep the change limited to registration behavior and direct tests.

LokiLiu:
Fixes #8012

## Summary
- Generate the registration user id once.
- Reuse that exact id for the access token `sub` claim.
- Add focused service coverage for a cross-millisecond `Date.now()` sequence.

## Validation
- `git diff --check`

### DIFF
--- a/apps/api/src/services/authService.js
+++ b/apps/api/src/services/authService.js
@@ -2,11 +2,13 @@ import { signAccessToken } from "../utils/jwt.js";
 
 export async function registerUser(payload) {
   // TODO: persist new user via Prisma
+  const userId = `usr_${Date.now()}`;
+
   return {
-    id: `usr_${Date.now()}`,
+    id: userId,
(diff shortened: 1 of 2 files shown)


LokiLiu:
Closing as a duplicate of #8009. The original issue #8005 is owned by this account and already has the focused PR.

PR #8017, opened 2026-06-19
Ignore notification internal fields from payload

### LINKED ISSUE #8013: Notification creation should ignore caller-controlled internal fields
Parent bounty: #743

## Bug
`createNotification()` spreads caller input after assigning internal fields, so callers can override the generated `id` and default `read: false` values.

## Expected behavior
- Caller-provided `id` and `read` fields must be ignored during notification creation.
- Generated notification IDs and unread defaults should always win.
- Ordinary notification payload fields should still be preserved.
- Add focused service coverage for the internal-field override case.

## Scope
Keep the change limited to notification creation behavior and direct tests.

LokiLiu:
Fixes #8013

## Summary
- Ignore caller-provided `id` and `read` fields during notification creation.
- Preserve ordinary notification payload fields.
- Add direct service coverage for the internal-field override case.

## Validation
- `git diff --check`

### DIFF
--- a/apps/api/src/services/notificationService.js
+++ b/apps/api/src/services/notificationService.js
@@ -5,7 +5,12 @@ export async function listNotifications() {
 }
 
 export async function createNotification(payload) {
-  const notification = { id: `ntf_${Date.now()}`, read: false, ...payload };
+  const { id, read, ...notificationPayload } = payload;
+  const notification = {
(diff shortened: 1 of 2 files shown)


LokiLiu:
Closing as a duplicate of #8008. The original issue #8004 is owned by this account and already has the focused PR.

PR #8075, opened 2026-06-20
[codex] Log the actual ephemeral API port

### LINKED ISSUE #8074: API startup should log the actual ephemeral port
## Problem

When the API is started with `PORT=0`, Node correctly asks the operating system to select an available ephemeral port, but the startup message still uses the configured value and prints:

```text
API listening on http://localhost:0
```

Port `0` is not the port clients can connect to. This makes the documented startup output unusable for integration tests, preview environments, and other workflows that intentionally request an ephemeral port.

## Reproduction

```bash
PORT=0 node apps/api/src/server.js
```

The process binds successfully, but the logged URL ends in `:0` instead of the positive port returned by `server.address()`.

## Expected behavior

After the server starts, the log should contain the actual bound port selected by the operating system. Explicit non-zero ports should continue to be reported unchanged.

## Proposed fix

- Capture the `http.Server` returned by `app.listen()`.
- Read the bound address inside the listening callback and log its actual port.
- Add a regression test that starts the API with `PORT=0` and verifies a positive reachable port is reported.

Related bounty: #743

xingdi614-alt:
## Summary

- log the actual bound port returned by `server.address()` after API startup
- preserve the configured port as a fallback for non-TCP server addresses
- add a regression test that starts with `PORT=0`, checks the reported port is positive, and reaches `/health` through that exact port

## Validation

- `node --test apps/api/src/tests/*.test.js` (2 passed)
- `npm run build -w apps/web` (passed)
- `git diff --check` (passed)

## Demo

A short demo video showing ephemeral-port startup, the reachable health endpoint, and the passing regression test is attached in the PR comments.

Closes #8074

## Bounty scope

This resolves scoped issue #8074 under parent bounty #743.

/claim #743

### DIFF
--- a/apps/api/src/server.js
+++ b/apps/api/src/server.js
@@ -5,8 +5,10 @@ import { createApp } from "./app.js";
 async function bootstrap() {
   await connectDb();
   const app = createApp();
-  app.listen(env.port, () => {
-    console.log(`API listening on http://localhost:${env.port}`);
+  const server = app.listen(env.port, () => {
+    const address = server.address();
+    const port = typeof address === "object" && address !== null ? address.port : env.port;
+    console.log(`API listening on http://localhost:${port}`);
   });
 }
 
--- a/apps/api/src/tests/server.test.js
+++ b/apps/api/src/tests/server.test.js
@@ -1,0 +1,62 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { spawn } from "node:child_process";
+import { once } from "node:events";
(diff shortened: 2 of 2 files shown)


xingdi614-alt:
Demo video: [demo-8074.mp4](https://github.com/xingdi614-alt/bug-bounty/releases/download/demo-pr-8075/demo-8074.mp4)

It shows the API starting with `PORT=0`, reporting the operating-system-selected port, serving `/health` through that exact port, and the focused regression suite passing.

PR #8103, opened 2026-06-21
[codex] Fix API test discovery

### LINKED ISSUE #8102: API test script should discover test files
## Problem

The API workspace's standard test command is configured as:

```json
"test": "node --test src/tests"
```

On current Node.js releases, the test runner treats `src/tests` as a module path instead of discovering the test files inside it. As a result, running the repository's documented root command fails before executing any tests:

```text
Error: Cannot find module '.../apps/api/src/tests'
```

## Reproduction

```bash
npm test
```

## Expected behavior

The standard test command should discover and execute the API's `*.test.js` files and exit successfully when they pass.

## Proposed fix

Update the API workspace test script to pass the test files to `node --test` explicitly, then verify both the workspace and root test commands.

Related bounty: #743

xingdi614-alt:
## Summary

- pass the API test files to Node's test runner explicitly
- make the documented root `npm test` command execute the existing suite
- preserve the workspace-level test command

## Root cause

The script passed the `src/tests` directory to `node --test`. Current Node.js releases resolve that argument as a module path instead of discovering test files under the directory, so the command failed with `MODULE_NOT_FOUND` before running any test.

## Impact

Contributors and CI can use the repository's standard test command again. The change is limited to test discovery and does not affect runtime behavior.

## Validation

- `npm test` (1 passed)
- `npm run test -w apps/api` (1 passed)
- `git diff --check` (passed)

Closes #8102

## Bounty scope

This resolves scoped issue #8102 under parent bounty #743.

/claim #743

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",


xingdi614-alt:
Demo video: [demo-pr-8103.m4v](https://github.com/xingdi614-alt/bug-bounty/releases/download/demo-pr-8103/demo-pr-8103.m4v)

It shows the original `node --test src/tests` failure, the one-line test-discovery fix, and the standard root `npm test` command discovering and passing the API suite.

PR #8239, opened 2026-06-22
fix: enforce default job status on creation (mass assignment)

### LINKED ISSUE #8235: Job creation should enforce default status field
## Bug

The `createJob` service in `apps/api/src/services/jobService.js` allows the client to override the `status` field during job creation because the spread operator `...payload` is applied after the default value.

### Current code

```js
```

A client can send `{"status": "closed"}` in the request body and the job will be created with status "closed" instead of the default "open".

### Expected behavior

The `status` field should always default to `"open"` on creation and the client should not be able to override it.

### Scope

- File: `apps/api/src/services/jobService.js`
- Fix: Reorder the spread so default fields always win

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue Low Handing Fruit Automation#743 for more information.

Parent bounty: #743

kish-00:
## Summary

Fixes mass assignment vulnerability in `createJob` where clients could override the job `status` field by sending it in the request body.

## The Bug

The spread operator `...payload` was applied **after** the default `status: "open"` field, meaning a client sending `{"status": "closed"}` would override the default.

```js
// Before (bug): payload overrides defaults

// After (fix): defaults always win
const job = { ...payload, id: `job_${Date.now()}`, status: "open" };
```

Root cause: mass assignment — raw `req.body` was spread directly into the object without sanitization.

## Testing

- Verified default `status: "open"` cannot be overridden
- Backward compatible — existing valid fields (title, description, budget) still work normally

Closes #8235
/bounty $700

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -5,7 +5,7 @@ export async function listJobs() {
 }
 
 export async function createJob(payload) {
-  const job = { id: `job_${Date.now()}`, status: "open", ...payload };
+  const job = { ...payload, id: `job_${Date.now()}`, status: "open" };
   jobs.push(job);
   return job;
 }


kish-00:
Closing — this is a bounty farm. 4,300+ open PRs, 0 ever merged. The leaderboard gamification doesn't change the fact that terms require a merge for payment and merges never happen.

HeyBanditoz:
@kish-00 It's a legitimate repo. Re-open this pull request.

PR #8375, opened 2026-06-25
fix: prevent public registration from requesting admin role tokens

### LINKED ISSUE #8369: Public registration should not allow admin role tokens
The public registration schema currently accepts `role: "admin"`. Because `registerUser` returns the requested role and signs it into the access token, any public caller can request an admin role during signup and receive an admin-scoped token.

Expected behavior:
- public registration should only allow regular public roles such as `client` and `freelancer`
- requests for `role: "admin"` should return a clear 400 response
- normal public registration should continue to work
- the behavior should be covered with focused API tests

/attempt #743

BWM0223:
Closes #8369

Adds restrictPublicRole middleware allowing only client and freelancer roles. Admin role requests return 400.


BWM0223:

PR #8376, opened 2026-06-25
fix: return 400/413 for malformed and oversized JSON request bodies

### LINKED ISSUE #8365: Malformed JSON requests should return 400 instead of 500

Expected behavior:
- return a clear 400 JSON response for malformed JSON request bodies
- return a clear 413 JSON response for oversized JSON bodies when the parser reports that condition
- keep unexpected errors on the existing 500 response path
- cover the behavior with focused API tests

/attempt #743

BWM0223:
Closes #8365

Adds jsonErrorHandler catching SyntaxError (400) and entity.too.large (413) before 500 handler.


BWM0223:

PR #8378, opened 2026-06-25
fix: return 400 when upload endpoint receives no file

### LINKED ISSUE #8367: Upload endpoint should reject requests without a file
The upload endpoint currently returns a successful 201 response even when the request does not include a `file` field. That makes an invalid client request look like a completed upload and returns `status: "no-file"` in a success payload.

Expected behavior:
- return a clear 400 JSON error when no file is provided
- keep successful uploads returning 201 with the uploaded filename
- cover both missing-file and successful-upload behavior with focused API tests
- make the API test script target `*.test.js` files explicitly so the tests can run on Windows as well as Unix shells

/attempt #743

BWM0223:
Closes #8367

Rejects requests with no file field with 400 instead of returning misleading 201 with status: no-file.

### DIFF
--- a/apps/api/src/controllers/uploadController.js
+++ b/apps/api/src/controllers/uploadController.js
@@ -1,8 +1,6 @@
-import { ok } from "../utils/response.js";
-
-export async function uploadFile(req, res) {
-  return ok(res, {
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8379, opened 2026-06-25
fix: reject whitespace-only fields in job creation validation

### LINKED ISSUE #8363: Job validation should reject whitespace-only fields
## Bug

Job creation validation currently accepts whitespace-only values for fields that are supposed to be meaningful identifiers or text:

- `title` uses `z.string().min(4)`, so `"    "` passes.
- `description` uses `z.string().min(10)`, so ten spaces pass.
- `categoryId` uses `z.string().min(1)`, so `" "` passes.
- `skills` uses `z.string().min(1)`, so whitespace-only skills pass.

That lets `/api/jobs` persist visually empty or unusable job data while still returning a successful creation response.

## Expected

Job text fields should be trimmed before length validation, and whitespace-only title, description, category id, or skills should be rejected.

## Suggested fix

Use `z.string().trim().min(...)` for the relevant job schema fields and add focused regression tests that prove whitespace-only values are rejected while normal padded values are trimmed.

## Validation target

```bash
npm test -w apps/api
git diff --check
```

Parent bounty: #743

BWM0223:
Closes #8363

Adds rejectWhitespaceFields middleware. title, description, categoryId and skills must not be whitespace-only.

### DIFF
--- a/apps/api/src/middleware/rejectWhitespace.js
+++ b/apps/api/src/middleware/rejectWhitespace.js
@@ -1,0 +1,11 @@
+import{fail}from"../utils/response.js";
+export function rejectWhitespaceFields(fields){
+  return(req,res,next)=>{
+    for(const field of fields){
+      const v=req.body?.[field];
+      if(typeof v==="string"&&v.trim().length===0)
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8380, opened 2026-06-25
fix: return 400 for malformed JSON bodies instead of 500

### LINKED ISSUE #8371: Malformed JSON requests should return client errors

Expected behavior:
- return a clear 400 JSON response for malformed JSON request bodies
- return a clear 413 JSON response for oversized JSON bodies when the parser reports that condition
- keep unexpected errors on the existing 500 response path
- cover the behavior with focused API tests

Suggested fix:
- handle Express JSON parser `entity.parse.failed` errors before the generic 500 path
- handle `entity.too.large` errors with a 413 response
- add regression tests for both paths

Parent bounty: #743
Related prior scoped issue: #8365

/attempt #743

BWM0223:
Closes #8371

Adds jsonErrorHandler middleware catching SyntaxError from express.json() and body-parser. Returns 400 for malformed JSON, 413 for oversized bodies.

### DIFF
--- a/apps/api/src/middleware/jsonErrorHandler.js
+++ b/apps/api/src/middleware/jsonErrorHandler.js
@@ -1,0 +1,8 @@
+import{fail}from"../utils/response.js";
+export function jsonErrorHandler(err,req,res,next){
+  if(err instanceof SyntaxError&&err.status===400&&"body"in err)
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8382, opened 2026-06-25
fix: return 400 with field errors for Zod validation

### LINKED ISSUE #8335: Return 400 for Zod validation errors

Parent bounty: #743

## Problem

Controllers use Zod schemas with `.parse(req.body)` for request validation. When validation fails, Zod throws a `ZodError`. The global API error handler currently treats every unhandled error as an unexpected server error and returns HTTP 500:

```js
return res.status(500).json({
  success: false,
  message: "Unexpected server error"
});
```

Invalid client payloads should not be reported as server failures, and returning 500 makes API clients and monitoring treat bad input as backend instability.

## Expected behavior

Zod validation failures should return a client error, for example HTTP 400 with a validation-oriented message.

## Suggested fix

- Detect `ZodError` in the global error handler.
- Return HTTP 400 for validation failures.
- Add a regression test that passes a `ZodError` into the handler and verifies the 400 response.

BWM0223:
Closes #8335

### DIFF
--- a/apps/api/src/middleware/zodErrorHandler.js
+++ b/apps/api/src/middleware/zodErrorHandler.js
@@ -1,0 +1,8 @@
+import{fail}from"../utils/response.js";
+export function zodErrorHandler(err,req,res,next){
+  if(err?.name==="ZodError"||err?.issues){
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8383, opened 2026-06-25
fix: require valid JWT before issuing refresh tokens

### LINKED ISSUE #8333: Require authentication before issuing refresh tokens

Parent bounty: #743

## Problem

The auth routes expose `POST /api/auth/refresh` without any authentication middleware:

```js
authRoutes.post("/refresh", refresh);
```

The controller calls `refreshToken()`, which returns a signed access token. That means an anonymous request can mint an access token without proving it owns any existing session or credential.

## Expected behavior

The refresh endpoint should require an authenticated request before issuing a new token. A request without credentials should receive 401.

## Suggested fix

- Apply `authMiddleware` to the refresh route.
- Add tests showing anonymous refresh requests are rejected and authenticated refresh requests still succeed.

BWM0223:
Closes #8333

### DIFF
--- a/apps/api/src/middleware/requireAuth.js
+++ b/apps/api/src/middleware/requireAuth.js
@@ -1,0 +1,12 @@
+import jwt from"jsonwebtoken";
+import{fail}from"../utils/response.js";
+export function requireAuth(req,res,next){
+  const auth=req.headers.authorization;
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8384, opened 2026-06-25
fix: require admin role for admin routes

### LINKED ISSUE #8329: Require admin role for admin routes

Parent bounty: #743

## Problem

`adminRoutes` currently applies only `authMiddleware`:

```js
adminRoutes.use(authMiddleware);
adminRoutes.get("/metrics", metrics);
```

`authMiddleware` verifies that a token exists, but it does not require `req.user.role` to be `admin`. As a result, any authenticated client or freelancer token can access admin-only metrics.

## Expected behavior

Admin routes should require both a valid token and an explicit admin role check before handlers run.

## Suggested fix

- Add a small role guard middleware, for example `requireRole("admin")`.
- Apply it to `adminRoutes` after `authMiddleware`.
- Add tests proving non-admin and missing users receive 403 while admin users pass through.

BWM0223:
Closes #8329

### DIFF
--- a/apps/api/src/middleware/requireAdmin.js
+++ b/apps/api/src/middleware/requireAdmin.js
@@ -1,0 +1,6 @@
+import{fail}from"../utils/response.js";
+export function requireAdmin(req,res,next){
+  if(!req.user) return fail(res,"Authentication required",401);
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8385, opened 2026-06-25
fix: require JWT_SECRET in production startup

### LINKED ISSUE #8337: Require JWT_SECRET in production

Parent bounty: #743

## Problem

The API configuration silently falls back to a hard-coded JWT secret when `JWT_SECRET` is missing:

```js
```

That fallback is useful for local development, but it is unsafe in production. If a production deployment forgets to set `JWT_SECRET`, access tokens are signed with a public, predictable secret.

## Expected behavior

Production startup should fail fast when `JWT_SECRET` is missing. The development fallback should only be available outside production.

## Suggested fix

- Add a small helper that requires security-sensitive env vars in production.
- Use it for `jwtSecret`.
- Add regression tests for the development fallback, configured secret, and production missing-secret failure.

BWM0223:
Closes #8337

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,9 @@
-export const env = {
-  nodeEnv: process.env.NODE_ENV ?? "development",
-  port: Number(process.env.PORT ?? 4000),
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8386, opened 2026-06-25
fix: add requireAdmin middleware for admin metrics

### LINKED ISSUE #8344: Admin metrics should reject non-admin tokens
## Bug

The `/api/admin/metrics` route only requires a valid bearer token. It does not verify that the authenticated user has an admin role:

```js
adminRoutes.use(authMiddleware);
adminRoutes.get("/metrics", metrics);
```

Because `authMiddleware` accepts any valid access token and stores the decoded token on `req.user`, a non-admin token such as `{ "sub": "usr_client", "role": "client" }` can access administrative metrics.

## Expected

`/api/admin/*` routes should reject authenticated non-admin users with HTTP 403 while continuing to reject anonymous requests with HTTP 401. Admin tokens should continue to access `/api/admin/metrics`.

## Scope

- Add an admin-role authorization middleware that checks `req.user.role`.
- Apply it only to admin routes.
- Add focused API regression tests for anonymous, non-admin, and admin access to `/api/admin/metrics`.

## Parent Bounty

Refs #743.

BWM0223:
Closes #8344

### DIFF
--- a/apps/api/src/middleware/requireAdmin.js
+++ b/apps/api/src/middleware/requireAdmin.js
@@ -1,0 +1,5 @@
+import{fail}from"../utils/response.js";
+export function requireAdmin(req,res,next){
+  if(!req.user||req.user.role!=="admin") return fail(res,"Admin access required",403);
+  return next();
+}


BWM0223:

PR #8387, opened 2026-06-25
fix: return 400 when upload receives no file

### LINKED ISSUE #8350: Upload endpoint should reject requests without a file
`POST /api/uploads` currently returns `201` with `status: "no-file"` when the multipart request omits the `file` field. A successful upload endpoint should not report success when no upload was accepted.

Proposed fix:
- Return a `400` response when `req.file` is missing.
- Keep the existing `201` success response for valid uploads.
- Add API regression coverage for missing-file and valid-file requests.

Parent bounty: #743

BWM0223:
Closes #8350


BWM0223:

PR #8389, opened 2026-06-25
fix: prevent public registration from granting admin

### LINKED ISSUE #8327: Prevent public registration from granting admin roles

Parent bounty: #743

## Problem

The public registration schema currently allows `role: "admin"`:

```js
role: z.enum(["client", "freelancer", "admin"]).default("client")
```

`registerUser` then copies `payload.role` into the response and signs it into the access token. A caller can therefore submit a registration payload with `role: "admin"` and receive an admin-scoped token.

## Expected behavior

Public signup should only allow normal user roles, such as `client` and `freelancer`. Admin accounts should be provisioned through a separate privileged path.

## Suggested fix

- Remove `admin` from the public `registerSchema` role enum.
- Keep the default public role as `client`.
- Add regression tests proving admin signup is rejected and normal/default registrations still validate.

BWM0223:
Closes #8327

### DIFF
--- a/apps/api/src/middleware/restrictPublicRole.js
+++ b/apps/api/src/middleware/restrictPublicRole.js
@@ -1,0 +1,7 @@
+import{fail}from"../utils/response.js";
+const ALLOWED=new Set(["client","freelancer"]);
+export function restrictPublicRole(req,res,next){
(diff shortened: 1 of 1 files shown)


BWM0223:

PR #8391, opened 2026-06-25
fix: limit in-memory upload size to 5MB

### LINKED ISSUE #8331: Limit in-memory upload size

Parent bounty: #743

## Problem

The upload route currently uses `multer.memoryStorage()` without any `limits.fileSize` setting:

```js
const upload = multer({ storage: multer.memoryStorage() });
```

Because uploaded files are buffered in memory, accepting unbounded multipart bodies can exhaust API memory and turn oversized uploads into a denial-of-service risk. If a Multer size error is added without route handling, the current global error handler would also turn it into a generic 500 response.

## Expected behavior

The API should reject oversized uploads before buffering arbitrary amounts of data and return an explicit client error, such as HTTP 413.

## Suggested fix

- Set a conservative `limits.fileSize` value on the in-memory multer instance.
- Convert `LIMIT_FILE_SIZE` into a 413 response.
- Add a regression test for oversized multipart uploads.

BWM0223:
Closes #8331

### DIFF
--- a/apps/api/src/middleware/uploadLimiter.js
+++ b/apps/api/src/middleware/uploadLimiter.js
@@ -1,0 +1,10 @@
+import multer from"multer";
+const MAX_FILE_SIZE_BYTES=5*1024*1024; // 5 MB
+export const upload=multer({
+  storage:multer.memoryStorage(),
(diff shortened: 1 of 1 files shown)


BWM0223:

Issue #8516, opened 2026-06-25
Cap and validate GET /api/search query length


cwwjacobs:
Parent bounty: #743

/attempt #743

## Bug

The `GET /api/search` endpoint passes `req.query.q` into the search path without validating type or bounding length. Very large query strings can force unnecessary work or memory pressure inside the search path.

## Expected behavior

- Search queries are normalized by trimming surrounding whitespace.
- Array or multi-value query inputs are rejected with a client error.
- Non-string query inputs are rejected with a client error.
- Queries longer than 200 characters are rejected with a client error before reaching search logic.
- Safe, normal-length search queries continue to return the existing response shape.
- The fix stays scoped to the search controller or validation boundary and includes focused regression coverage.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue Low Handing Fruit Automation #743 for more information.

cwwjacobs:
Working on this in PR #8517.

cwwjacobs:
Withdrawn.

PR #8517, opened 2026-06-25
Cap and validate GET /api/search query length

cwwjacobs:
Parent bounty: #743
Creator-limited issue: #8516
## Summary

Adds input validation for the `GET /api/search` query parameter so oversized or malformed query values are rejected before reaching the search path.

## Changes

- Normalizes valid string search queries by trimming surrounding whitespace.
- Rejects array or multi-value `q` inputs with a client error.
- Rejects non-string `q` inputs with a client error.
- Rejects search queries longer than 200 characters.
- Preserves existing behavior for normal valid queries.
- Adds focused regression coverage for valid, missing, trimmed, oversized, and multi-value query inputs.

## Tests

Command run (from `apps/api`):

```bash
node --test 'src/tests/**/*.test.js'
```

Result:

```text
✔ GET /api/search returns existing shape for valid query
✔ GET /api/search treats missing query as empty string
✔ GET /api/search trims surrounding whitespace
✔ GET /api/search rejects query longer than 200 characters
✔ GET /api/search rejects array query input
ℹ tests 6
ℹ pass 6
ℹ fail 0
```

## Scope

This PR is intentionally limited to the search query validation boundary. It does not refactor search internals, change unrelated endpoints, or perform live probing.

## Bounty

Parent bounty: #743
Creator-limited issue: #8516

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,11 @@
-import { ok } from "../utils/response.js";
+import { ok, fail } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
+import { searchQuerySchema } from "../validators/search.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const parsed = searchQuerySchema.safeParse(req.query);
+  if (!parsed.success) {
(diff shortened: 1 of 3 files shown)


cwwjacobs:
Withdrawn.

PR #8837, opened 2026-06-26
fix: make api test script discover test files

### LINKED ISSUE #8836: API test script points at src/tests directory and fails under node --test
The API workspace test script currently runs:

```sh
node --test src/tests
```

On Windows with Node's test runner this treats `src/tests` as a module path instead of discovering the contained `*.test.js` files, so the root `npm test` fails before any tests run with `MODULE_NOT_FOUND` for `apps/api/src/tests`.

Repro:

```sh
npm install
npm test
```

Expected: the existing `src/tests/health.test.js` test should be discovered and executed.

Proposed fix: target test files explicitly with a cross-platform glob, e.g. `node --test "src/tests/**/*.test.js"`.

Referencing #743.

dollarop:
## Summary

Fixes the API workspace test script so Node's test runner targets the actual test files instead of the `src/tests` directory path.

Closes #8836
References #743

## Problem

The current script runs `node --test src/tests`. On Windows with Node's test runner, that directory argument is resolved as a module path and fails before discovery with `MODULE_NOT_FOUND` for `apps/api/src/tests`.

## Changes

- Updated `apps/api/package.json` to run `node --test "src/tests/**/*.test.js"`.
- This keeps the existing test layout and lets the current `health.test.js` file run from the root `npm test` command.

## Verification

- `npm install`
- `npm test` passes locally with 1 passing API test.

## Bounty claim

/claim #743

## Demo

Before/after API test script demo

[Open the animated demo](https://github.com/dollarop/bug-bounty/blob/fix-api-test-script-glob-743/docs/demo/api-test-script-demo.png)

The 7.5-second demo shows the original `MODULE_NOT_FOUND` failure, the exact test-script change, and the passing `npm test` result (1 test, 0 failures).


dollarop:
/claim #743

The claim marker is also present in the PR description. Posting it once here because it was added after the PR was opened and no bounty label or claim index appeared.

dollarop:
Hi maintainers, a concise review follow-up on this creator-scoped submission for #8836 under the official #743 bounty: the PR remains open, non-draft, mergeable/CLEAN, includes the required \/claim #743\ marker and short before/after demo, and \
pm test\ passes with the corrected cross-platform test-file discovery. There are no requested changes pending from my side. When convenient, could you review it and let me know if anything else is required for eligibility?

PR #8879, opened 2026-06-27
Return client errors for invalid JSON bodies

### LINKED ISSUE #8877: Malformed JSON requests should return client errors (Dreamstore2046 reissue via #743)

Expected behavior:
- Return a clear `400` JSON response for malformed JSON request bodies.
- Return a clear `413` JSON response for oversized JSON bodies when the parser reports that condition.
- Keep unexpected errors on the existing `500` response path.
- Add focused API tests verifying that malformed requests get 400 and oversized requests get 413.

Parent bounty: #743

Dreamstore2046:
/claim #743

Closes #8877

Replacement for #8878 because its `update-leaderboard` workflow generated the leaderboard commit but failed while pushing that generated commit to `main`, and contributors do not have permission to rerun that `pull_request_target` job.

## Summary
- Return `400` JSON responses for malformed JSON body parser errors.
- Return `413` JSON responses for oversized JSON body parser errors.
- Keep unexpected API errors on the existing `500` fallback path.
- Add focused API regression coverage for malformed and oversized request bodies.
- Make the API test script target `src/tests/*.test.js` explicitly so the workspace test command runs on this checkout.

## Root Cause
The shared API error handler treated body-parser client errors the same as unexpected server failures, so malformed JSON and payload-too-large errors fell through to the generic 500 response.

## Validation
- `npm test -w apps/api`
- `node --check apps/api/src/middleware/errorHandler.js`
- `node --check apps/api/src/tests/jsonBodyErrors.test.js`

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,23 @@
 export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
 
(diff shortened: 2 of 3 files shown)


Dreamstore2046:
Closing in favor of #8880, which contains the same validated fix on a fresh commit SHA so it is not associated with #8878's failed leaderboard check run.

PR #8982, opened 2026-06-27
fix: return immutable copies from create service records

### LINKED ISSUE #1159: Create services expose mutable stored records
## Bug

The in-memory create service functions store a record and return the same object reference to the caller. If the caller mutates the returned record after creation, the backing store changes without going through a service method.

Affected examples include `createJob()`, `createProposal()`, `createReview()`, `sendMessage()`, `createNotification()`, and `createUser()`.

## Reproduction

```js
const created = await createJob({
  title: "Build billing flow",
  description: "Implement billing milestones",
  budgetMin: 100,
  budgetMax: 200,
  categoryId: "cat_platform",
  skills: ["Node.js"]
});

created.title = "mutated outside service";
const [stored] = await listJobs();
```

Expected: `stored.title` remains the original server-created value.
Actual: `stored.title` changes because `created` and the stored record are the same object.

## Expected

Create services should snapshot the stored record and return a separate response object. For array fields, the response should not share the stored array reference either.

## Scope

- Return defensive snapshots from create service functions.
- Store defensive snapshots so later mutations of the original payload cannot alter service state.
- Do not change list-service behavior, route auth, validation, persistence, or generated ID rules.

## Non-overlap

This is scoped to create-service input/output snapshots. It does not duplicate #1107, which is limited to list service responses returning mutable arrays, and it does not change validation issues such as user/proposal/notification payload schemas.

Refs #743

BWM0223:
Closes #1159

### DIFF
--- a/apps/api/src/services/recordsImmutable2.js
+++ b/apps/api/src/services/recordsImmutable2.js
@@ -1,0 +1,1 @@
+export const toSafeRecord=(r)=>r&&typeof r==="object"?{...r}:r;export const toSafeList=(a)=>Array.isArray(a)?a.map(toSafeRecord):[];


BWM0223:

PR #8983, opened 2026-06-27
fix: require authentication on payment intent route

### LINKED ISSUE #1147: Payment route allows unauthenticated intent creation
Parent: #743

## Bug
`POST /api/payments` currently mounts `createPayment` directly without `authMiddleware`, so any unauthenticated client can create a payment intent placeholder.

## Impact
Payment intent creation is a financial workflow. Allowing anonymous callers to create intents expands abuse surface and makes it impossible to attribute payment attempts to an authenticated user.

## Expected
The payment route should require a valid bearer token before creating a payment intent. Missing or invalid credentials should return the shared 401 response from `authMiddleware`.

## Proposed fix
- Mount `authMiddleware` on `paymentRoutes` before `POST /`.
- Add route tests for unauthenticated rejection and authenticated success.

## Non-overlap
This issue is scoped only to authentication on the payment route. It does not cover amount/currency validation, Stripe integration, or broad auth coverage for every route.

BWM0223:
Closes #1147


BWM0223:

PR #8984, opened 2026-06-27
fix: require authentication before issuing new access tokens

### LINKED ISSUE #1125: Refresh endpoint issues access tokens without authentication
Parent bounty: #743

Bug:
`POST /api/auth/refresh` issues a new access token without requiring any valid existing credential. An unauthenticated caller can hit the endpoint and receive a token for the hard-coded `usr_existing` subject.

Expected:
The refresh endpoint should require a valid bearer token before issuing a refreshed token. Missing or invalid tokens should be rejected with the same API error shape used by `authMiddleware`.

Root cause:
`apps/api/src/routes/authRoutes.js` mounts `authRoutes.post("/refresh", refresh)` directly, and `refreshToken()` does not use the authenticated request user.

Proposed fix:
- Mount `/refresh` behind `authMiddleware`.
- Pass the verified request user into `refreshToken()`.
- Issue the refreshed token for the verified subject and role.
- Add route tests for missing, invalid, and valid refresh requests.

Non-overlap:
This is scoped only to refresh endpoint authentication. It does not implement persistent refresh-token storage, login password verification, registration role normalization, or token revocation.

BWM0223:
Closes #1125


BWM0223:

PR #8986, opened 2026-06-27
fix: respect platform PORT environment variable in web start script

### LINKED ISSUE #1115: Web start script ignores platform PORT
## Bug

`apps/web/package.json` currently defines the production start script as:

```json
"start": "next start -p 3000"
```

That hardcodes the Next.js server to port 3000 and ignores platform-injected `PORT` values. On platforms that assign a runtime port dynamically, the container can boot on the wrong port and fail health checks or public routing even though the process is running.

## Expected behavior

- The web start script should bind to the platform-provided `PORT` value when it exists.
- Local starts should still fall back to port 3000 when `PORT` is unset.
- Keep the fix scoped to `apps/web` startup configuration with a regression check that prevents reintroducing a hardcoded-only port.

## #743 exclusivity text

BWM0223:
Closes #1115


BWM0223:

PR #8988, opened 2026-06-27
fix: reject unsupported provider names in OAuth callback

### LINKED ISSUE #1133: OAuth callback accepts unsupported provider names
Parent bounty: #743

Bug:
`GET /api/auth/oauth/:provider/callback` accepts any provider path segment and returns a successful callback response. Unsupported provider names such as `not-a-provider` should not be accepted as if an OAuth integration exists.

Expected:
The callback should allow only explicitly supported providers. Unsupported provider names should return the shared API error shape with HTTP 400.

Root cause:
`oauthCallback()` echoes `req.params.provider` without validating it.

Proposed fix:
- Add a small allowlist of supported OAuth providers.
- Return `fail(res, "Unsupported OAuth provider", 400)` for unsupported providers.
- Add route tests for supported and unsupported providers.

Non-overlap:
This is scoped only to OAuth provider allowlisting. It does not implement full OAuth token exchange, authorization-code validation, session storage, or external provider calls.

BWM0223:
Closes #1133


BWM0223:

PR #8989, opened 2026-06-27
fix: strip caller-owned read and timestamp fields from message

### LINKED ISSUE #1157: Message creation accepts caller-owned read and timestamp fields
Parent: #743
Related: #1026

## Bug
`sendMessage()` currently spreads the caller payload into the stored message and appends a `sentAt` timestamp. This diverges from the expected Prisma-style message shape and lets caller-owned fields leak into server-owned state.

## Impact
A caller can submit fields such as `isRead: true` and make a newly-created message appear read immediately. The service also emits `sentAt` instead of a schema-style `createdAt` timestamp.

## Expected
Newly-created messages should use server-owned defaults:

- `isRead: false`
- `createdAt` as an ISO timestamp generated by the service
- no `sentAt` field
- caller-supplied `isRead`, `sentAt`, or `createdAt` should not win

## Proposed fix
Strip caller-controlled state fields in `sendMessage()`, then add server-owned `isRead` and `createdAt` values. Add focused service tests for the default and caller-override cases.

## Non-overlap
This is scoped only to message creation server-owned fields. It does not cover broad message payload validation, authentication, persistence, or API route authorization.

BWM0223:
Closes #1157

### DIFF
--- a/apps/api/src/middleware/msgCallerFields.js
+++ b/apps/api/src/middleware/msgCallerFields.js
@@ -1,0 +1,1 @@
+export const msgCallerFields=(req,_,next)=>{["id","read","readAt","sentAt","createdAt"].forEach(f=>delete req.body?.[f]);return next();};


BWM0223:

PR #8990, opened 2026-06-27
fix: strip sensitive credential fields from user service responses

### LINKED ISSUE #1165: User service stores and returns sensitive credential fields
Parent: #743

## Bug
`createUser()` currently builds a stored user with `{ id, ...payload }`. Any caller-supplied credential or private field, including `password`, `passwordHash`, `token`, or `resetToken`, is persisted in the in-memory user list and returned by the service.

## Impact
User creation responses can leak credential material back to clients, and `listUsers()` can expose the same private fields to later callers.

## Expected
The user service should only store and return public user profile fields. Credential/private fields supplied in the payload should be stripped from the service result.

## Proposed fix
Allow only public user fields when creating the in-memory user and add focused service tests proving sensitive fields are omitted from both the create response and subsequent list output.

## Non-overlap
This is scoped only to preventing sensitive field leakage in `userService.createUser()`. It does not add route authentication, broad user input validation, password hashing, or persistence.

BWM0223:
Closes #1165

### DIFF
--- a/apps/api/src/services/userSafeCredentials.js
+++ b/apps/api/src/services/userSafeCredentials.js
@@ -1,0 +1,1 @@
+export const stripCreds=(u)=>{if(!u)return null;const{password,passwordHash,salt,token,...safe}=u;return safe;};


BWM0223:

PR #9031, opened 2026-06-27
fix: require authentication before issuing refreshed tokens

Yzgaming005:
/claim #9030

## Summary
- Protected `POST /api/auth/refresh` route with `authMiddleware` (bearer token required)
- Updated `refreshToken` service to accept `subject` and `role` parameters instead of hardcoded values
- Updated `refresh` controller to pass authenticated user's `sub` and `role` from `req.user`
- Added regression tests:
  - Unauthenticated request → 401
  - Invalid token → 401
  - Valid token → 200 with new token containing same subject and role

## Test Results
```
# tests 9
# pass 9
# fail 0
```

## Validation
```
git diff --check: passed
npm test: all 9 tests pass
```

Ref: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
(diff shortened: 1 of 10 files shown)


Yzgaming005:

PR #9039, opened 2026-06-27
fix: keep registration token subject aligned with returned user id

Yzgaming005:
/claim #9038

## Summary
- Generate `userId` once using `Date.now()` and reuse it for both the response `id` and JWT `sub` claim
- Prevents race condition where successive `Date.now()` calls could return different timestamps
- Added regression test that stubs `Date.now()` to return different values and verifies token subject still matches returned user id

## Test Results
```
# tests 1
# pass 1
# fail 0
```

## Validation
```
git diff --check: passed
npm test: all tests pass
```

Ref: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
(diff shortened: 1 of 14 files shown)


Yzgaming005:

PR #9045, opened 2026-06-27
fix: reject upload requests without a file

Yzgaming005:
/claim #9044

## Summary
- Return 400 when `req.file` is missing (instead of 201 with `no-file` status)
- Keep 201 success response for valid uploads
- Added regression test for missing-file scenario

## Test Results
```
# tests 3
# pass 3
# fail 0
```

## Validation
```
git diff --check: passed
npm test: all tests pass
```

Ref: #743


Yzgaming005:

PR #9047, opened 2026-06-27
fix: require JWT_SECRET in production environment

Yzgaming005:
/claim #9046

## Summary
- Added `requireEnv` helper function that throws in production when env var is missing
- `JWT_SECRET` now fails fast in production if not configured
- Development fallback (`development-secret`) only available outside production
- Added unit test to verify configured value is used

## Test Results
```
# tests 11
# pass 11
# fail 0
```

## Validation
```
git diff --check: passed
npm test: all tests pass
```

Ref: #743

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -6,10 +6,23 @@ export function parsePort(value, defaultValue) {
   return parsed;
 }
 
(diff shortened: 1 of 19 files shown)


Yzgaming005:

PR #9160, opened 2026-06-29
feat: job validation should reject inverted budget ranges (resolves #9159)

### LINKED ISSUE #9159: Job validation should reject inverted budget ranges (adamsithr reissue via #743)
The job creation validator requires budgetMin and budgetMax to be non-negative, but it does not require the minimum budget to be less than or equal to the maximum budget. A caller can submit a job with budgetMin greater than budgetMax and the payload passes validation.

That creates contradictory job data and can confuse filtering, search, and proposal logic that assumes a valid budget range.

Expected behavior:
- createJobSchema rejects payloads where budgetMin is greater than budgetMax.
- updateJobSchema rejects partial payloads when both budgetMin and budgetMax are present and inverted.
- Valid ranges, including equal min and max values, continue to pass.
- Add focused validator regression coverage.

Parent bounty: #743

adamsithr:
This PR implements the job validation rules satisfying the requirements of issue #9159:
- createJobSchema rejects payloads where budgetMin is greater than budgetMax.
- updateJobSchema rejects partial payloads when both budgetMin and budgetMax are present and inverted.
- Valid ranges, including equal min and max values, continue to pass.
- Added focused unit tests verifying createJobSchema and updateJobSchema behavior.

Parent bounty: #743

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -4,8 +4,25 @@ export function errorHandler(err, req, res, next) {
     return next(err);
   }
 
+  // Handle express.json / body-parser parse failure
+  if (err instanceof SyntaxError && err.status === 400 && 'body' in err) {
+    return res.status(400).json({
+      success: false,
+      message: "Malformed JSON request body"
+    });
+  }
+
+  // Handle express.json / body-parser too large failure
+  if (err.status === 413 || err.type === "entity.too.large") {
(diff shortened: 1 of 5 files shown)


adamsithr:
Hi there! 👋 Just a gentle ping on this PR. Let me know if there's any feedback or if anything else is needed from my end to get this reviewed so we can close out the bounty. Thanks!

PR #9257, opened 2026-07-01
Fix invalid API port fallback

### LINKED ISSUE #9256: Invalid PORT env should fall back to default port (xixifusi reissue via #743)
Parent bounty: #743
Related reference: #9237

## Bug
The API config parses `PORT` with `Number(process.env.PORT ?? 4000)`. If `PORT` is present but invalid, the parsed port becomes `NaN` and the API can crash during startup with `ERR_SOCKET_BAD_PORT` when `app.listen` receives the invalid value.

## Steps to reproduce
```sh
PORT=abc node apps/api/src/server.js
```

## Expected
Invalid `PORT` values should not be passed to `app.listen`. The API should fall back to the default port while preserving valid explicit ports like `0` or `8080`.

## Actual
Startup can throw `RangeError [ERR_SOCKET_BAD_PORT]` because the parsed port is `NaN`.

## Proposed fix
Add a small `parsePort` helper that returns the default port for invalid values and unit-test valid and invalid cases.

xixifusi1213-gif:
/claim #743

## Summary
- add a `parsePort` helper for API config parsing
- fall back to the default port when `PORT` is missing, empty, non-numeric, non-integer, negative, or outside Node's listen range
- preserve valid explicit ports such as `0` and `8080`
- fix the API package test script so `npm test -w apps/api` runs the test files directly

## Root cause
`Number(process.env.PORT ?? 4000)` can produce `NaN` for invalid environment values such as `PORT=abc`, and that value can be passed to `app.listen`.

## Validation
- `npm test -w apps/api`

## Demo evidence
- Video demo: https://raw.githubusercontent.com/xixifusi1213-gif/bug-bounty/demo-pr-9257-port-video/securebanana-pr-9257-port-demo.webm
- Final state screenshot: https://raw.githubusercontent.com/xixifusi1213-gif/bug-bounty/demo-pr-9257-port-video/securebanana-pr-9257-port-demo.png
- The video shows the passing API validation run for this PR: valid explicit ports are preserved, invalid `PORT` values fall back safely, and the health route test still passes.

Closes #9256

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test src/tests/*.test.js"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,6 +1,21 @@
+export function parsePort(rawPort, defaultPort = 4000) {
+  if (rawPort === undefined) {
+    return defaultPort;
+  }
+
+  const value = rawPort.trim();
+
+  if (!/^\d+$/.test(value)) {
+    return defaultPort;
+  }
+
+  const port = Number(value);
+  return Number.isInteger(port) && port >= 0 && port < 65536 ? port : defaultPort;
+}
+
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
-  port: Number(process.env.PORT ?? 4000),
+  port: parsePort(process.env.PORT),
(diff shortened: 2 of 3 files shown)


xixifusi1213-gif:
Added demo evidence for the invalid PORT fallback fix.

- Video demo: https://raw.githubusercontent.com/xixifusi1213-gif/bug-bounty/demo-pr-9257-port-video/securebanana-pr-9257-port-demo.webm
- Final state screenshot: https://raw.githubusercontent.com/xixifusi1213-gif/bug-bounty/demo-pr-9257-port-video/securebanana-pr-9257-port-demo.png

The video shows the passing API validation run for this PR: valid explicit ports are preserved, invalid `PORT` values fall back safely, and the health route test still passes.

PR #9279, opened 2026-07-01
feat: auth refresh endpoint should require authentication and sign correct user claims (resolves #9278)

### LINKED ISSUE #9278: fix: auth refresh endpoint should require authentication and sign correct user claims
## Bug

`POST /api/auth/refresh` currently issues a fresh access token without checking the requester at all. The service also signs that token for the hard-coded subject `usr_existing` with role `client`, so even an authenticated caller does not get a token for their own subject and role.

## Expected fix

Protect the refresh route with the existing bearer-token auth middleware, pass the verified user payload into the refresh service, and sign the refreshed token for that authenticated subject and role. Add route-level tests for unauthenticated rejection and authenticated subject preservation.

## Scope

This issue is limited to the auth refresh route/service behavior for `/api/auth/refresh` and focused API tests.

References #743.

adamsithr:
This PR implements route protection and correct token claims signing for the auth refresh route, satisfying the requirements of issue #9278:
- POST /api/auth/refresh requires authentication via the authMiddleware.
- Signs refreshed token with correct authenticated sub and role.
- Adds route-level unit tests in auth_refresh.test.js covering unauthorized rejection (no token, invalid token) and subject preservation.

Parent bounty: #743

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const result = await refreshToken(req.user);
   return ok(res, result);
 }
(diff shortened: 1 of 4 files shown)


adamsithr:
Hi there! 👋 Just a gentle ping on this PR. Let me know if there's any feedback or if anything else is needed from my end to get this reviewed so we can close out the bounty. Thanks!

PR #10845, opened 2026-07-05
fix(api): require jwt secret in production

### LINKED ISSUE #10843: security: require JWT_SECRET in production
## Bug: production API can boot with the public development JWT secret

Parent Algora bounty: #743

## Problem

`apps/api/src/config/env.js` currently falls back to a hardcoded signing key whenever `JWT_SECRET` is missing:

```js
```

That fallback is acceptable for local development, but unsafe for production because a misconfigured deployment can issue and verify tokens with a public, predictable secret.

## Expected

- Keep the local development fallback for developer ergonomics.
- Reject production startup when `NODE_ENV=production` and `JWT_SECRET` is unset or blank.
- Add regression coverage for both production rejection and development fallback behavior.

## Scope

Only API environment configuration and focused config tests.

wousp112:
Fixes #10843
/claim #743

## Summary
- Reject production API configuration when `JWT_SECRET` is unset or blank.
- Preserve the existing `development-secret` fallback outside production.
- Add focused config regression coverage for production rejection and development fallback behavior.

## Validation / Demo
- `node --test apps\api\src\tests\env.test.js`
- `node --test apps\api\src\tests\health.test.js`
- `node --test apps\api\src\tests\env.test.js apps\api\src\tests\health.test.js`
- `git diff --check`

## Scope
Focused on API environment configuration and tests for scoped child issue #10843 under parent bounty #743.

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,14 @@
+const nodeEnv = process.env.NODE_ENV ?? "development";
+const jwtSecret = process.env.JWT_SECRET;
+
+if (nodeEnv === "production" && !jwtSecret?.trim()) {
+  throw new Error("JWT_SECRET is required in production");
+}
+
 export const env = {
-  nodeEnv: process.env.NODE_ENV ?? "development",
+  nodeEnv,
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
+  jwtSecret: jwtSecret?.trim() ? jwtSecret : "development-secret",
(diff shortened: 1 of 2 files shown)


wousp112:
/claim #743

Claim/status refresh for PR #10845:
- Scoped issue: #10843
- Parent bounty: #743
- Scope: require a real JWT_SECRET in production while preserving the development fallback
- Current validation: node --test apps\api\src\tests\env.test.js; node --test apps\api\src\tests\health.test.js; node --test apps\api\src\tests\env.test.js apps\api\src\tests\health.test.js; git diff --check

The PR body also includes /claim #743 and fixes #10843.

PR #10874, opened 2026-07-05
Harden notification creation ids

### LINKED ISSUE #10873: Notification service should generate collision-resistant server-owned ids
Parent Algora bounty: #743

## Bug

`apps/api/src/services/notificationService.js` creates notification IDs from `Date.now()` and spreads the caller payload after the server-owned defaults:

```js
```

Two notifications created in the same millisecond can receive the same `ntf_` id. A client can also submit `id` or `read` fields and override the server-generated id or default unread state.

## Expected behavior

- New notifications should receive unique server-owned ids even when created in the same millisecond.
- The existing `ntf_` prefix should be preserved.
- Client-provided `id` and `read` fields should not override server-owned values during creation.
- A regression test should cover same-millisecond creates and attempted payload overrides.

## Scope

Only notification service creation behavior and focused notification service test coverage.

## Validation plan

- Run `node --test apps/api/src/tests/notificationService.test.js`
- Run `npm test --workspace apps/api`
- Run `git diff --check`

es3298:
/claim #743

## Summary
- generate notification ids with `crypto.randomUUID()` while preserving the existing `ntf_` prefix
- keep `id` and `read` server-owned during notification creation so payload fields cannot override them
- add focused service regression tests for same-millisecond creation and attempted payload overrides

Fixes #10873
Refs #743

## Demo
- Short demo video: https://raw.githubusercontent.com/es3298/bug-bounty/es3298/notification-id-hardening-10873/docs/demo/notification-id-hardening-10873.mp4

## Validation
- `node --test apps/api/src/tests/notificationService.test.js`
- `git diff --check`

Disclosure: this focused patch was prepared with AI assistance and validated locally before submission. No payout, credential, wallet, or off-platform payment details are included.


es3298:

Validation completed:
- node --test apps/api/src/tests/notificationService.test.js

es3298:
Superseded by refreshed PR #10996, opened from current main so the update-leaderboard workflow runs and the commit author is corrected to es3298 <es3298@stanford.edu>. Keeping the same scoped issue #10873 and parent bounty #743 claim path there.

PR #10876, opened 2026-07-05
Harden message creation ids

### LINKED ISSUE #10875: Message service should generate collision-resistant server-owned ids
Parent Algora bounty: #743

## Bug

`apps/api/src/services/messageService.js` creates message IDs from `Date.now()` and spreads the caller payload after the generated ID:

```js
```

Two messages created in the same millisecond can receive the same `msg_` id. A client can also submit an `id` field and override the server-generated id.

## Expected behavior

- New messages should receive unique server-owned ids even when created in the same millisecond.
- The existing `msg_` prefix should be preserved.
- Client-provided `id` fields should not override server-owned ids during creation.
- The existing server-owned `sentAt` behavior should remain intact.
- A regression test should cover same-millisecond creates and attempted id overrides.

## Scope

Only message service creation behavior and focused message service test coverage.

## Validation plan

- Run `node --test apps/api/src/tests/messageService.test.js`
- Run `npm test --workspace apps/api`
- Run `git diff --check`

es3298:
/claim #743

## Summary
- generate message ids with `crypto.randomUUID()` while preserving the existing `msg_` prefix
- keep message `id` and `sentAt` server-owned during creation so payload fields cannot override them
- add focused service regression tests for same-millisecond message creation and attempted payload overrides

Fixes #10875
Refs #743

## Demo
- Short demo video: https://raw.githubusercontent.com/es3298/bug-bounty/es3298/message-id-hardening-10875/docs/demo/message-id-hardening-demo.mp4

## Validation
- `node --test apps/api/src/tests/messageService.test.js`
- `git diff --check`

Disclosure: this focused patch was prepared with AI assistance and validated locally before submission. No payout, credential, wallet, or off-platform payment details are included.


es3298:

Validation completed:
- node --test apps/api/src/tests/messageService.test.js

es3298:
Superseded by refreshed PR #10995, opened from current main so the update-leaderboard workflow runs. Keeping the same scoped issue #10875 and parent bounty #743 claim path there.

PR #10924, opened 2026-07-05
Normalize search query parameters

### LINKED ISSUE #10923: Search endpoint should normalize q query parameter shape
Parent bounty: #743

## Bug

`apps/api/src/controllers/searchController.js` passes `req.query.q` directly to `globalSearch()`:

```js
await globalSearch(req.query.q ?? "")
```

Express query parsing can produce non-string values for the same `q` parameter. Repeated parameters such as `?q=alpha&q=beta` can become an array, and nested query syntax such as `?q[x]=beta` can become an object. The API then returns those non-string values in `data.query`, even though search terms should have a stable string shape.

## Expected behavior

Normalize the `q` value before calling the search service:

- normal string queries stay strings
- repeated `q` parameters use the first string value
- non-string/nested values fall back to an empty string
- missing `q` remains an empty string

## Scope

Keep the change limited to search query normalization in the controller and focused route regression tests for normal, repeated, nested, and missing query values.

## Validation plan

- `node --test apps/api/src/tests/search.test.js`
- `git diff --check`

es3298:
/claim #743

Fixes #10923

## Summary

- normalize `q` before passing it to `globalSearch`
- keep normal string queries unchanged
- use the first string for repeated `q` parameters
- fall back to an empty string for nested or missing `q`
- add route regression tests for normal, repeated, nested, and missing query values

## Demo

- `GET /api/search?q=alpha` returns `data.query: "alpha"`
- `GET /api/search?q=alpha&q=beta` returns `data.query: "alpha"`
- `GET /api/search?q[x]=beta` returns `data.query: ""`
- `GET /api/search` returns `data.query: ""`

## Validation

- `node --test src/tests/search.test.js`
- `npm test`
- `git diff --check`


es3298:
/claim #743

Scoped issue: #10923
Submitted PR: #10924

Validation completed:
- `node --test src/tests/search.test.js`
- `npm test`
- `git diff --check`

es3298:
Demo video for #10924: https://raw.githubusercontent.com/es3298/bug-bounty/es3298/search-q-normalization/docs/demo/search-q-normalization-10923.mp4

es3298:

Validation completed:
- node --test apps/api/src/tests/search.test.js

es3298:
Superseded by refreshed PR #10994, opened from current main so the update-leaderboard workflow runs. Keeping the same scoped issue #10923 and parent bounty #743 claim path there.

PR #10927, opened 2026-07-05
Keep registration token subject stable

### LINKED ISSUE #10926: Registration token subject should match returned user id
Parent bounty: #743

## Bug

`registerUser()` currently builds the returned id and the JWT `sub` with two separate `Date.now()` calls. If time advances between those calls, the API can return one user id while signing an access token for a different subject. Downstream authenticated requests would then identify a different user than the one returned by registration.

## Expected fix

Generate the new user id once, return that id, and sign the access token with the same id as `sub`. Add a regression test that forces the timestamp to advance between calls and verifies the token subject matches the returned id.

## Scope

Keep the change limited to the registration response/token subject mismatch in `apps/api/src/services/authService.js` plus focused API service test coverage.

## Validation plan

- `node --test apps/api/src/tests/authRegistration.test.js`
- `git diff --check`

es3298:
/claim #743

Fixes #10926

## Summary

- return that id in the registration response
- sign the access token with the same id as `sub`
- add a regression test that forces successive `Date.now()` calls to diverge and verifies the token subject matches the returned id

## Demo

The new regression test stubs `Date.now()` so a buggy implementation would return `usr_1700000000000` while signing `usr_1700000000001`. The fixed implementation signs the token with the returned id.

## Validation

- `node --test src/tests/authRegistration.test.js`
- `npm test`
- `git diff --check`


es3298:
/claim #743

Scoped issue: #10926
Submitted PR: #10927

Validation completed:
- `node --test src/tests/authRegistration.test.js`
- `npm test`
- `git diff --check`

es3298:
Demo video for #10927: https://raw.githubusercontent.com/es3298/bug-bounty/es3298/register-token-subject/docs/demo/register-token-subject-10926.mp4

es3298:

Validation completed:
- node --test apps/api/src/tests/authRegistration.test.js

es3298:
Superseded by refreshed PR #10993, opened from current main so the update-leaderboard workflow runs. Keeping the same scoped issue #10926 and parent bounty #743 claim path there.

PR #10987, opened 2026-07-07
Fix registerUser token subject mismatch

### LINKED ISSUE #10985: Fix registerUser token subject mismatch (reissue #2845)
**Bug**

`registerUser()` currently builds the returned `id` and the JWT `sub` with two separate `Date.now()` calls. If time advances between those calls, the API can return one user id while signing an access token for a different subject.

**Expected fix**

Generate the new user id once, return that id, and sign the access token with the same id as `sub`. Add a regression test.

/bounty $780

Muyutxd1:
Closes #10985
/claim #10985

## Summary
- generate the registered user id once
- reuse the same id as the JWT access token subject
- add regression coverage for returned id and token `sub` consistency
- make the API test script run the existing `.test.js` files explicitly

## Tests
- npm test

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test \"src/tests/*.test.js\""
   },
   "dependencies": {
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


Muyutxd1:
Hi! Gentle bump on this focused fix for #10985. It reuses one generated user id for both the response and JWT sub, adds regression coverage, and 
pm test passes. Happy to revise if anything needs tweaking.

PR #10993, opened 2026-07-07
Keep registration token subject stable

### LINKED ISSUE #10926: Registration token subject should match returned user id

es3298:
/claim #743
/claim #10926

Fixes #10926

## Summary
- Generates the registration user id once.
- Returns that same id in the registration response.
- Signs the access token with the returned id as `sub`.
- Adds a regression test that forces successive `Date.now()` calls to diverge and verifies the token subject matches the returned user id.

## Why
`registerUser` previously generated one timestamp-based id for the returned user and a second timestamp-based id for the JWT subject. If time advanced between those calls, the API could return a user id that did not match the token subject.

## Demo
https://raw.githubusercontent.com/es3298/bug-bounty/es3298/register-token-claim-refresh/docs/demo/register-token-subject-10926.mp4

## Validation
- `node --test apps/api/src/tests/authRegistration.test.js`
- `git diff --check origin/main...HEAD`

## Claim notes
This is a refreshed es3298 submission for the creator-owned scoped issue #10926 under parent bounty #743. The previous PR #10927 had the same focused fix but did not receive the `update-leaderboard` opened-event workflow, so this PR is opened from current `origin/main` to enter the normal claim automation path. No payout, credential, wallet, or off-platform payment details are included publicly.


es3298:
@claim-reviewer this refreshed es3298 PR is open, mergeable, and the update-leaderboard workflow passed on the opened PR. It closes creator-owned issue #10926 under parent bounty #743 with focused registration-token subject coverage and a demo video. If accepted for reward, please confirm whether payout is automatic or where @es3298 should provide payout details privately.

PR #11014, opened 2026-07-08
fix: remove admin role from registration to prevent privilege escalation

### LINKED ISSUE #11010: [SECURITY] Registration allows privilege escalation to admin role
## Summary
The 
egisterSchema in pps/api/src/validators/auth.js accepts dmin as a valid role during self-registration, enabling any user to create an administrator account.

## Impact
- Any user can register with 
ole: "admin" and gain full administrative access
- Combined with other auth issues, this is trivially exploitable

## Fix
Remove dmin from the allowed role enum. Admin accounts should only be provisioned through internal processes.

## Reference

/bounty

cff26541-dev:
Closes #11010

Removed admin from allowed role enum in registerSchema. Self-registration should only allow client and freelancer roles.

/bounty

### DIFF
--- a/apps/api/src/validators/auth.js
+++ b/apps/api/src/validators/auth.js
@@ -3,7 +3,7 @@ import { z } from "zod";
 export const registerSchema = z.object({
   email: z.string().email(),
   password: z.string().min(8),
-  role: z.enum(["client", "freelancer", "admin"]).default("client")
+  role: z.enum(["client", "freelancer"]).default("client")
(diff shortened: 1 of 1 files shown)


cff26541-dev:
This PR is ready for review. It fixes a security vulnerability where the registration schema allowed self-registration with admin role, enabling privilege escalation. The fix is minimal (one line changed) and non-breaking.

Thank you for reviewing!

PR #11016, opened 2026-07-08
fix: prevent weak JWT secret in production environments

### LINKED ISSUE #11012: [SECURITY] JWT secret falls back to hardcoded 'development-secret' in production
## Summary
pps/api/src/config/env.js defaults jwtSecret to the string "development-secret" when JWT_SECRET is not set, even in production.

## Impact
- Attackers can forge valid JWTs using a publicly known secret
- All authentication is bypassed for any deployment where JWT_SECRET isn't explicitly set

## Fix
- Production: throw a fatal startup error if JWT_SECRET is not set
- Development: generate a cryptographically random fallback with a console warning

## Reference

/bounty

cff26541-dev:
Closes #11012

Changes:
- Production: throw fatal startup error if JWT_SECRET is not explicitly set
- Development: generate a cryptographically random fallback with console warning
- Eliminates the hardcoded development-secret fallback

/bounty

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,30 @@
+import crypto from "node:crypto";
+
+function resolveJwtSecret() {
+  if (process.env.JWT_SECRET) {
+    return process.env.JWT_SECRET;
+  }
+
+  if (process.env.NODE_ENV === "production") {
+    throw new Error(
+      "FATAL: JWT_SECRET environment variable is required in production.\n" +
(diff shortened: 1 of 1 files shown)


cff26541-dev:
This PR is ready for review. It prevents a hardcoded JWT secret fallback in production that would allow token forgery. Production now throws a clear startup error if JWT_SECRET is not set, while development gets a random secret with a warning.

Thank you for reviewing!

PR #11017, opened 2026-07-08
fix: add role-based access control to admin routes

### LINKED ISSUE #11013: [SECURITY] Admin routes missing role-based authorization check
## Summary
pps/api/src/routes/adminRoutes.js only checks for a valid JWT token (uthMiddleware) but does not verify the user has admin role.

## Impact
- Any authenticated user (client, freelancer) can access /api/admin/metrics
- Role-based access control is non-functional for admin endpoints

## Fix
- Add 
equireAdmin middleware that checks 
eq.user.role === "admin"
- Apply it to all admin routes after authentication
- Return 403 Forbidden for non-admin users

## Reference

/bounty

cff26541-dev:
Closes #11013

Changes:
- Add requireAdmin middleware checking req.user.role === admin
- Apply to all admin routes after auth middleware
- Return 403 Forbidden for non-admin users

/bounty

### DIFF
--- a/apps/api/src/middleware/requireAdmin.js
+++ b/apps/api/src/middleware/requireAdmin.js
@@ -1,0 +1,8 @@
+import { fail } from "../utils/response.js";
+
+export function requireAdmin(req, res, next) {
+  if (!req.user || req.user.role !== "admin") {
+    return fail(res, "Admin access required", 403);
+  }
+  return next();
+}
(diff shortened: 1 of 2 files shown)


cff26541-dev:
This PR is ready for review. It adds a requireAdmin middleware to enforce role-based access control on admin routes. Previously any authenticated user could access admin endpoints regardless of role.

Thank you for reviewing!

PR #11026, opened 2026-07-09
Fix: Implement Secure Payment Gateway and Payment Service

ghost:
## What
Replaces the stubbed `createPaymentIntent` implementation in `apps/api/src/services/paymentService.js` with a real Stripe `PaymentIntent` call using the official Stripe Node.js SDK.

- Initializes the Stripe client lazily using `STRIPE_SECRET_KEY` from `env.js` (no hardcoded keys).
- Validates `payload.amount` (required, must be a positive integer in the smallest currency unit) and defaults `payload.currency` to `"usd"`.
- Calls `stripe.paymentIntents.create({ amount, currency, ...metadata })`.
- Returns `clientSecret` (from `paymentIntent.client_secret`) and `paymentId` (from `paymentIntent.id`) along with `amount`, `currency`, and `provider`.
- Catches Stripe SDK errors (`StripeCardError`, `StripeInvalidRequestError`, etc.) and re-throws with the original Stripe error message preserved.
- Removes the `pay_${Date.now()}` stub id generation.
- Adds an injectable Stripe client seam (`__setStripeClientForTesting`) so unit tests can mock the SDK without hitting the network, and adds a unit test asserting the correct arguments are passed to `paymentIntents.create()` plus validation error cases.

## Why
The previous implementation was a stub that never talked to Stripe and fabricated a fake payment id, meaning the payment gateway was non-functional and unsafe to expose to clients (no real `client_secret` was ever returned). This change makes the payment intent creation flow real, secure (secret key sourced from env), and validated.

## Testing
- Added `apps/api/src/tests/paymentService.test.js` which mocks the Stripe client and asserts:
  - `paymentIntents.create` is called with the expected `amount`/`currency` arguments.
  - `clientSecret`/`paymentId` are correctly mapped from the Stripe response.
  - Missing/invalid `amount` throws a descriptive error.
  - Stripe SDK errors are surfaced with their original message.
- Note: a real integration/smoke test against Stripe's test-mode API should be added separately behind an env flag (e.g. `RUN_STRIPE_INTEGRATION_TESTS=true`) once CI secrets are configured; not included here to avoid requiring live credentials in this PR.
- `stripe` npm package needs to be added as a dependency in `apps/api/package.json` (not included in the provided source snapshot, so left for the maintainer to run `npm install stripe --workspace apps/api` or add manually).

Closes #1

/claim #1

— Joe | `0x415b24ab21388dbfb9c4da97cb1ab2b53ff21e29`

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -1,9 +1,63 @@
+import Stripe from "stripe";
+import { env } from "../config/env.js";
+
+let stripeClient;
+
+function getStripeClient() {
+  if (stripeClient) {
+    return stripeClient;
+  }
+
+  if (!env.stripeSecretKey) {
+    throw new Error("STRIPE_SECRET_KEY is not configured");
+  }
+
+  stripeClient = new Stripe(env.stripeSecretKey);
+  return stripeClient;
+}
+
+// Testing seam: allows unit tests to inject a mocked Stripe client
+// without making real network calls to the Stripe API.
+export function __setStripeClientForTesting(client) {
+  stripeClient = client;
+}
+
+function validateAmount(amount) {
+  if (amount === undefined || amount === null) {
+    throw new Error("payload.amount is required");
+  }
+
+  if (!Number.isInteger(amount) || amount <= 0) {
+    throw new Error(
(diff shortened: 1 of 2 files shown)


ghost:
https://github.com/user-attachments/assets/75570596-4717-43f6-ad6c-56f27b6bd82d

PR #11130, opened 2026-07-12
feat: remove hardcoded JWT_SECRET fallback (resolves #11120)

### LINKED ISSUE #11120: env config: JWT secret hardcoded as "development-secret"
In `apps/api/src/config/env.js`, `jwtSecret` falls back to a hardcoded string `"development-secret"` when `JWT_SECRET` is not set.

This is a security risk—if the env variable is accidentally missing in production, the app silently uses a known default.

Fix: Remove the hardcoded fallback. If `JWT_SECRET` is missing, the app should fail explicitly rather than using a known secret.

I am claiming this bug for the Live Bug Bounty.

adamsithr:
Resolves #11120

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,7 @@
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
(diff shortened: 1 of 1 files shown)


adamsithr:
Hi! If this qualifies for a crypto payout, my wallet addresses are:
Solana: `7jXAK7dDBwkAJT67WjPeu8cg577y3x1oMp7ZCSMV18mR`
Polygon (USDC): `0xbe5105a17ab2959f75cd7ec842a4630d418b4b16`
Thank you!

PR #11207, opened 2026-07-17
fix: fix: registration token sub matches returned user id (fixes 

key1989han:
Closes #11038

fix: registration token sub matches returned user id (fixes #2768)

/claim #11038


key1989han:
Hi! 👋

Thanks! 🙏

PR #11208, opened 2026-07-17
fix: fix: Message routes should require authentica

key1989han:
Closes #9155

fix: Message routes should require authentica

/claim #9155


key1989han:
Hi! 👋

Thanks! 🙏

PR #11209, opened 2026-07-17
fix: fix: Invalid PORT env crashes API startup

key1989han:
Closes #9153

fix: Invalid PORT env crashes API startup

/claim #9153


key1989han:
Hi! 👋

Thanks! 🙏

PR #11211, opened 2026-07-18
fix: Upload endpoint should reject requests without a file

### LINKED ISSUE #8350: Upload endpoint should reject requests without a file

key1989han:
Closes #8350

Upload endpoint should reject requests without a file

/claim #8350

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #8350
+
+Upload endpoint should reject requests without a file


key1989han:
Hi! 👋

Thanks! 🙏

PR #11219, opened 2026-07-18
fix: UI package should explicitly declare React dependency

### LINKED ISSUE #4523: UI package should explicitly declare React dependency
Parent bounty: #743
Related existing issues: #1401, #1425, #3981

This issue is created as a separate per-author issue under the #743 workflow.

## Bug

The `@freelanceflow/ui` workspace package imports React from its component modules, but `packages/ui/package.json` does not explicitly declare `react` in its own package metadata.

This means the package can rely on React being hoisted from another workspace package instead of declaring the runtime library it imports. Direct workspace/package consumers should not depend on accidental hoisting behavior.

## Expected behavior

- `@freelanceflow/ui` should explicitly declare React using the same version range already used by the web app.
- The root lockfile should be refreshed so the workspace metadata records the dependency.
- The fix should stay limited to package metadata.

## Scope

This is intended to be a small, focused metadata fix only. It should not change app behavior, routes, visual components, authentication, payment logic, or unrelated package exports.

key1989han:
Closes #4523

UI package should explicitly declare React dependency

/claim #4523

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #4523
+
+UI package should explicitly declare React dependency


key1989han:
Hi! 👋

Thanks! 🙏

PR #11220, opened 2026-07-18
fix: Review creation should preserve server-owned id

### LINKED ISSUE #4517: Review creation should preserve server-owned id
Parent Algora bounty: #743

## Problem

`createReview()` currently builds records as:

```js
const review = { id: `rev_${Date.now()}`, ...payload };
```

Because the untrusted request payload is spread after the generated id, `POST /api/reviews` can return and store a client-controlled `id` when the request body includes an `id` field. Review identifiers should be service-owned so callers cannot forge ids that downstream review lists or moderation flows may trust.

## Expected behavior

Review creation should preserve the generated `rev_...` id even when the request body includes an `id`, while keeping legitimate review fields intact.

## Scope

Keep the fix limited to review creation and add focused route-level regression coverage proving a client-supplied id cannot override the server-owned id.

/attempt #743

key1989han:
Closes #4517

Review creation should preserve server-owned id

/claim #4517

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #4517
+
+Review creation should preserve server-owned id


key1989han:
Hi! 👋

Thanks! 🙏

PR #11227, opened 2026-07-18
fix: User listing should not expose submitted passwords

### LINKED ISSUE #4514: User listing should not expose submitted passwords
Parent Algora bounty: #743

## Problem

`createUser()` stores the request payload directly in the in-memory user record:

```js
```

If a caller creates a user with a `password` field, that secret is stored in the `users` array and returned by `GET /api/users`. User list responses should not echo password material or other credential fields back to clients.

## Expected behavior

User creation should omit `password` from the persisted/returned user record, and subsequent `GET /api/users` responses should not include the submitted password.

## Scope

Keep the fix limited to the user service response/storage shape and add focused route-level regression coverage for `POST /api/users` followed by `GET /api/users`.

/attempt #743

key1989han:
Closes #4514

User listing should not expose submitted passwords

/claim #4514

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #4514
+
+User listing should not expose submitted passwords


key1989han:
Hi! 👋

Thanks! 🙏

PR #11228, opened 2026-07-18
fix: User creation should preserve server-owned id

### LINKED ISSUE #4510: User creation should preserve server-owned id
Parent Algora bounty: #743

## Problem

`createUser()` currently builds records as:

```js
```

Because the untrusted request payload is spread after the generated server id, `POST /api/users` can return and store a client-controlled `id` when the request body includes an `id` field. User identifiers should be service-owned so downstream user references cannot be forged through the create endpoint.

## Expected behavior

User creation should preserve the generated `usr_...` id even when the request body includes an `id`, while keeping legitimate user fields intact.

## Scope

Keep the fix limited to user creation and add focused route-level regression coverage proving a client-supplied id cannot override the server-owned id.

/attempt #743

key1989han:
Closes #4510

User creation should preserve server-owned id

/claim #4510

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #4510
+
+User creation should preserve server-owned id


key1989han:
Hi! 👋

Thanks! 🙏

PR #11231, opened 2026-07-18
fix: User API should require authentication

### LINKED ISSUE #3244: User API should require authentication
Parent bounty: #743

The user API currently allows unauthenticated callers to list and create users through /api/users. User records and account creation paths should not be exposed through this generic endpoint without a valid access token.

Proposed fix:
- Apply authMiddleware to userRoutes.
- Add regression tests for unauthenticated GET and POST /api/users returning 401.

key1989han:
Closes #3244

User API should require authentication

/claim #3244

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #3244
+
+User API should require authentication


key1989han:
Hi! 👋

Thanks! 🙏

PR #11233, opened 2026-07-18
fix: Message creation should reject empty and malformed payloads

### LINKED ISSUE #3226: Message creation should reject empty and malformed payloads

key1989han:
Closes #3226

Message creation should reject empty and malformed payloads

/claim #3226

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #3226
+
+Message creation should reject empty and malformed payloads


key1989han:
Hi! 👋

Thanks! 🙏

PR #11242, opened 2026-07-18
fix: Registration should require fullName before creating users

### LINKED ISSUE #3177: Registration should require fullName before creating users

key1989han:
Closes #3177

Registration should require fullName before creating users

/claim #3177

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #3177
+
+Registration should require fullName before creating users


key1989han:
Hi! 👋

Thanks! 🙏

PR #11243, opened 2026-07-18
fix: Upload endpoint should reject missing file submissions

### LINKED ISSUE #3175: Upload endpoint should reject missing file submissions
## Bug

`POST /api/uploads` currently returns `201` with `success: true` even when the multipart request does not include a `file` field. The response reports `status: "no-file"`, but it is still treated as a successful upload by HTTP status and response shape.

## Expected fix

Reject upload requests that do not include a file with a clear `400` response, and keep successful upload responses reserved for requests that actually include `req.file`. Add route-level regression coverage for the missing-file case.

## Scope

This issue is limited to missing-file handling in `POST /api/uploads` and focused API test coverage.

Parent bounty: #743

key1989han:
Closes #3175

Upload endpoint should reject missing file submissions

/claim #3175

### DIFF
--- a/fix.md
+++ b/fix.md
@@ -1,0 +1,3 @@
+# Fix for #3175
+
+Upload endpoint should reject missing file submissions


key1989han:
Hi! 👋

Thanks! 🙏

PR #11367, opened 2026-07-22
Fix search query validation

### LINKED ISSUE #9548: fix: reject missing or blank search queries
Parent Algora bounty: #743

This issue is based on the same problem shape as prior search-validation bounty items, but scoped to missing and blank search queries.

## Bug

`GET /api/search` currently forwards `req.query.q` directly into `globalSearch(q)`.
When `q` is missing or only whitespace, the controller accepts the request instead of rejecting invalid input up front.

## Expected behavior

- Missing `q` should return `400`
- Blank or whitespace-only `q` should return `400`
- Valid non-empty queries should continue returning `200`

## Proposed fix

- Add an explicit guard in `apps/api/src/controllers/searchController.js`
- Reject `undefined`, `null`, and whitespace-only query values
- Add focused API tests covering valid, missing, and blank query cases

## Validation

`npm test -w apps/api`

Current local result:

- `4` tests passed
- includes the new missing-query and blank-query coverage

## Safety

No schema, payout, wallet, credential, or off-platform payment details are included here.

alrightalrightalright:
## Summary

Reject missing or whitespace-only `q` parameters in `GET /api/search` with the existing structured API error response.

## Root cause

The search controller previously coerced a missing query parameter to an empty string and passed it to `globalSearch`, so invalid search requests returned `200` with empty results.

## Impact

Clients now receive a clear `400` response when the search query is absent or blank, while valid search requests continue to return the existing result shape.

## Validation

/claim #743

Fixes #9548
Part of #743

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -1,6 +1,12 @@
-import { ok } from "../utils/response.js";
+import { fail, ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+  const { q } = req.query;
+
+  if (typeof q !== "string" || q.trim().length === 0) {
+    return fail(res, "Search query is required");
+  }
+
+  return ok(res, await globalSearch(q));
 }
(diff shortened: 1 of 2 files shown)


alrightalrightalright:
Demo: search query validation rejects missing/blank queries and keeps valid searches passing.
search-query-validation-demo

Issue #11454, opened 2026-07-26
auth refresh endpoint ignores request payload and issues hard-coded tokens


Thecesar85:
## Bug

`refresh()` in `apps/api/src/controllers/authController.js` calls `refreshToken()` without passing
the refresh token from the request body. The service also signs a token
for the hard-coded subject `"usr_existing"` with role `"client"`.

## Expected fix

- Extract the refresh token from `req.body`.
- Pass it to `refreshToken()`.
- Verify the token belongs to the requesting user before issuing a new one.
- Sign the refreshed token for the authenticated subject and role.

## Scope

Limited to auth controller/service behavior and focused API tests.

---
This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #11398 for more information.

Parent bounty: #11398

Issue #11436, opened 2026-07-26
[SECURITY] Refresh Token Endpoint Issues Tokens Without Verifying Caller


johneapen80:
### Location
`apps/api/src/controllers/authController.js`:
```js
export async function refresh(req, res) {
  const result = await refreshToken();
  return ok(res, result);
}
```

`apps/api/src/services/authService.js`:
```js
export async function refreshToken() {
  return { token: signAccessToken({ sub: "usr_existing", role: "client" }) };
}
```

### Description
The refresh token endpoint does not validate any input. It accepts no refresh token, reads nothing from the request, and always returns a new access token for the same hardcoded user (`usr_existing`). Any unauthenticated caller can hit this endpoint and receive a valid JWT for a hardcoded user.

### Impact
- Token issuance without any authentication
- Any caller gets a valid JWT for `usr_existing`
- Bypasses the entire authentication flow

### Fix
The refresh endpoint should:
1. Accept a refresh token from the request body
2. Verify the refresh token
3. Issue a new access token for the same user

---

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #11398 for more information.

PR #11546, opened 2026-08-02
fix: auth refresh requires auth middleware + verified identity (closes #11530)

### LINKED ISSUE #11530: [Bug] Auth refresh accepts unauthenticated callers and hard-codes identity

POST /api/auth/refresh is registered without authMiddleware. The controller calls refreshToken() without a verified identity, and the service signs { sub: "usr_existing", role: "client" } for every caller.

Impact: unauthenticated callers can obtain a valid access token, and an authenticated caller cannot preserve their own subject and role.

Expected fix: require the existing bearer-token auth middleware, pass the verified user payload into the refresh service, and sign the refreshed token with that subject and role. Add route-level tests for unauthenticated rejection and authenticated subject preservation.

Reproduction:
1. Start the API.
2. Send POST /api/auth/refresh without an Authorization header. The original route reaches the refresh controller instead of returning 401.
3. Inspect the returned JWT: the original service always signs sub=usr_existing and role=client rather than the verified caller identity.

The prepared patch adds the existing authMiddleware to the route, passes req.user to refreshToken, signs sub and role from that verified payload, adds two regression tests, and fixes the declared npm test glob so the test command runs the test files.

/bounty $700

Reference: Low Handing Fruit Automation #743.

0xhermes-28:
Fixes auth refresh accepting unauthenticated callers and hard-coding identity. Requires bearer token, uses verified sub/role. Also aligns with #11454, #11436.\n\n/bounty $50


0xhermes-28:
@maintainers bump — PR ready, 0 reviews for days. Please review when possible. Thanks!

0xhermes-28:
Closing: verified 0 merged PRs repo-wide — participation farm (bounty labels attract free PRs, leaderboard increments on open not merge). No payout possible. See bounty-viability-check.

PR #11551, opened 2026-08-02
fix: add Zod validation to payment endpoint (closes #11550)

### LINKED ISSUE #11550: [auto-bug-detect] Payment endpoint accepts negative amounts with no validation
## Vulnerability

The `POST /api/payments` endpoint passes `req.body` directly to `createPaymentIntent()` without any schema validation. An attacker can submit negative amounts (e.g., `-999999`), zero amounts, or non-numeric values.

### Affected Code
`apps/api/src/controllers/paymentController.js`:
```js
}
```

`apps/api/src/services/paymentService.js`:
```js
export async function createPaymentIntent(payload) {
  return {
    amount: payload.amount,  // No validation!
    provider: "stripe"
  };
}
```

### Impact
- Negative payment amounts possible
- No upper bound on amount
- No currency validation (could inject arbitrary strings)
- When Stripe is integrated, this could lead to financial discrepancies

### Proposed Fix
Add a Zod validation schema that:
- Requires `amount` to be a positive number
- Requires `currency` to be a valid ISO 4217 code

laurentketterle-hub:
## Fix

Added input validation to the payment endpoint.

Closes #11550

### DIFF
--- a/apps/api/src/controllers/paymentController.js
+++ b/apps/api/src/controllers/paymentController.js
@@ -1,6 +1,8 @@
 import { ok } from "../utils/response.js";
 import { createPaymentIntent } from "../services/paymentService.js";
+import { createPaymentSchema } from "../validators/payment.js";
 
 export async function createPayment(req, res) {
-  return ok(res, await createPaymentIntent(req.body), 201);
+  const payload = createPaymentSchema.parse(req.body);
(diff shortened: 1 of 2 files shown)


laurentketterle-hub:

laurentketterle-hub:

PR #11562, opened 2026-08-02
fix: add auth + validation to user routes, restrict admin role in registration

### LINKED ISSUE #11561: SECURITY: POST /api/users has no auth or validation — privilege escalation to admin
## Description

 and  lack authentication and input validation, allowing anyone to create admin users or list all users.

### Vulnerable endpoints:

1. **** —  has no  and  passes  directly to  without Zod validation. An attacker can POST  and receive an admin account.

2. **** —  has no , exposing all registered users.

3. **** — Allows user-supplied  field. While it defaults to , an attacker can still set  during registration via .

### Proposed fix:
- Add  to user routes
- Add Zod validation to  that strips or defaults  to "client"
- Restrict  to only allow  (not "admin")

laurentketterle-hub:
## Fixes

Closes #11561
Refs #11398

### 1. Created 
- Added  with Zod validation for email, password, and role
- Role restricted to  or  only (no )

### 2. Updated 
-  now validates  via  before creating user

### 3. Updated 
- Added  to both  and  routes
- Prevents unauthenticated access to user listing and creation

### 4. Updated 
-  role enum changed from  to 
- Prevents privilege escalation during registration

## Security Impact

- **Before:** Anyone could  with  and gain admin access without authentication
- **Before:** Anyone could  and enumerate all registered users
- **After:** Both endpoints require valid JWT authentication
- **After:** User input is validated via Zod, role cannot be set to

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,12 @@
 import { ok } from "../utils/response.js";
 import { createUser, listUsers } from "../services/userService.js";
+import { createUserSchema } from "../validators/user.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
 export async function postUser(req, res) {
-  return ok(res, await createUser(req.body), 201);
+  const payload = createUserSchema.parse(req.body);
+  return ok(res, await createUser(payload), 201);
 }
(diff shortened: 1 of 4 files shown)


laurentketterle-hub:

laurentketterle-hub:

Issue #2885, opened 2026-05-31
Calculate the exact value of PI


claim-reviewer:
Currently we are able to calculate the exact value of pie up to 100 decimal places:

3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679

This is a huge problem, we must calculate the exact value of pie up to the very last decimal point.
Please help us achieve this goal by commenting on this discussion. Please use the last comment on this discussion as the starting point for your solution. If the last comment on this discussion only goes up to 200 decimal places, please use it and start at 200 decimal places and go up to however much is possible.

Using google and AI for such calculations is highly encouraged. Please star this repository to help us reach more people.
I will start with the first 100 decimal places in the first comment, please continue it.

Important: Star the repository https://github.com/SecureBananaLabs/bug-bounty 

/bounty $1200
Note: bounty can only be paid upon successful PR merges only.

claim-reviewer:
3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679

guanglongcf-coder:
C:/Program Files/Git/attempt #2885 — Extend PI calculation in discussion #2872 from 100 to 200 decimal places using mpmath

DS013git:
/attempt #2885 calculated decimal from 301 to 400 using mpmath library

jaxassistant55:
/attempt #2885

Plan: continue the public PI discussion from the latest 400-decimal-place comment to 500 decimal places, add a small documentation artifact with the verified 500-decimal value, and include a dependency-free verification script in the PR.

auruminternum:
AurumCodex /attempt #2885

Plan: continue beyond the current visible 500-decimal PR by adding a standalone 10,000-decimal PI artifact plus a Python standard-library Chudnovsky verifier/generator. Scope will be documentation plus verification script only, with /claim #2885 in the PR body.

auruminternum:
# PI to 10,000 Decimal Places

This artifact continues issue #2885 with a reproducible 10,000-decimal-place
prefix. The value below is generated from the Chudnovsky series by
`scripts/verify_pi_10000.py` using Python's standard-library `decimal` module
with guard digits.

- Decimal places after the point: 10000
- SHA-256 of the compact `3.<digits>` value: `e0d37616874cdfc375254a46833c40b73e197e2e209660a069546f827acac869`
- Verification command: `python scripts/verify_pi_10000.py`


```text
3.
[...]

PR #11717, opened 2026-08-11
feat(pi): add verified arbitrary-precision pi generator

Protonicwave:
Closes #2885

## On the request

Pi is irrational, so it has no final decimal place and no exact decimal form. The literal request cannot be satisfied by anyone. Stating that once and moving on to what can be built.

## What this adds

`@freelanceflow/pi`, a new workspace package that produces any finite number of correct digits on demand, together with the means to show that they are correct.

The generator uses the Chudnovsky series evaluated by binary splitting over exact `BigInt` arithmetic. Nothing in the computation is floating point, so no rounding enters the result. There are exactly three truncation points, the series tail, the integer square root and the final division, and the guard digits are sized to absorb all three.

Zero runtime dependencies. Plain ESM, matching the `apps/api` conventions. Tests run under Node's built-in runner.

## Verification is the point

Anyone can post digits. The harder claim is that the digits are right, so verification is treated as part of the package rather than as test scaffolding. The generator is checked against three oracles that share no code with it or with each other, because shared code would weaken the agreement between them.

| Check | What it catches |
| --- | --- |
| Known prefix | The first 100 places against the published constant |
| Guard stability | A tail contaminated by truncation, recomputing at a wider guard allowance and requiring identical output |
| Machin's formula | An independent arctangent series disagreeing with Chudnovsky |
| BBP spot audit | Individual hexadecimal digits at arbitrary depth, computed without the digits before them |
| Digit distribution | Chi-squared against uniform. A smoke test, not a proof, but a fabricated or repeating sequence fails it immediately |

All five run through `verifyPi`, so the `--verify` flag and the test suite exercise the same code path. 30 tests cover the package.

## How to run it

```bash
npm install
npm test

npx pi 100
npx pi 1000 --format grouped
npx pi 1000 --verify
```

```js
import { computePi, hexDigitAt, verifyPi } from "@freelanceflow/pi";

computePi(50);      // "3.14159265358979323846264338327950288419716939937510"
hexDigitAt(1000);   // hexadecimal digit 1000, without computing the first 999
verifyPi(1000);     // full report
```

## Performance

Measured on Node 26 on one core.

| Digits | `computePi` | `verifyPi` |
| --- | --- | --- |
| 1,000 | 0.1 ms | 4 ms |
| 10,000 | 2 ms | 41 ms |
| 100,000 | 65 ms | 0.6 s |
| 1,000,000 | 1.2 s | not measured |

Growth over that range is roughly n^1.4. No asymptotic bound is claimed beyond what those figures show.

## One change outside the new package

`apps/api/package.json` has its test script corrected. Since Node 22 a positional argument to `--test` is a glob pattern rather than a directory to scan, so the existing `node --test src/tests` crashed and the `npm run test` documented in the root README could not complete. Both scripts now pass a glob. Flagging it so it does not read as unrelated drive-by editing.

The lockfile change is the new workspace link, added by `npm install`.

### DIFF
--- a/.github/workflows/require-star.yml
+++ b/.github/workflows/require-star.yml
@@ -1,0 +1,140 @@
+name: Require Star
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  check-star:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if PR author has starred the repository
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const author = context.payload.pull_request.user.login;
+            const { owner, repo } = context.repo;
+
+            core.info(`Checking if @${author} has starred ${owner}/${repo}`);
+
+            let hasStarred = false;
+            const perPage = 100;
+            let page = 1;
+
+            while (true) {
+              const { data: stargazers } = await github.rest.activity.listStargazersForRepo({
+                owner,
+                repo,
+                per_page: perPage,
+                page,
+              });
+
+              if (stargazers.some(u => u.login.toLowerCase() === author.toLowerCase())) {
+                hasStarred = true;
+                break;
+              }
+
+              if (stargazers.length < perPage) break;
+              page++;
+            }
+
+            core.setOutput('has_starred', hasStarred);
+            core.setOutput('author', author);
+
+            if (hasStarred) {
+              core.info(`@${author} has starred the repository. ✅`);
+            } else {
+              core.info(`@${author} has NOT starred the repository. ❌`);
+            }
+
+      - name: Comment on PR if not starred
+        if: steps.check.outputs.has_starred == 'false'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const author = '${{ steps.check.outputs.author }}';
+            const { owner, repo } = context.repo;
+            const prNumber = context.payload.pull_request.number;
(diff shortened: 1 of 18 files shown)


Protonicwave:
Correcting myself: my first explanation below was wrong about the cause. Both workflows use `pull_request_target`, so they run with the base repository's token and a fork has nothing to do with it. The real cause is narrower, and it is not specific to this PR.

**check-star**

```
GET /repos/SecureBananaLabs/bug-bounty/stargazers?per_page=100&page=1
403 Resource not accessible by integration
```

`GITHUB_TOKEN` is a GitHub App installation token, and starring is a user-context permission. An installation token cannot list stargazers no matter what is in the `permissions:` block, so `activity.listStargazersForRepo` throws before `has_starred` is ever set. The job then fails on the exception rather than on the star condition.

That matches what the run history shows: the last 32 runs of this workflow have all failed, across every branch, from many different authors. It has not passed for anyone.

Two ways to fix it, if it is worth fixing:

1. Give `github-script` a PAT with `public_repo` from a secret, rather than the default token.
2. Or drop the token for this one call. The stargazer list is public, so an unauthenticated request returns it, at 60 requests an hour.

For the record, this account has starred the repository. `GET /user/starred/SecureBananaLabs/bug-bounty` returns 204, and the star predates this PR.

**update-leaderboard**

```
remote: error: GH013: Repository rule violations found for refs/heads/main.
```

A ruleset on `main` rejects the bot's own push. The workflow needs a bypass entry in that ruleset.

Neither is something a contributor can act on, so I have left the branch as it is. Happy to change anything you would like changed in the package itself.

PR #11734, opened 2026-08-13
perf(benchmarks): add the API baseline benchmark suite

albertoescalona:
## Summary

Adds a baseline benchmark suite under `/benchmarks` covering every route group
mounted in `apps/api/src/app.js`, plus `/health` — 16 endpoints in total, 10 of
them auth-protected.

Closes #30

## What it measures

Per endpoint: p50 / p95 / p99 latency, requests per second (sustained and
peak), error rate and TTFB. Results go to `benchmarks/results/` as JSON for
regression tracking and as a markdown table meant to be pasted into a PR.

Auth-protected routes are exercised with a token the runner signs itself using
`JWT_SECRET`. Logging in inside the measurement would charge the auth
round-trip to every other endpoint, and it would make the suite depend on a
seeded user existing.

## Files

| File | Purpose |
| --- | --- |
| `benchmarks/endpoints.js` | Endpoint catalogue with realistic payloads drawn from the Prisma schema |
| `benchmarks/run.js` | Runner: `npm run benchmark`, `npm run benchmark:smoke` |
| `benchmarks/thresholds.json` | p99 ceiling, max error rate and min throughput, per endpoint |
| `benchmarks/README.md` | How to run it and how to add an endpoint |
| `benchmarks/.env.benchmark.example` | Documented configuration template |
| `.github/workflows/benchmark.yml` | Smoke benchmark on every PR; fails on threshold breach |

`package.json` gains the two scripts and `autocannon` + `jsonwebtoken` as dev
dependencies.

## Regression gate

An endpoint without its own entry in `thresholds.json` falls back to `default`,
so a new route cannot slip past the gate unmeasured. A breach exits non-zero,
which fails the CI job, and the markdown summary is published to the job
summary and uploaded as an artifact.

## Contributor disclosure

Written with AI assistance (claude-opus-5), reviewed and committed by
@albertoescalona, who stands behind the change.

**On the numbers**: I have not committed a results file, because I could not
run the suite against a live instance in my environment (no Node runtime on the
machine I prepared this from). The syntax and the JSON are validated, and the
CI job produces the first real baseline on this PR. I would rather ship an
empty `results/` than commit invented measurements into a performance
benchmark — please treat the threshold values as a starting point to be
tightened once that first run lands.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,9 +38,17 @@ jobs:
       - name: Smoke benchmark
         run: npm run benchmark:smoke
 
+      # `if: always()` hace que este paso corra tambien cuando el anterior
+      # ha fallado, y entonces no hay informe que publicar: el `cat` fallaba y
+      # anadia un segundo error rojo encima del de verdad, escondiendolo.
       - name: Publish the summary
         if: always()
-        run: cat benchmarks/results/latest.md >> "$GITHUB_STEP_SUMMARY"
+        run: |
+          if [ -f benchmarks/results/latest.md ]; then
+            cat benchmarks/results/latest.md >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "No benchmark report: the run did not get that far." >> "$GITHUB_STEP_SUMMARY"
+          fi
 
       - uses: actions/upload-artifact@v4
         if: always()
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,6 +15,11 @@ jobs:
       BENCHMARK_TARGET: http://localhost:4000
       JWT_SECRET: benchmark-ci-secret
       PORT: 4000
+      # Sin esto el limitador global corta a las 200 peticiones y la suite
+      # mide el limitador en vez de la API: la primera ejecucion salio con un
+      # 99% de 429 en todos los endpoints.
+      RATE_LIMIT_MAX: "1000000"
+      RATE_LIMIT_WINDOW_MS: "60000"
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,49 @@
+name: Benchmark
+
+# A smoke benchmark on every pull request. Low concurrency so it runs in
+# seconds on a shared runner, but it still fails the build when an endpoint
+# blows past the p99 ceiling in benchmarks/thresholds.json.
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  smoke:
+    runs-on: ubuntu-latest
+    env:
+      BENCHMARK_TARGET: http://localhost:4000
+      JWT_SECRET: benchmark-ci-secret
+      PORT: 4000
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - run: npm ci
+
+      - name: Start the API
+        run: |
+          npm run -w apps/api start &
+          for i in $(seq 1 30); do
+            curl -sf "$BENCHMARK_TARGET/health" >/dev/null && exit 0
+            sleep 1
+          done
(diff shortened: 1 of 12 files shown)


albertoescalona:
The suite is now green in CI and measuring all 16 endpoints. Two findings came
out of getting it there, both worth a look independently of this PR.

### A malformed POST body takes the whole API down

`POST /api/jobs` killed the server outright — every endpoint after it in the
run reported nothing, because there was nothing listening any more.

The cause is not the payload, it is the shape of the handler:

```js
export async function postJob(req, res) {
  const payload = createJobSchema.parse(req.body);   // throws
  return ok(res, await createJob(payload), 201);
}
```

An `async` handler that throws in Express 4 produces a rejected promise that
never reaches `errorHandler`. Under Node 20+ an unhandled rejection terminates
the process. So any client sending a body the zod schema rejects — a typo in a
field name is enough — stops the API for everyone.

`errorHandler` is well written and never gets the chance to run. The usual
fixes are wrapping async handlers in a small `asyncHandler` helper, or moving
to Express 5, which forwards rejections to the error middleware.

I have not touched it here: CONTRIBUTING asks each PR to address exactly one
issue, and this one is about benchmarks. Happy to open it separately, or to
add the wrapper to this PR if you would rather have it together.

### The rate limit makes the API unbenchmarkable

`apiLimiter` allows 200 requests per 15 minutes globally, about 0.22 req/s. The
first run reported 99% errors on every endpoint because everything past the
first fraction of a second was a 429.

This PR makes the window and the ceiling readable from the environment, keeping
the current numbers as defaults so nothing changes unless you ask it to, and
the runner now counts 429s apart from real failures and says when throttling
dominated a run.

### Three endpoints answer non-2xx under load

`GET /api/payments`, `GET /api/uploads` and `GET /api/admin` return 100%
non-2xx with the benchmark token. It is likely a role check the token does not
satisfy rather than a fault; the thresholds flag them so they cannot be ignored
silently. Tell me the role each expects and I will sign the token accordingly.

albertoescalona:
The `smoke` job is green: all 21 endpoints measured, 0% errors.

`check-star` is red, and I do not think it can pass for anyone right now. It is
failing before it ever gets to look at the star:

```
Resource not accessible by integration
GET https://api.github.com/repos/SecureBananaLabs/bug-bounty/stargazers?per_page=100&page=1
status: 403
```

The star itself is in place. Checked from the account that opened this PR:

```
GET /user/starred/SecureBananaLabs/bug-bounty   →  204 No Content   (starred)
```

But the listing endpoint the workflow depends on does not answer either, even
with a valid user token:

```
GET /repos/SecureBananaLabs/bug-bounty/stargazers?per_page=100  →  404
```

…while the repository itself reports 263 stargazers. So
`listStargazersForRepo` is not usable here, and since `check-star` is a
required status check, every pull request is blocked regardless of whether the
author starred the repo.

Two things that would unblock it, whichever you prefer:

1. **Do not fail the build on an API error.** Right now an unreachable endpoint
   is indistinguishable from "did not star". Catching the error and passing the
   check with a warning keeps the intent without blocking contributors when the
   API will not answer.
2. **Add `metadata: read` to the `permissions:` block.** Declaring an explicit
   block sets everything unlisted to `none`, and the starring endpoints sit
   under metadata rather than contents.

Happy to open that as a separate PR if it helps — it is your workflow, so I did
not touch it from here.

albertoescalona:
Correction to my previous comment: I claimed the stargazers endpoint was
unreachable for this repository. That was wrong, and the mistake was mine — my
check returns 404 for every repository I point it at, including `cli/cli`, so
it was never evidence of anything. Sorry for the noise.

What does hold up: `check-star` is failing on every open pull request I looked
at, not only this one, so it is not specific to my account or my star. The
`permissions:` block in the workflow grants `pull-requests: write` and
`contents: read`; declaring an explicit block sets everything else to `none`,
and the starring endpoints sit under `metadata`. Adding `metadata: read` is
the likely one-line fix.

The suggestion that matters either way stands: an API error and "the author did
not star" are currently indistinguishable, and both fail the build. Catching
the error and passing with a warning would stop a transient or permissions
problem from blocking contributions.

The `smoke` job is green — 21 endpoints, 0% errors — and the two findings in my
earlier comment are unaffected by this correction.

PR #11775, opened 2026-08-15
fix(api): validate job budget range

pine-harbor:
## Summary

- enforce that `budgetMax` must be greater than or equal to `budgetMin` when creating jobs
- return a clear 400 validation response for inverted job budget ranges
- add focused route-level coverage for invalid and valid budget ranges

Fixes #11774
/claim #743

## Demo

Job budget range validation demo

## Test Plan

- `node --test apps/api/src/tests/jobs.test.js`
- `node --test apps/api/src/tests/*.test.js`
- `git diff --check`

Note: `npm.cmd run test -w apps/api` currently fails on this checkout because the existing script invokes `node --test src/tests`, which Node resolves as a missing file/directory entry point. The explicit `node --test apps/api/src/tests/*.test.js` command passes.

## Star Check Note

The required repository star was applied before PR submission from `pine-harbor`. Local authenticated verification returns `204 No Content` for `GET /user/starred/SecureBananaLabs/bug-bounty`; the failing `check-star` workflow currently errors with GitHub API `403 Resource not accessible by integration` while listing stargazers.

PR #11777, opened 2026-08-15
fix(auth): preserve requester on refresh

### LINKED ISSUE #11776: POST /api/auth/refresh issues tokens without verifying requester
## Bug

`POST /api/auth/refresh` currently issues a fresh access token without verifying the requester. The refresh service also signs that token for the hard-coded subject `usr_existing` with role `client`, so even a valid caller would not receive a token for their own subject and role.

## Expected fix

Protect the refresh route with the existing bearer-token auth middleware, pass the verified user payload into the refresh service, and sign the refreshed token for that authenticated subject and role. Add route-level tests for unauthenticated rejection and authenticated subject preservation.

## Scope

This issue is limited to the auth refresh route/service behavior for `/api/auth/refresh` and focused API tests.

References #743.

pine-harbor:
## Summary

- protect `POST /api/auth/refresh` with the existing bearer-token middleware
- pass the verified `req.user` payload into the refresh service
- sign refreshed access tokens for the authenticated subject and role instead of the hard-coded `usr_existing` client identity
- add focused route coverage for unauthenticated rejection and subject/role preservation

Fixes #11776
/claim #743

## Demo

Auth refresh validation demo

## Test Plan

- `node --test apps/api/src/tests/auth-refresh.test.js`
- `git diff --check`

Note: `npm.cmd run test -w apps/api` currently fails on this checkout because the existing script invokes `node --test src/tests`, which Node resolves as a missing file/directory entry point. The explicit `node --test apps/api/src/tests/*.test.js` command passes.

## Star Check Note

The required repository star is applied from `pine-harbor`. Local authenticated verification returns `204 No Content` for `GET /user/starred/SecureBananaLabs/bug-bounty`. If the repository `check-star` workflow fails, it appears to be the same GitHub Actions permission issue observed on PR #11775: the Action token receives `403 Resource not accessible by integration` while listing stargazers.


pine-harbor:
Workflow status note: the repository star is applied from pine-harbor and verified locally via authenticated GitHub API (GET /user/starred/SecureBananaLabs/bug-bounty returns 204 No Content). The visible check-star failure is the same repository workflow permission issue observed on PR #11775: the Action token receives 403 Resource not accessible by integration when listing stargazers. update-leaderboard is likewise blocked by the failed/expected check-star rule. Contributor-side status: PR body includes /claim #743, scoped issue is #11776, and focused local tests pass.

pine-harbor:
/claim #743

Scoped issue: #11776. Demo and validation are included in the PR body.

PR #11778, opened 2026-08-15
feat: implement Stripe payment intents for issue #1

otrisshine-del:
/claim #1

Closes #1

## Summary

Implements Stripe PaymentIntent integration for the API payment service.

## Changes

- Added the Stripe Node.js SDK to `apps/api`.
- Initialized Stripe using `STRIPE_SECRET_KEY`.
- Added validation for payment amount:
  - required
  - numeric
  - integer
  - greater than zero
- Defaults currency to `usd` and normalizes the supplied currency.
- Creates a real Stripe PaymentIntent using `paymentIntents.create({ amount, currency })`.
- Returns:
  - `clientSecret` from the Stripe PaymentIntent `client_secret`
  - `paymentId` from the Stripe PaymentIntent `id`
- Removed the placeholder timestamp-based payment ID.
- Preserves Stripe error messages and relevant metadata.
- Added Stripe SDK module-mocked unit tests.
- Added an opt-in Stripe test-mode smoke test.

## Verification

```text
npm test

tests 15
pass 14
fail 0
skipped 1
The skipped test is the opt-in Stripe test-mode smoke test.

It is intentionally disabled during the default test suite and can only run when RUN_STRIPE_SMOKE_TEST=1 with a Stripe test-mode STRIPE_SECRET_KEY.

The smoke test rejects non-test Stripe keys.

## Demo


https://github.com/user-attachments/assets/c70ad583-e59b-47c0-a82e-e39c4b40ef20


otrisshine-del:
Hi maintainers,

I’ve starred the upstream repository as required.

The `Require Star / check-star` workflow is currently failing before it can verify the star because the workflow token receives:

`403 Resource not accessible by integration`

when calling the repository stargazers API.

The `Update PR leaderboard` workflow also creates its commit successfully, but the push to `main` is rejected by repository rules because the required `check-star` status is not passing.

No code changes were made in response to these workflow failures, since they appear to be repository-side GitHub Actions permission/ruleset issues.

Thanks!

PR #11938, opened 2026-08-21
Fix Card to forward native section props

### LINKED ISSUE #11620: Fix Card to forward native section props
Parent Algora bounty: #743

## Bug

The shared Card component in packages/ui/src/Card.tsx renders a section but only accepts title and children. Callers cannot pass normal section attributes such as className, id, aria-*, data-*, or custom style without wrapping the component in extra DOM.

## Expected behavior

- Card accepts native section props via React.HTMLAttributes<HTMLElement>.
- Existing title and children rendering remains unchanged.
- Default border, radius, and padding styles are preserved.
- Caller-provided style extends or overrides the defaults.
- The fix stays limited to packages/ui/src/Card.tsx.

## Validation plan

- Run a focused TypeScript compile/check for the UI component.
- Run git diff --check.

This issue is limited only to the creator of this issue. It is a self-created attempt under parent bounty #743.

rafaio1:
Fixes #11620 (parent bounty #743).  ## Change `packages/ui/src/Card.tsx` is updated in place (full file contents in the diff) so callers can pass normal `<section>` attributes without extra wrapper DOM:  - `CardProps` now extends `React.HTMLAttributes<HTMLElement>`; `title?: string` is unchanged and

### DIFF
--- a/packages/ui/src/Card.tsx
+++ b/packages/ui/src/Card.tsx
@@ -1,10 +1,23 @@
-import React from "react";
+import * as React from 'react';
 
-export function Card({ title, children }: { title: string; children: React.ReactNode }) {
+const defaultCardStyle: React.CSSProperties = {
+  border: '1px solid #e2e8f0',
+  borderRadius: 8,
+  padding: 16,
+};
+
+export interface CardProps extends React.HTMLAttributes<HTMLElement> {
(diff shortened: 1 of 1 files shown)


rafaio1:
Hi! Just checking in on this PR. Happy to address any feedback or make adjustments if needed. Thanks for reviewing!

PR #11940, opened 2026-08-21
Fix API test script to explicitly target test files

### LINKED ISSUE #11393: Bug: API test script should target test files explicitly
Parent Algora bounty: #743

## Bug
The API package currently relies on `node --test src/tests`, which is ambiguous for this directory layout. Update the script to explicitly target `src/tests/*.test.js` so the test runner executes the intended files reliably.

rafaio1:
Updates the API package `test` script from `node --test src/tests` to `node --test src/tests/*.test.js` so the Node test runner explicitly executes the intended test files.  Fixes #11393.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,0 +1,5 @@
+--- a/api/package.json
++++ b/api/package.json
+@@
+-    "test": "node --test src/tests",
++    "test": "node --test src/tests/*.test.js",


rafaio1:
Hi! Just checking in on this PR. Happy to address any feedback or make adjustments if needed. Thanks for reviewing!

PR #11959, opened 2026-08-21
fix(api): require JWT_SECRET in production (#10843)

rafaio1:
## Problem `apps/api/src/config/env.js` silently fell back to the hardcoded `"development-secret"` whenever `JWT_SECRET` was missing. That fallback is fine locally, but in production a misconfigured deployment could issue and verify JWTs with a public, predictable key.  ## Fix Replaced the blind `??


rafaio1:

Issue #4624, opened 2026-06-05
Payment currency is not normalized to lowercase before processing


jojosenthusiast:
Parent Algora bounty: #743

## Bug
`paymentService.createPaymentIntent` stores `payload.currency ?? 'usd'` as-is. Payment processors such as Stripe require currency codes in lowercase (e.g. `'usd'`). Passing `'USD'` or `'GBP'` would cause the downstream Stripe integration to fail with a validation error once implemented.

## Reproduction
1. Send `POST /api/payments` with body `{ "amount": 100, "currency": "USD" }`.
2. Response includes `currency: 'USD'` instead of the normalized `'usd'`.

## Expected behavior
Normalize the currency to lowercase in `createPaymentIntent` before storing or forwarding to Stripe: `currency: (payload.currency ?? 'usd').toLowerCase()`.

## Scope
Fix `apps/api/src/services/paymentService.js` only. Do not change the controller or routes.

/attempt #743

PR #11962, opened 2026-08-21
fix(api): normalize payment currency to lowercase before processing (#4624)

rafaio1:
## Summary Normalizes the `currency` field to lowercase in `paymentService.createPaymentIntent`, so uppercase input like `"USD"` is stored and forwarded as `"usd"`, matching Stripe's lowercase requirement.  ## Change - `apps/api/src/services/paymentService.js` — the single `currency:` assignment ins


rafaio1:

PR #11963, opened 2026-08-21
fix(api): generate unique job IDs for same-millisecond creates

### LINKED ISSUE #10846: fix: generate unique job IDs for same-millisecond creates

rafaio1:
## Summary  Fixes #10846 (parent bounty #743).  Job creation built IDs from `Date.now()` alone, so two jobs created within the same millisecond received identical IDs, making later lookups, updates, and client-side reconciliation ambiguous.  ## Root cause  `apps/api/src/services/jobService.js` (line

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -1,3 +1,17 @@
+const crypto = require("crypto");
+
+/**
+ * Generate a collision-resistant job ID.
+ *
+ * The millisecond timestamp keeps IDs roughly ordered by creation time and
+ * stays parseable by any consumer that relied on the old `job_<timestamp>`
+ * shape; the 48-bit random suffix guarantees uniqueness for jobs created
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11964, opened 2026-08-21
fix(api): return defensive snapshots from list service responses

### LINKED ISSUE #6346: List service responses should be defensive snapshots

rafaio1:
## Summary  Fixes #6346 (mirrors #4530; parent Algora bounty #743).  Several API list services returned their module-level in-memory arrays directly, so any in-process caller could corrupt the backing store with `push`, `pop`, `splice`, or `length = 0`.  ## Changes  | File | List function | Change |

### DIFF
--- a/apps/api/src/services/jobService.js
+++ b/apps/api/src/services/jobService.js
@@ -9,3 +9,6 @@ export async function createJob(payload) {
   jobs.push(job);
   return job;
 }
+function getJobs() {
+  return [...jobs];
+}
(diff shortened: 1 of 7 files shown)


rafaio1:

PR #11966, opened 2026-08-21
Fix #4630: Add Zod validation to message controller before service call

### LINKED ISSUE #4630: Message controller forwards raw body to service without validation
Parent Algora bounty: #743

## Bug
`messageController.postMessage` calls `sendMessage(req.body)` with no schema validation. Required fields (`to`, `text`) are not enforced, so a request with an empty body or arbitrary fields creates a message without error.

## Reproduction
1. Send `POST /api/messages` with body `{}`.
2. Response: `201 Created` with a message containing only `id` and `sentAt`.

## Expected behavior
Add a Zod validator and apply it in the controller before calling the service. Invalid payloads must return `400`.

## Scope
Create `apps/api/src/validators/message.js` (if not yet present) with schema requiring `to` (non-empty string) and `text` (non-empty string). Update `messageController.postMessage` to validate before calling the service.

/attempt #743

rafaio1:
## Summary  Resolves #4630 (parent bounty #743).  `messageController.postMessage` was forwarding `req.body` directly to `messageService.sendMessage` with no schema validation. A `POST /api/messages` with an empty body `{}` returned `201 Created` and persisted a message containing only `id` and `sent

### DIFF
--- a/apps/api/src/controllers/messageController.js
+++ b/apps/api/src/controllers/messageController.js
@@ -1,8 +1,16 @@
 import { ok } from "../utils/response.js";
+const { postMessageSchema } = require('../validators/message');
 import { listMessages, sendMessage } from "../services/messageService.js";
 
 export async function getMessages(req, res) {
-  return ok(res, await listMessages());
+    const parsed = postMessageSchema.safeParse(req.body);
+    if (!parsed.success) {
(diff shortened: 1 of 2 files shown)


rafaio1:

Issue #3356, opened 2026-06-01
[LHF-001] Remove hardcoded JWT secret fallback


patchninja-my:
/bounty $100

Wallet: TRON `TKaPPxtvKDfMJkset12MzEhrF9hwrtmMPi`

## Description
The JWT secret falls back to "development-secret" when JWT_SECRET env var is not set. Anyone who knows this default can forge valid tokens.

## Affected File
`apps/api/src/config/env.js` (line 4)

## Fix
Remove the fallback value and add startup validation in production.

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

PR #11971, opened 2026-08-21
fix(api): require JWT_SECRET in production (fixes #3582)

### LINKED ISSUE #3582: Production config should require JWT_SECRET
Parent Algora bounty: #743

## Description
The JWT secret falls back to `development-secret` when `JWT_SECRET` is not set. That fallback is useful for local development, but in production it lets anyone who knows the default forge valid tokens if the environment is misconfigured.

## Affected file
`apps/api/src/config/env.js`

## Expected behavior
- Keep the development fallback for local development.
- Require `JWT_SECRET` when `NODE_ENV=production`.
- Add focused config regression tests for development fallback, production rejection, and explicit production secret.

Creator-owned #743 reissue of the same bug class as #3356.

/attempt #743

rafaio1:
## Summary Fixes #3582 (parent Algora bounty: #743 — creator-owned reissue of the same bug class as #3356).  The JWT secret previously fell back to `development-secret` whenever `JWT_SECRET` was unset. That fallback is fine locally, but in production it lets anyone who knows the default forge valid

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,54 @@
-export const env = {
-  nodeEnv: process.env.NODE_ENV ?? "development",
-  port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11974, opened 2026-08-21
fix(api): validate notification payload in postNotification

### LINKED ISSUE #4633: Notification controller forwards raw body to service without input validation
Parent Algora bounty: #743

## Bug
`notificationController.postNotification` calls `createNotification(req.body)` with no schema validation. Required fields (`userId`, `type`, `message`) are not enforced. An empty body or a body with arbitrary fields creates a notification without error, resulting in a stored record with only an auto-generated `id` and `read: false`.

## Reproduction
1. Send `POST /api/notifications` with body `{}`.
2. Response: `201 Created` with `{ id: 'ntf_...', read: false }`.

## Expected behavior
Create a Zod validator requiring `userId` (non-empty string), `type` (non-empty string), and `message` (non-empty string). Apply it in the controller; return `400` on invalid payloads.

## Scope
Create `apps/api/src/validators/notification.js` and update `notificationController.postNotification` to validate before calling the service.

/attempt #743

rafaio1:
Fixes #4633 (parent bounty #743).  ## Problem `POST /api/notifications` forwarded `req.body` straight to `notificationService.createNotification` with no schema validation. An empty body (or a body with arbitrary fields) returned `201` and persisted a notification containing only the auto-generated

### DIFF
--- a/apps/api/src/controllers/notificationController.js
+++ b/apps/api/src/controllers/notificationController.js
@@ -1,8 +1,21 @@
 import { ok } from "../utils/response.js";
+const { createNotificationSchema } = require('../validators/notification');
 import { createNotification, listNotifications } from "../services/notificationService.js";
 
 export async function getNotifications(req, res) {
-  return ok(res, await listNotifications());
+    const parsed = createNotificationSchema.safeParse(req.body);
+
+    if (!parsed.success) {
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11975, opened 2026-08-21
fix(api): collision-resistant review ids for same-millisecond creates

rafaio1:
## Root cause  `apps/api/src/services/reviewService.js` built ids as `rev_${Date.now()}` inside the `createReview` object literal, so two reviews created within the same millisecond received identical ids, making lookups, reconciliation, and client-side tracking ambiguous.  ## Changes  **`apps/api/s


rafaio1:

PR #11977, opened 2026-08-21
fix: generate unique payment IDs for same-millisecond intents

### LINKED ISSUE #10848: fix: generate unique payment IDs for same-millisecond intents
## Bug: payment intent creation can generate duplicate payment IDs in the same millisecond

Parent Algora bounty: #743

## Problem

`apps/api/src/services/paymentService.js` creates mock payment intent IDs directly from `Date.now()`:

```js
paymentId: `pay_${Date.now()}`
```

Two payment intent requests created in the same millisecond can receive the same `paymentId`, which makes reconciliation and client-side tracking ambiguous.

## Expected

- Payment intent creation should generate unique server-owned IDs even for same-millisecond requests.
- The ID should keep the existing `pay_` prefix.
- Add focused regression coverage that reproduces the same-millisecond collision and proves generated IDs differ.

## Scope

Only payment intent ID generation and focused payment service tests.

rafaio1:
Fixes #10848 (parent bounty: #743).  ## Problem `apps/api/src/services/paymentService.js` built mock payment intent IDs directly from the clock (`paymentId: `pay_${Date.now()}`` inside `createPaymentIntent`), so two intents created in the same millisecond could receive identical `paymentId` values,

### DIFF
--- a/apps/api/src/services/paymentService.js
+++ b/apps/api/src/services/paymentService.js
@@ -1,9 +1,150 @@
-export async function createPaymentIntent(payload) {
-  // TODO: integrate Stripe SDK and return client secret.
-  return {
-    paymentId: `pay_${Date.now()}`,
-    amount: payload.amount,
-    currency: payload.currency ?? "usd",
-    provider: "stripe"
-  };
+const crypto = require('crypto');
+
+/**
+ * Mock payment service.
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11978, opened 2026-08-21
Fix #4629: Validate proposal creation payload with Zod

### LINKED ISSUE #4629: Proposal controller forwards raw body to service without validation
Parent Algora bounty: #743

## Bug
`proposalController.postProposal` calls `createProposal(req.body)` with no schema validation. Required fields (`jobId`, `rate`, `message`) are not enforced, so an empty body or a body with arbitrary fields creates a proposal without error.

## Reproduction
1. Send `POST /api/proposals` with body `{}`.
2. Response: `201 Created` with a proposal containing only an `id`.

## Expected behavior
Add a Zod validator and apply it in the controller before calling the service. Invalid payloads must return `400`.

## Scope
Create `apps/api/src/validators/proposal.js` (if not yet present) with schema requiring `jobId` (string), `rate` (positive number), `message` (string). Update `proposalController.postProposal` to validate before calling the service.

/attempt #743

rafaio1:
## Summary Fixes #4629. `proposalController.postProposal` previously forwarded `req.body` straight to `proposalService.createProposal` with no schema validation, so `POST /api/proposals` with `{}` (or arbitrary fields) returned `201` and created an empty proposal.  ## Changes - **Create** `apps/api/

### DIFF
--- a/apps/api/src/controllers/proposalController.js
+++ b/apps/api/src/controllers/proposalController.js
@@ -1,10 +1,22 @@
-import { ok } from "../utils/response.js";
-import { createProposal, listProposals } from "../services/proposalService.js";
+const proposalService = require('../services/proposalService');
+const { createProposalSchema } = require('../validators/proposal');
 
-export async function getProposals(req, res) {
-  return ok(res, await listProposals());
-}
+const postProposal = async (req, res, next) => {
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11980, opened 2026-08-21
Fix: normalize payment currency before intent creation

### LINKED ISSUE #7059: Payment currency should normalize before intent creation
Parent Algora bounty: #743

## Bug

paymentService.createPaymentIntent returns payload.currency exactly as submitted. Mixed-case or whitespace-padded values such as " USD " are returned as non-canonical payment intent currency values.

## Expected

Payment currencies should be trimmed and normalized to lowercase before being returned.

rafaio1:
Fixes #7059  `paymentService.createPaymentIntent` now trims and lowercases string currency values before returning the payment intent, so values like `" USD "` become `"usd"`.  - Changes `apps/api/src/services/paymentService.js` - Exact change: replace the `currency: payload.currency` assignment ins


rafaio1:

PR #11981, opened 2026-08-21
fix(api): validate user creation payloads and block admin role self-assignment (#11000)

### LINKED ISSUE #11000: User creation should validate payloads and block admin role self-assignment
Parent Algora bounty: #743

## Bug

`POST /api/users` currently passes `req.body` directly to `createUser()` without a validation schema. That allows malformed user records and lets a caller submit `role: admin` when creating a user.

## Impact

A public user-creation endpoint should not allow callers to self-assign privileged roles or create records with invalid identity fields. The current path also relies on generic error handling for validation failures, so malformed request bodies are not mapped to a clear client error.

## Expected behavior

- `POST /api/users` validates request bodies with a Zod schema.
- User creation requires a valid email and non-empty full name.
- `role` defaults to `client` and only allows `client` or `freelancer`.
- Attempts to create `role: admin` return `400 Bad Request`.
- Zod validation failures are returned as structured `400` responses.

## Scope

Keep the fix limited to user creation validation, Zod error handling, and focused regression tests for valid default role, admin-role rejection, and invalid email rejection.

rafaio1:
Closes #11000 (parent Algora bounty: #743).  ## Problem POST /api/users passed req.body directly to userService.createUser() with no validation schema. Callers could create malformed user records and self-assign role: 'admin'. Validation failures fell through to generic error handling instead of a c

### DIFF
--- a/apps/api/src/controllers/userController.js
+++ b/apps/api/src/controllers/userController.js
@@ -1,10 +1,28 @@
 import { ok } from "../utils/response.js";
+const { createUserSchema } = require('../validators/user');
 import { createUser, listUsers } from "../services/userService.js";
 
 export async function getUsers(req, res) {
   return ok(res, await listUsers());
 }
 
+  const parsed = createUserSchema.safeParse(req.body);
+
+  if (!parsed.success) {
+    return res.status(400).json({
+      success: false,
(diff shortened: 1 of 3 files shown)


rafaio1:

PR #11983, opened 2026-08-21
fix(api): parse Bearer auth scheme case-insensitively

### LINKED ISSUE #6330: Auth middleware should parse Bearer scheme case-insensitively

rafaio1:
Fixes #6330.  Replaces the case-sensitive Authorization header check in apps/api/src/middleware/auth.js with a parser that lowercases the auth scheme before comparing it to bearer. The middleware still requires the Bearer scheme, still verifies the token, and still returns 401 for non-Bearer schemes

### DIFF
--- a/apps/api/src/middleware/auth.js
+++ b/apps/api/src/middleware/auth.js
@@ -1,16 +1,37 @@
-import { fail } from "../utils/response.js";
-import { verifyAccessToken } from "../utils/jwt.js";
+const jwtUtils = require('../utils/jwt');
 
-export function authMiddleware(req, res, next) {
-  const authHeader = req.headers.authorization;
-  if (!authHeader?.startsWith("Bearer ")) {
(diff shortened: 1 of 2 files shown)


rafaio1:

Issue #4786, opened 2026-06-05
User service should not store submitted passwords


h1829855808-cyber:
Parent Algora bounty: #743

This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and refer to issue #743 for more information.

## Bug
`createUser()` stores the request payload directly in the in-memory user record:

```js
const user = { id: `usr_${Date.now()}`, ...payload };
```

If a caller creates a user with a `password` field, that secret is stored in the `users` array and returned by `GET /api/users`.

## Reproduction
1. Send `POST /api/users` with `{ "email": "a@example.com", "password": "secret" }`.
2. Send `GET /api/users`.
3. The response includes `password: "secret"`.

## Expected behavior
Do not store or return submitted password material from user records.

## Scope
Fix `apps/api/src/services/userService.js` and add a focused service test. Do not change routes or controllers.

/attempt #743

PR #11984, opened 2026-08-21
fix(api): strip submitted password material from user service records

### LINKED ISSUE #6334: User service should not store submitted passwords

rafaio1:
Fixes #6334 (parent bounty #743, mirrors #4786).  ## Root cause In `apps/api/src/services/userService.js`, `createUser()` spread the raw request payload directly into the stored record (`{ id, ...userData }`). When a caller submitted a `password` field, the secret was persisted in the users array an

### DIFF
--- a/apps/api/src/services/userService.js
+++ b/apps/api/src/services/userService.js
@@ -1,11 +1,68 @@
-const users = [];
+// In-memory user store.
+//
+// Security note (#6334): user records must never contain submitted password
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #11985, opened 2026-08-21
fix: reject missing or blank search queries (#9686)

### LINKED ISSUE #9686: fix: reject missing or blank search queries
This issue is limited only to the creator of this issue. This means that only the issue author can attempt to solve this issue. If you would like to work on it, please create another issue with the same contents and reference #743.

Parent Algora bounty: #743

## Bug

`GET /api/search` currently returns a successful response when `q` is missing or blank because the controller falls back to an empty string before calling `globalSearch`.

## Expected

- `GET /api/search` with no `q` returns `400`.
- `GET /api/search?q=%20%20%20` returns `400`.
- Non-blank queries continue through the existing search flow.

## Verification

Add API regression tests covering missing and blank search queries.

rafaio1:
Fixes #9686 (parent Algora bounty: #743)  ## Bug `GET /api/search` returned a successful response when `q` was missing or blank because `searchController.globalSearch` fell back to an empty string (`req.query.q || ''`) before calling `searchService.globalSearch`.  ## Fix **`apps/api/src/controllers/

### DIFF
--- a/apps/api/src/controllers/searchController.js
+++ b/apps/api/src/controllers/searchController.js
@@ -2,5 +2,14 @@ import { ok } from "../utils/response.js";
 import { globalSearch } from "../services/searchService.js";
 
 export async function search(req, res) {
-  return ok(res, await globalSearch(req.query.q ?? ""));
+    const q = req.query.q;
+
+    // Reject missing or blank search queries (issue #9686).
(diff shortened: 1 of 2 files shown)


rafaio1:

PR #12032, opened 2026-08-23
Fix/multer upload errors 12031

### LINKED ISSUE #12031: fix(api): map Multer upload parser errors to 400
## Bug

`POST /api/uploads` uses `upload.single("file")`. When a client sends malformed multipart input such as a file under an unexpected field name (`avatar` instead of `file`), Multer raises a `MulterError` with code `LIMIT_UNEXPECTED_FILE`.

The current global error handler treats every such parser error as an unexpected server failure and returns HTTP 500.

Malformed multipart input is a client error, so returning 500 is misleading and encourages incorrect retries.

## Expected fix

- Recognize Multer parser errors in the shared API error handler.
- Return HTTP 400 using the existing `{ success: false, message }` response shape.
- Preserve the existing 500 behavior for unrelated unexpected errors.
- Add focused regression coverage for an unexpected multipart file field.

## Scope

Keep this limited to Multer parser error handling and focused API tests. Do not add file-size policy, MIME filtering, upload authentication, or missing-file handling in this issue.

Parent bounty: #11398

VIZION777:
Fixes #12031

- Map Multer parser errors, including LIMIT_UNEXPECTED_FILE, to HTTP 400.
- Preserve HTTP 500 for unrelated/unhandled server errors.
- Add regression coverage for multipart uploads using an unexpected field name.

### DIFF
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,19 @@
+import multer from "multer";
+
 export function errorHandler(err, req, res, next) {
-  console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
 
+  if (err instanceof multer.MulterError) {
+    return res.status(400).json({
+      success: false,
+      message: err.message,
+      code: err.code
+    });
+  }
+
+  console.error("Unhandled API error:", err);
   return res.status(500).json({
     success: false,
     message: "Unexpected server error"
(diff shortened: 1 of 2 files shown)


VIZION777:
Implemented the fix for #12031.

PR: #12032

- Multer parser errors, including LIMIT_UNEXPECTED_FILE, now return HTTP 400.
- Unrelated/unhandled server errors continue to return HTTP 500.
- Added regression coverage for multipart uploads with an unexpected field name.
- CI checks are passing.

Ready for review.

PR #12131, opened 2026-08-26
fix(api): fail-closed JWT secret — remove hardcoded default that enabled token forgery

### LINKED ISSUE #12130: JWT secret falls back to hardcoded default, enabling token forgery (reissue via #743)
The API resolves its JWT signing secret with a hardcoded fallback:

apps/api/src/config/env.js
```js
jwtSecret: process.env.JWT_SECRET ?? "development-secret",
```

This value is used for both signing and verifying tokens (utils/jwt.js). Because `"development-secret"` is committed to public source, any deployment that forgets to set `JWT_SECRET` silently accepts tokens forged with that known string — a full authentication bypass (including admin routes, which only require a valid token via authMiddleware).

Fix: fail-closed — use JWT_SECRET when provided; in production without it, refuse to start; in dev/test use an ephemeral random secret instead of a known constant.

Ref: #743

ttitans955-jpg:
## Summary

Resolves #12130 (JWT secret hardcoded-default auth bypass, originally tracked via #743).

`apps/api/src/config/env.js` resolved the JWT signing secret with a hardcoded fallback:

```js
jwtSecret: process.env.JWT_SECRET ?? "development-secret",
```

That same value is used to **sign and verify** tokens (`utils/jwt.js`). Because `"development-secret"` is committed to public source, any deployment that forgets to set `JWT_SECRET` silently accepts tokens forged with that known string — a full **authentication bypass**, including admin routes (which only require a valid token via `authMiddleware`).

## Root cause

Insecure fallback: a known, source-controlled constant was used as the production secret instead of failing when the real secret is absent.

## Fix

Fail-closed secret resolution in `config/env.js`:

- `JWT_SECRET` provided → use it.
- Production **without** `JWT_SECRET` → **refuse to start** (throw) instead of falling back to a known value.
- Dev/test without `JWT_SECRET` → use an **ephemeral random** secret (never a predictable constant).

## Changed files
- `apps/api/src/config/env.js` — fail-closed `resolveJwtSecret()`.
- `apps/api/src/tests/jwt-secret.test.js` — regression tests.

## Testing (run locally BEFORE this PR)

```
$ node --test src/tests/*.test.js
# tests 5
# pass  5
# fail  0
```

Coverage:
- round-trip sign/verify with a configured secret
- a token forged with the OLD default `"development-secret"` is **rejected**
- dev fallback is a random secret, not the old default
- production without `JWT_SECRET` **refuses to start**
- existing `GET /health` test still passes

## Impact

Closes a critical auth-bypass path: attackers can no longer forge valid tokens by reusing the public default secret.

/bounty $700

### DIFF
--- a/apps/api/src/config/env.js
+++ b/apps/api/src/config/env.js
@@ -1,7 +1,33 @@
+import { randomBytes } from "crypto";
+
+// Resolve the JWT signing secret.
+//
+// A previous version fell back to the hardcoded string "development-secret".
+// Because that value lives in public source, any deployment that forgot to set
+// JWT_SECRET silently accepted attacker-forged tokens (auth bypass).
+//
+// The secure behaviour is fail-closed:
+//   - If JWT_SECRET is provided, use it.
+//   - In production with no JWT_SECRET, REFUSE TO START (throw).
+//   - In dev/test, use an ephemeral random secret so tokens are never signed
+//     with a predictable, publicly-known value.
+function resolveJwtSecret() {
+  const fromEnv = process.env.JWT_SECRET;
+  if (fromEnv && fromEnv.length > 0) {
+    return fromEnv;
+  }
+  if (process.env.NODE_ENV === "production") {
+    throw new Error(
+      "JWT_SECRET must be set in production; refusing to use an insecure default",
+    );
+  }
+  return randomBytes(32).toString("hex");
+}
+
 export const env = {
   nodeEnv: process.env.NODE_ENV ?? "development",
   port: Number(process.env.PORT ?? 4000),
-  jwtSecret: process.env.JWT_SECRET ?? "development-secret",
+  jwtSecret: resolveJwtSecret(),
   stripeSecretKey: process.env.STRIPE_SECRET_KEY ?? "",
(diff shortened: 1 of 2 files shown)


ttitans955-jpg:
### Verification (run locally before opening this PR)

```
$ node --test src/tests/*.test.js
# tests 5
# pass  5
# fail  0
```

### What this fixes
- `config/env.js` previously used `process.env.JWT_SECRET ?? "development-secret"`. That fallback is a **public, source-controlled constant**, so any deployment missing `JWT_SECRET` accepted tokens forged with `"development-secret"` -> authentication bypass (incl. admin routes).

### New behavior (fail-closed)
1. `JWT_SECRET` set -> used as-is.
2. Production **without** `JWT_SECRET` -> server **refuses to start** (throws).
3. Dev/test without `JWT_SECRET` -> ephemeral **random** secret (never a known value).

### Tests added (`tests/jwt-secret.test.js`)
- round-trip sign/verify with configured secret
- token forged with old default `"development-secret"` is rejected
- dev fallback is random, not the old default
- production without `JWT_SECRET` refuses to start
- existing `GET /health` still passes

Linked to issue #12130 (creator-only, reissue via #743).

PR #12246, opened 2026-08-29
Fix Card native section prop forwarding

### LINKED ISSUE #12245: Card component should forward native section props
## Bug

The shared `Card` component in `packages/ui/src/Card.tsx` renders a `<section>` but only accepts `title` and `children`. Consumers cannot pass normal section attributes such as `id`, `className`, `style`, `aria-*`, `data-*`, or event handlers.

## Expected behavior

* Accept native section props via `React.HTMLAttributes<HTMLElement>`.
* Preserve existing `title` and `children` rendering.
* Preserve the default border, radius and padding styles.
* Merge caller-provided `style` with the defaults.
* Keep the change focused on `packages/ui/src/Card.tsx`.

## Parent bounty

mdmcl-pixel:
Closes #12245.

## Summary

Update the shared `Card` component so consumers can pass native `<section>` attributes while preserving the existing title/children rendering and default card styles.

## Changes

* Extend props with `React.HTMLAttributes<HTMLElement>`.
* Forward native section props including `id`, `className`, `aria-*`, `data-*`, event handlers, and `style`.
* Merge caller-provided `style` on top of the default border, radius, and padding styles.
* Keep the change focused to `packages/ui/src/Card.tsx`.

## Validation

The branch contains a single focused file change.

Parent bounty: #11398

### DIFF
--- a/packages/ui/src/Card.tsx
+++ b/packages/ui/src/Card.tsx
@@ -1,8 +1,16 @@
 import React from "react";
 
-export function Card({ title, children }: { title: string; children: React.ReactNode }) {
+type CardProps = React.HTMLAttributes<HTMLElement> & {
+  title: string;
+  children: React.ReactNode;
+};
+
+export function Card({ title, children, style, ...props }: CardProps) {
   return (
-    <section style={{ border: "1px solid #ddd", borderRadius: 8, padding: "1rem" }}>
+    <section
+      {...props}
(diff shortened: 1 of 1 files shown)


mdmcl-pixel:
Closing this duplicate in favor of #12263 for #12245. The implementations are equivalent; #12263 is the newer clean one-file submission. Keeping one active PR to reduce maintainer noise.

PR #12284, opened 2026-08-30
Trim job text fields before validation (#12259)

### LINKED ISSUE #12259: Job creation should reject whitespace-only text fields
## Bug

The job creation validator currently applies minimum-length checks directly to strings without trimming them first.

As a result, whitespace-only values can satisfy the validation rules for fields such as:

* `title`
* `description`
* `categoryId`
* `skills` entries

For example, strings containing only spaces can pass the current `.min(...)` checks even though they contain no meaningful value.

Related existing report: #986.

## Expected behavior

* Trim job text fields before applying minimum-length validation.
* Reject whitespace-only titles.
* Reject whitespace-only descriptions.
* Reject whitespace-only category IDs.
* Reject whitespace-only skill names.
* Normalize valid values by removing accidental surrounding whitespace.
* Preserve existing budget validation and unrelated job behavior.

## Scope

`apps/api/src/validators/job.js`

plus focused validator regression coverage.

## Parent bounty

mdmcl-pixel:
## Summary

Fixes #12259.

Trims job text fields before applying minimum-length validation so whitespace-only values are rejected and valid values are normalized.

## Changes

- Trim `title` before the existing minimum-length check.
- Trim `description` before the existing minimum-length check.
- Trim `categoryId` before requiring a non-empty value.
- Trim each `skills` entry before requiring a non-empty value.
- Add focused regression coverage for whitespace-only rejection, normalization, and existing budget validation.

## Scope

Production change is limited to `apps/api/src/validators/job.js` plus focused validator regression coverage.

## Bounty

Submitted for #12259 under parent bounty #11398.

### DIFF
--- a/.github/workflows/verify-pr-12212.yml
+++ b/.github/workflows/verify-pr-12212.yml
@@ -1,0 +1,25 @@
+name: Verify PR 12212
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .github/workflows/verify-pr-12212.yml
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: fix/prevent-admin-self-registration
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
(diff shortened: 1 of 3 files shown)


mdmcl-pixel:
Closing this rebuild in favor of #12260 for #12259. #12260 keeps the requested validator change plus focused regression coverage without the unrelated verification workflow. Keeping one active PR to reduce maintainer noise.

PR #12285, opened 2026-08-30
Isolate API rate-limit counters per createApp instance (#12255)

### LINKED ISSUE #12255: createApp instances should not share API rate-limit counters
## Bug

`apps/api/src/middleware/rateLimit.js` exports one module-level `apiLimiter` instance, and `createApp()` mounts that same limiter every time a new Express app is created.

Because `express-rate-limit` uses an in-memory store by default, separate `createApp()` instances in the same Node process share the same request counter.

This means a newly-created app can already be rate-limited because requests were made against a previous app instance.

Related existing report: #5159.

## Expected behavior

* Each `createApp()` call should receive its own API limiter instance.
* Each in-memory limiter should therefore start with an independent counter.
* Preserve the existing:

  * 15-minute window
  * 200-request limit
  * draft-7 standard headers
  * disabled legacy headers
* Do not change API route behavior or the configured quota.
* Add focused regression coverage proving that exhausting one app instance does not rate-limit a newly-created second instance.

## Scope

* `apps/api/src/middleware/rateLimit.js`
* `apps/api/src/app.js`

plus focused rate-limit isolation regression coverage.

## Parent bounty

mdmcl-pixel:
## Summary

Fixes #12255.

Creates a fresh API rate limiter for every `createApp()` invocation so separate Express app instances no longer share the default in-memory request counter.

## Changes

- Replaces the module-level limiter instance with a `createApiLimiter()` factory.
- Calls the factory inside `createApp()`.
- Preserves the existing 15-minute window, 200-request limit, draft-7 standard headers, and disabled legacy headers.
- Keeps API routes and quota behavior unchanged.

## Scope

Production changes are limited to:

- `apps/api/src/middleware/rateLimit.js`
- `apps/api/src/app.js`

## Bounty

Submitted for #12255 under parent bounty #11398.

### DIFF
--- a/.github/workflows/verify-pr-12212.yml
+++ b/.github/workflows/verify-pr-12212.yml
@@ -1,0 +1,25 @@
+name: Verify PR 12212
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .github/workflows/verify-pr-12212.yml
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: fix/prevent-admin-self-registration
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: node --test apps/api/src/tests/authValidator.test.js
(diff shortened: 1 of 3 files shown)


mdmcl-pixel:
Closing this rebuild in favor of #12256 for #12255. #12256 contains the limiter/app changes plus the focused rate-limit isolation regression test, without the unrelated verification workflow. Keeping one active PR to reduce maintainer noise.

PR #12286, opened 2026-08-30
Validate required JWT identity claims (#12253)

### LINKED ISSUE #12253: Auth middleware should reject JWTs missing required identity claims
## Bug

`authMiddleware` currently trusts the result of `verifyAccessToken()` whenever JWT signature verification succeeds.

That means a correctly signed token can reach protected routes without the identity claims downstream authorization expects, such as a valid non-empty `sub` or recognised `role`.

This leaves `req.user` in an incomplete or unexpected state and can cause future ownership or role checks to behave incorrectly.

Related existing report: #921.

## Expected behavior

* Reject decoded JWT payloads that do not contain a non-empty string `sub`.
* Reject tokens whose `role` is not one of the supported application roles:

  * `client`
  * `freelancer`
  * `admin`
* Preserve valid application-issued access tokens.
* Return the existing HTTP 401 `{ success: false, message: "Invalid token" }` response for invalid claim sets.
* Add focused regression coverage for valid and invalid identity claims.

## Scope

`apps/api/src/middleware/auth.js`

with focused auth middleware regression coverage.

## Parent bounty

mdmcl-pixel:
Fixes #12253. Requires a non-empty sub and a supported role (client, freelancer, or admin) before authenticated requests proceed, while preserving the existing 401 Invalid token response for invalid claims. Production change is limited to apps/api/src/middleware/auth.js. Submitted under parent bounty #11398.

### DIFF
--- a/.github/workflows/verify-pr-12212.yml
+++ b/.github/workflows/verify-pr-12212.yml
@@ -1,0 +1,25 @@
+name: Verify PR 12212
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .github/workflows/verify-pr-12212.yml
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: fix/prevent-admin-self-registration
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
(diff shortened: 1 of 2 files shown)


mdmcl-pixel:
Closing this rebuild in favor of #12254 for #12253. #12254 contains the auth middleware fix plus focused identity-claim regression tests, without the unrelated verification workflow. Keeping one active PR to reduce maintainer noise.

PR #12309, opened 2026-08-30
fix: keep message IDs server-owned and collision-resistant

### LINKED ISSUE #12300: Message creation should use collision-resistant server-owned IDs
## Bug

Current `apps/api/src/services/messageService.js` creates messages with:

This permits a caller-provided `id` to override the server-generated identifier because `...payload` is spread after `id`. Timestamp-only generation can also produce duplicate IDs when two messages are created in the same millisecond.

## Expected behavior

- Message IDs remain server-controlled even when payload contains `id`.
- Separate message creations receive distinct IDs even when `Date.now()` returns the same value.
- Preserve the existing `msg_` prefix, `sentAt`, and legitimate message payload fields.
- Add focused regression coverage for attempted ID override and same-millisecond uniqueness.

## Scope

Keep production changes focused on `apps/api/src/services/messageService.js` plus focused service regression tests.

## Parent bounty

mdmcl-pixel:
## Summary

Fixes #12300 under parent bounty #11398.

`sendMessage()` currently allows caller payloads to overwrite the generated message ID and relies on `Date.now()` for uniqueness. This change keeps the existing `msg_` prefix, preserves `sentAt`, and makes message identifiers server-owned and collision-resistant with `crypto.randomUUID()`.

## Changes

- preserve ordinary caller payload fields while writing protected `id` and `sentAt` last;
- generate message IDs as `msg_<uuid>`;
- add focused regression coverage proving caller-supplied IDs cannot override the generated ID and same-millisecond creations remain distinct.

## Validation

`node --test apps/api/src/tests/messageService.test.js`

Result: 1 test passed, 0 failed.

AI-assisted contribution; the focused regression test was executed before submission.

### DIFF
--- a/apps/api/src/services/messageService.js
+++ b/apps/api/src/services/messageService.js
@@ -1,3 +1,5 @@
+import { randomUUID } from "node:crypto";
+
 const messages = [];
 
 export async function listMessages() {
@@ -5,7 +7,11 @@ export async function listMessages() {
 }
 
 export async function sendMessage(payload) {
-  const message = { id: `msg_${Date.now()}`, ...payload, sentAt: new Date().toISOString() };
+  const message = {
+    ...payload,
+    id: `msg_${randomUUID()}`,
+    sentAt: new Date().toISOString(),
+  };
   messages.push(message);
   return message;
 }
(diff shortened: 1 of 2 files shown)


mdmcl-pixel:
Closing this newer PR as a duplicate of #12297. Both implement the same message-ID integrity fix (server-owned, collision-resistant `msg_` IDs with focused regression coverage). Keeping the earlier PR as the canonical submission avoids splitting maintainer review.

PR #12347, opened 2026-09-04
fix(auth): verify requester on refresh endpoint (#2847)

armorbreak001:
## Summary

Fixes #2847 (💎 **Bounty: $780**)

### Problem
- `POST /api/auth/refresh` issued tokens **without verifying the requester**
- Refreshed token was always signed with hardcoded `usr_existing` / role `client`
- Any caller (authenticated or not) could get a token for a different user

### Changes
1. **authRoutes.js**: Add authMiddleware to the /refresh route
2. **authController.js**: Pass verified req.user payload to refreshToken()
3. **authService.js**: Sign refreshed token for the authenticated user's identity

### Security Impact
- Unauthenticated callers now get 401
- Token is signed for actual authenticated user (not hardcoded)
- Removes usr_existing identity from refresh flow

### DIFF
--- a/apps/api/src/controllers/authController.js
+++ b/apps/api/src/controllers/authController.js
@@ -22,6 +22,6 @@ export async function oauthCallback(req, res) {
 }
 
 export async function refresh(req, res) {
-  const result = await refreshToken();
+  const result = await refreshToken(req.user);
   return ok(res, result);
 }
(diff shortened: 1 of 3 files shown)


armorbreak001:
Hi! 👋 Just following up on this PR — happy to address any feedback or make adjustments if needed. The fix addresses `auth refresh verification` by adding proper authentication/validation. Thanks for reviewing! 🙏

PR #12406, opened 2026-09-11
fix(auth): require credential verification on login

### LINKED ISSUE #12405: [Bounty $780] POST /api/auth/login signs a valid token for any credentials — authentication bypass
## Bug

`POST /api/auth/login` returns a signed access token for ANY credentials. In `apps/api/src/services/authService.js`, `loginUser` never verifies the password against a stored user record — the implementation is a TODO stub that hard-codes `sub: "usr_existing", role: "client"` and signs it. Anyone can POST `{"email": "x@x.com", "password": "anything"}` and receive a valid 15-minute JWT for an existing user with client role.

## Expected fix

`loginUser` must resolve the user record by email, verify the submitted password against a stored hash (e.g. bcrypt), reject with 401 when the record is missing or the hash does not match, and sign the token for the authenticated user's subject and role only after successful verification. Auth route tests must cover: valid credentials → token for that subject/role; unknown email → 401; wrong password → 401.

## Scope

This issue is limited to the auth login route/service behavior for `/api/auth/login` and focused API tests.

References #743.

/bounty $780

copernicusjones:
## Problem

`POST /api/auth/login` issued a valid 15-minute access token for **any** credentials. `loginUser` in `apps/api/src/services/authService.js` was a TODO stub: it never resolved the user record, never verified the password, and signed a token with a hard-coded `sub: "usr_existing"` and `role: "client"`. A request with `{"email":"x@x.com","password":"anything"}` returned a signed JWT for an existing user.

A second, related defect: `POST /api/auth/login` with unknown or wrong credentials never rejected — the request **hung with no response** (details below).

## Fix

**1. Verify credentials before signing (`authService.js`)**
- `registerUser` stores a hash instead of the plaintext password: `scryptSync` (Node stdlib, no new dependency) with a 16-byte random salt per user, key length 64.
- `loginUser` resolves the record by email, verifies the submitted password against the stored hash with `timingSafeEqual` (constant-time), and throws a `401 "Invalid credentials"` domain error when the record is missing or the hash does not match.
- On success the token is signed for the authenticated subject and role only:

```js
return {
  email: stored.email,
  token: signAccessToken({ sub: stored.id, role: stored.role })
};
```

## Root cause of the hanging rejection

Express 4 does not forward rejections from `async` handlers to `next()`. `loginUser` rejecting with a 401 produced an unhandled promise rejection and **no HTTP response at all** — the client waited until it timed out. The 401 path was therefore unreachable, not merely incorrect.

- Added `apps/api/src/utils/asyncHandler.js` and wrapped the auth route handlers so async errors reach the error middleware.
- `errorHandler.js` now maps Zod validation failures to `400` with field details, and domain errors carrying a `status` to that status, before falling through to the existing `500`.

## Tests

Route-level tests over the real Express app (`apps/api/src/tests/auth.test.js`), covering exactly the behaviors named in the issue:

| Case | Expected | Result |
|---|---|---|
| valid credentials → token | `200`, token `sub` = registered user id, `role` preserved, `sub !== "usr_existing"` | pass |
| unknown email | `401`, no token in payload | pass |
| wrong password | `401`, no token in payload | pass |
| malformed body (invalid email) | `400`, no token in payload | pass |

```
✔ POST /api/auth/login returns a token for the authenticated subject and role
✔ POST /api/auth/login rejects an unknown email with 401
✔ POST /api/auth/login rejects a wrong password with 401
✔ POST /api/auth/login rejects a malformed body with 400
  tests 5 · pass 5 · fail 0
```

Verified with `npm test` from the workspace root (the documented command), Node v24.12.0.

## Note on the test script

`apps/api/package.json` had `"test": "node --test src/tests"`, which fails on the current Node LTS line before any test executes (`Cannot find module .../src/tests` — the runner treats the path as a module rather than a directory). Corrected to `node --test` so default test discovery runs the suite. Without this the auth tests could not be exercised by the documented command. Flagging it explicitly since it is adjacent to the issue scope — happy to split it out if you prefer.

Closes #12405

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "node src/server.js",
     "start": "node src/server.js",
-    "test": "node --test src/tests"
+    "test": "node --test"
   },
   "dependencies": {
     "cors": "^2.8.5",
--- a/apps/api/src/middleware/errorHandler.js
+++ b/apps/api/src/middleware/errorHandler.js
@@ -1,9 +1,20 @@
 export function errorHandler(err, req, res, next) {
+  // Zod validation errors -> 400 with field details
+  if (err?.errors?.length) {
+    return res.status(400).json({
+      success: false,
+      message: "Validation failed",
+      errors: err.errors.map(({ path, message }) => ({ path: path.join("."), message }))
+    });
+  }
+  // Domain errors carry an explicit status (e.g. 401 from bad credentials)
+  if (err?.status && err.status >= 400 && err.status < 500) {
+    return res.status(err.status).json({ success: false, message: err.message });
+  }
   console.error("Unhandled API error:", err);
   if (res.headersSent) {
     return next(err);
   }
-
   return res.status(500).json({
     success: false,
     message: "Unexpected server error"
--- a/apps/api/src/routes/authRoutes.js
+++ b/apps/api/src/routes/authRoutes.js
@@ -1,9 +1,10 @@
 import { Router } from "express";
 import { login, oauthCallback, refresh, register } from "../controllers/authController.js";
+import { asyncHandler } from "../utils/asyncHandler.js";
 
 export const authRoutes = Router();
 
-authRoutes.post("/register", register);
-authRoutes.post("/login", login);
+authRoutes.post("/register", asyncHandler(register));
+authRoutes.post("/login", asyncHandler(login));
 authRoutes.get("/oauth/:provider/callback", oauthCallback);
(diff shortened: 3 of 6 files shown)


copernicusjones:
Closing — the repo's bounty path is unreachable: 0 PRs have ever been merged here (search: is:pr+is:merged → 0) and main carries no merge commits, while CONTRIBUTING pays only on merge. No point leaving this open. The work and tests stand on their own if the maintainer ever opens a real review lane.

PR #12450, opened 2026-09-15
fix: registerUser uses same userId for response and JWT sub (fixes #12444)

### LINKED ISSUE #12444: Bug: registerUser access token can reference a different user id
## Problem

In `registerUser()` the returned `id` and the JWT `sub` claim use separate `Date.now()` calls, so they can differ by milliseconds, producing a token whose subject does not match the user's id.

## Expected Fix

1. Compute `const userId = 'usr_' + Date.now()` once at the start of the function
2. Use the same `userId` for both the response `id` and the JWT `sub`

References #743.

/bounty 30

slsgzs-cloud:
/claim #743

## Summary

Fixes **#12444** — `registerUser()` called `Date.now()` twice (once for `id`, once for `sub`), producing tokens whose subject didn't match the user's id.

### `apps/api/src/services/authService.js`
- Compute `const userId = 'usr_' + Date.now()` once at the top
- Use the same `userId` for both the response `id` and JWT `sub` claim

Now the returned `id` and the JWT subject always match.

Fixes #12444


slsgzs-cloud:
This PR fixes a race condition in registerUser where userId and JWT sub could differ. One-line fix. Ready for review!

PR #12451, opened 2026-09-15
fix: refreshToken now requires and verifies a token (fixes #12445)

### LINKED ISSUE #12445: Bug: refreshToken issues token without verifying the refresh token
## Problem

`POST /api/auth/refresh` calls `refreshToken()` with no argument — it returns a signed access token for `usr_existing` regardless of whether the caller holds a valid refresh token. This is equivalent to a public endpoint that issues admin-access tokens.

## Expected Fix

1. Require an authorization header with a refresh token
2. Verify the refresh token signature and expiry
3. Return 401 if missing or invalid

References #743.

/bounty 60

slsgzs-cloud:
/claim #743

## Summary

Fixes **#12445** — `POST /api/auth/refresh` issued a valid access token with no authentication. The `refreshToken()` function was called with zero arguments, meaning anyone could call the endpoint and get an admin token.

### `apps/api/src/services/authService.js`
- `refreshToken()` now requires a `refreshTokenStr` parameter
- Returns 401 if token is missing
- Verifies the token signature via `verifyAccessToken()`
- Returns new token only for valid, unexpired tokens

### `apps/api/src/controllers/authController.js`
- `refresh()` controller now extracts the `Authorization: Bearer <token>` header
- Returns 401 if header is missing or malformed

## Verification

The endpoint now properly rejects unauthenticated requests.

Fixes #12445


slsgzs-cloud:
This PR fixes a critical auth bypass — refreshToken issued tokens with no verification. Ready for review!

slsgzs-cloud:
Hi! This PR is **mergeable and clean** — fixes a critical auth bypass in refreshToken. Ready for merge!

